PR #208, opened 2020-06-08
Fix #207: UI text is selectable

### LINKED ISSUE #207: UI Text is selectable
Various UI elements like buttons and section titles can be selected like normal text.
[Like this](https://gyazo.com/0b7d614e30b8f21f59f5a903875de1ba.gif)

This can be avoided by using the [user-select property](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select) on all elements and then having a css class that overrides this for some fields that it makes sense.

Personally, I think the only selectable fields should be: 
- Mod names and descriptions
- Code blocks such as in the default Launch parameters

`user-select` doesn't affect input fields, so that won't be an issue, and the config values keys' are input fields so they are also selectable

If you think there should be more selectable fields feel free to chime in.

I'll be preparing a PR for this.

skourismanolis:
Fixes issue #207.


skourismanolis:
also made the error modal text selectable, did I miss anything?

ebkr:
I'll give it a quick test on my end when I'm free, if there's any issues I'll let you know, thanks again!

ebkr:
So, I've given it a go, and the only issue I really have is that text becomes quite awkward to de-select.
I had a quick play with the CSS you edited, and I think that:
```
.not-selectable, .not-selectable *, button, a {
    user-select: none;
}
```
Provides a slightly nicer feel. This does then mean that some stuff will require not-selectable. I'm also debating whether or not "a" text shouldn't allow selection either, and could be best to leave those.

This would mean that not-selectable would be required on the navigation menu, and possibly some card elements too, but would also provide slightly more control over what actually gets blacklisted. One noticeable impact was on the help section, nothing could be copy/pasted.

Thoughts?

skourismanolis:
I used your idea with the single `non-selectable` class. Should the titles of the mods be selectable? if so, I can add a `selectable` class to override this.

ebkr:
Preferably yes, although labels don't need to be. I don't mind if labels can or can't be though.

skourismanolis:
Made mod titles selectable: 
image
image

skourismanolis:
Made settings selectable.
image

ebkr:
I've had a play around with it, and I think it'd be better to remove the settings one as it's a little awkward de-selecting things.
It might also be a good idea to remove the <a> restriction, as that isn't selectable when copying help text.

Once that's done, I think it'll be good to merge.

skourismanolis:
I reverted the settings commit and made the links in the help section selectable.

Issue #296, opened 2020-09-09
[FEATURE REQUEST] - Improve mod list clarity


ChiriCuddles:
**What problem does the feature request solve?**
There were some issues with the clarity of the mod list.

**Example mock-up**
Author bylines differentiated from mod names:
![image](https://user-images.githubusercontent.com/6081834/92620783-664e9200-f306-11ea-8011-46a6669aab5a.png)

Disabled mods are more transparent than enabled mods:
![image](https://user-images.githubusercontent.com/6081834/92622008-e6292c00-f307-11ea-85bd-34d3e27b7db2.png)

Mod footer buttons don't have borders anymore that make it hard to tell which card they belong to:
![image](https://user-images.githubusercontent.com/6081834/92621006-a877d380-f306-11ea-871a-cc36f018ef33.png)

Mod footer buttons have a similar appearance to navigation buttons on the left side of the application:
![image](https://user-images.githubusercontent.com/6081834/92621081-bb8aa380-f306-11ea-9fde-9c41d29fea9d.png)

The version link has the same click region & appearance as the other footer buttons:
![image](https://user-images.githubusercontent.com/6081834/92621184-d8bf7200-f306-11ea-92d6-2fedcb3f2a23.png)

**Additional information**
This "mockup" is actually implemented on my end, but I wasn't sure if I should jump straight into making pull requests for all of it, and how much I should break it up for said PRs, so I figured I'd make an issue first. 

Lemme know your thoughts! (I'm in the Discord as well btw, my username is Chirichirichiri. If you wanna discuss things via DM)

ebkr:
Hi, thanks for raising it!
It's best to keep them appearing as clickable (rather than only on hover), so having them blend in isn't ideal.
Possibly adding vertical borders might be better?

How does it appear for light theme?

Yuudaari:
Borders don't really read to me as "clickable", they read to me as "table". (Especially given that the mod link previously wasn't clickable for the full size of the button.) Generally I find that "clickable" can be inferred through different colours, which is a thing in both themes (though it's more clear in the light theme given that the text is blue rather than grey.)

Sorry about not uploading light theme, forgot:
![image](https://user-images.githubusercontent.com/6081834/92623394-8469c180-f309-11ea-9132-1f6ad21e9df1.png)

ebkr:
Yeah, you're right that they don't appear clickable in the dark theme. The blue definitely helps with that. I think the borders are fine to leave.

[...]

PR #298, opened 2020-09-11
CSS theme cleanup

ChiriCuddles:
These are some pretty sweeping changes and they might be divisive, so I understand if you don't want to merge this as-is (or at all 😅) With that being said, here's how I got here (don't worry, there's a TL;DR at the bottom):

## The Problem
When I was working on #297, making edits to the dark theme was really annoying because I wasn't in an SCSS file. It was also hard to figure out why things had the colours they did, since clicking on any CSS file source in the Style panel of the DevTools was just sending me to these massive lines of compressed CSS code, and it just felt inconsistent and hard to work with. What I wanted was a few things: 
1. An SCSS colour palette of variables.
2. Being able to see why things got the styles they did.

That doesn't sound too hard, no? Well...

## Changing how Bulma is imported
The first step was actually really easy in the end, I just needed to [move the Bulma importing from the `App.vue` file to the `app.scss` file.](https://github.com/ebkr/r2modmanPlus/commit/38ef26e6d13d840f2c6813a7c3b7ded448b4b26b) Turns out Bulma can be imported super easily because of `sass-loader`!

Just doing this much actually helped quite a bit, because this allowed clicking on CSS file sources to show Bulma's SCSS instead.

Seeing Bulma's SCSS sent me down the next rabbit hole — *Bulma has overrideable CSS variables for a ton of stuff.*

## Removing `superhero.bulma.css`
So, if I wanted to be able to use SCSS in the dark theme, I wasn't going to be able to rely on the stuff in `superhero.bulma.css` anymore. That meant either figuring out how to enable/disable a SCSS stylesheet at runtime (which felt like it was going to be one of those things I'd be banging my head on a desk for ages about), or going all in, removing the `superhero.bulma.css` theme, and reimplementing it by replacing all the Bulma variables with CSS vars that I could much more easily change at runtime.

I went for the latter option, I just liked the idea of it more. I wasn't attached to that theme anyway, since [its primary colour was a little hard to parse interactables for](https://github.com/ebkr/r2modmanPlus/issues/296#issuecomment-689678785), and it didn't have full parity with the light mode anyway. (Which, in my opinion, it should.)

## How to toggle "dark mode" without appending/unappending a CSS stylesheet
I quickly found out that Quasar actually had a dark mode built in, so I started with that. Maybe Bulma would support that dark mode? All I had to do is replace the ThemeManager's method with one that calls Quasar's `Dark.set()`, making it behave exactly as the old button did, for some base elements... except for all the Bulma stuff, which has exactly zero support for Quasar's Dark Mode. Ah well.

Note: Quasar's Dark Mode seems like it's supposed to be able to detect whether the user prefers dark or light mode, but I'm not sure if that's in, or enabled or whatever. Could use some more research.

Unfortunately, for some reason, Quasar only applies a dark mode class to `<body>`, not `<html>`, which means the `<html>` has no way to change its background by default. As a result of this, we now watch for `$q.dark.isActive` changing in `App.vue` and toggle an `.html--dark` class as-needed.

## `light.scss` and `dark.scss`
So, next up is mapping those Bulma variables I found with CSS variables. Here I was sad to find that Bulma does a bunch of operations on some main vars, like `darken` and `rgba` and such, rather than using them as-is, which meant that I wasn't able to change their core palette and instead had to override all the component variables we're using individually. That was okay, meant a bit more work to set up though.

Here's what `light.scss` ended up containing:
1. **"custom"/core scheme variables.** These variables don't override anything, but they contain pretty much the whole palette. They each follow the pattern `var(<CSS variable that may or may not exist>, <fallback from Bulma's default theme>)`
2. **Bulma overrides.** These variables were like the above, but they actually overrode the defaults in Bulma's SCSS files. A lot of these reference the core scheme variables.
3. **Other Bulma overrides.** Some overrides like shadows are a bit more complex, so they go in their own section and reference core scheme variables directly.
4. **"custom" variables (again).** These are variables for non-Bulma elements/components that we apply ourselves in our SCSS files.

Then comes `dark.scss`, which is all about overriding the CSS variables we reference in `light.scss` when the `.html--dark` class is applied. Here's what it ended up containing:
1. **A colour palette.** (Hallelujah! My dreams have become reality!) This palette is based on a combination of the old Superhero theme and the default Bulma theme, in order to have as complete of parity between the two themes as possible.
2. **CSS var overrides.** These will almost always reference a palette colour.

### `custom.scss`
1. Most of the SCSS code is in here now rather than `app.scss`, because any SCSS entry file that uses variables (it detects a `$`) has the Quasar variables injected, and these Quasar variables infect Bulma and break its defaults (which we rely on.) Again, just moving our custom code to `custom.scss` so that there's no `$`s there fixes that issue.
2. Bulma doesn't actually have variables for everything, it's actually a little inconsistent, so there's a bit of manual colour application done in the beginning of `custom.scss` in order for the themes to have full parity.

### Misc cleanup
1. Any colours used in the old `app.scss` code (now in `custom.scss`) are now referencing one of our new variables in `light.scss`.
2. The sticky headers which use the `.card` class had a border-radius by default, and were previously fixed by adding a background to `.sticky-top`. I changed this so that instead there's an `.is-square` class on those elements that just removes the `border-radius`
3. To fix some icon colour stuff in `ExpandableCard`, I changed the header anchor element to a `div` and added the `.cursor-pointer` class to it, and I also changed the disabled sort buttons to have the `.disabled` class instead of being a completely different element.

## Screenshots
Settings menu
image

Help menu with hovered link
image

Config editor with hovered `select`
image

The actual config editor with a hovered `input`
image

Mod list with hovered card icon
image

View associated popup
image

Custom launch parameters popup
image

Profile selection
image

Create a profile
image

# TL;DR
Here's an overview of all the changes:

### In `light.scss`, replace any bulma color variables used by components we're using with CSS variables
- When we can't use a CSS variable, we instead override the color manually with a custom rule in `custom.scss` (most of the time this is unnecessary, tho)

### Use Quasar's "Dark Mode" as base
- Rather than appending/unappending `#darkThemeStyle` element, use `Dark.set(settings.darkTheme)`
- Watch for when Quasar's "Dark Mode" is toggled, and when it is, toggle the `.html--dark` class on the `html` element
- Override Quasar `$dark-page` to a CSS var

### Add a `dark.scss` file for, when `.html--dark`, replacing the values of all the CSS vars we have defined in `light.scss`
- Contains a colour palette based on superhero theme for consistency
- Uses the same blue colour for links and buttons as the light theme!

### Other misc changes
- Add `.is-square` for `border-radius: 0` on sticky headers
- Change `ExpandableCard` header from an anchor to a `div` with `.cursor-pointer`, fixing some colour issues in the process
- Cleanup `ExpandableCard` `.card-header-icon`s, don't duplicate the whole icon for each if/else, instead just toggle the `.disabled` class

# #297?
If it gets merged, I'll merge its changes into this pull.

# ...Anyway...
I'm so sorry at how long this explanation got and how sweeping these changes are. I didn't expect it to get quite this big going into this project. Feel free to do with this what you want, honestly. Merge, discard, whatever. If you need anything more from me, I'll be watching the pull and stuff, and you can always contact me thru email or Discord (Chirichirichiri on the server) if you need to. Here for whatever!

I hope this stuff is at least useful in some way! Thanks for making a fun project to hack on! 🤍


ebkr:
I'd honestly love to be able to get rid of that superhero.bulma.css file. The primary reason was to easily allow switching between themes without having to restart the app. I haven't actually checked any of the code yet (or for the other PR yet, plan to in a few days), however is this still possible? (I'm assuming so from the .html--dark class).

Not sure if you have access to the mod-management channel in the modding discord, although there's also the r2modman discord server if you're able to join that.

I'm glad you've found it fun to work on, I hope it hasn't been too much of a mess! If there's anything you feel like needs work, or any issues you ever want to pick up, just let me know!

Yuudaari:
Oh, I didn't reply to everything. 

Not gonna lie, I've been like "whyyy is this like this" a couple times, but honestly I do that for every codebase, even my own that I set up only days prior 😜

Yeah, I'd be up for tackling other things (tho mainly UI stuff, since I'm a UI/UX designer by trade.) I'll see about maybe tackling other stuff once these current PRs are dealt with, I don't want it to get too overwhelming to merge it all

src/css/scheme/dark.scss
```diff
@@ -0,0 +1,58 @@
...
+    $night-b  : #bdcbdb;
+    $night-c  : #cedef1;
+    $night-d  : #d9e3ee;
+    $night-e  : #e6eff8;
+    $primary-1: #3273dc;
+    $primary-2: #5596ff;
+    
+    // overriding variables with dark theme colors
+    
+    --background                        : #{$night-1};
```
ebkr, line 18: Not a fan of this formatting, I'd prefer it to be:
```suggestion
    --background: #{$night-1};
```

ebkr (changes requested):
Just formatting needs updating, haven't tested locally yet but styling changes look good.

ebkr:
Yeah that's fair enough, and by all means don't feel as though you have to commit to anything! 
There's some pretty weird stuff that goes on, especially around the earlier areas of the project. This was basically my first attempt with TypeScript and Vue ¯\_(ツ)_/¯

src/css/scheme/dark.scss
Yuudaari: Ah, gotcha. Yeah, the main reason I did that was so that it's easier to do multi-cursor copying of names and values and such, but that's probably not a big priority. I can remove all that extra whitespace for sure

ebkr (approved):

PR #300, opened 2020-09-15
Limit line.split to 2 substrings to allow for config values containing =

xoxfaby:


ebkr:
Going to close this an implement the fix via Discord.

Image for reference:

image

PR #309, opened 2020-09-21
Redesign mod cards

ChiriCuddles:
- Create `LocalModCard` & `OnlineModCard` for the two different mod wrappers, `ManifestV2` and `ThunderstoreMod`. There's slightly more duplicated HTML code as a result of this but each now only contains the halves specialised to themselves and it's all extracted from the mod lists, all of which is a win in my book.
    - Future refactoring would include creating a base `ModCard.vue`, and maybe even a `Card.vue` which would have slots for all of this. It was easier for now to do it as two separate files
- Remove "card display type" option (unused)
- Add "icon display mode" option which toggles between mod cards showing both notifications & actions, vs only showing notifications
- Add `bulma-switch` dependency
- Add support for more card display types
- Wrap config cards list in a `div` for styling
- Drag-sort installed mods

Misc:
- Rename `Link` to `ExternalLink` bcuz my editor was yelling at me for this and it was really distracting


ChiriCuddles:
- No draggable transition yet

ChiriCuddles:
- [ ] Options dropdown should have more padding on the top/bottom

src/components/cards/LocalModCard.vue
```diff
@@ -0,0 +1,132 @@
...
+    @Prop()
+    id: string | undefined;
+	
+	@Prop()
+	manifest: ManifestV2 | undefined;
+	
+	@Prop({default: () => []})
+	missingDependencies: string[] | undefined;
+
+    @Prop({default: false})
```
ebkr, line 108: Formatting

src/components/cards/LocalModCard.vue
```diff
@@ -0,0 +1,132 @@
...
+	   :class="{'card--is-disabled': !manifest.isEnabled(), 'card--is-deprecated': manifest.isDeprecated()}">
+		<figure class='card-image image is-72x72 image-parent mod-logo' v-if="manifest.icon !== ''">
+			<img :src='manifest.icon' alt='Mod Logo' class='image-overlap'/>
+			<img v-if="funkyMode" src='../../assets/funky_mode.png' alt='Mod Logo' class='image-overlap'/>
+		</figure>
+		<header class='card-header is-shadowless' :id='id'>
+			<span class='card-header-title'>
+				<div class="field card-toggle has-tooltip-top"
+					 :data-tooltip='`This mod is ${manifest.isEnabled() ? "enabled" : "disabled"}.`'>
+						<input :id="`toggle${id}`" type="checkbox" class="switch is-info is-rounded"
```
ebkr, line 12: To keep it consistent it'd usually be toggle-${id}

src/components/cards/LocalModCard.vue
```diff
@@ -0,0 +1,132 @@
...
+		}, 1);
+	}
+	
+	// when the manifest changes, re-render this component
+	@Watch("manifest", {deep: true})
+	onManifestChange() {
+		this.$forceUpdate();
+	}
+	
+    emitMove(direction: string) {
```
ebkr, line 124: Not needed after drag/drop functionality

src/components/views/LocalModList.vue
```diff
@@ -116,78 +116,28 @@
...
+        <draggable ref="draggable" v-model='sortableModList' group="mods"
+                   animation="200"
+                   @start="startSort()" 
+                   @end="endSort()">
+                    <local-mod-card v-for='(key, index) in sortableModList' :key="'local-' + key.getName()"
+                                    :id="index"
+                                    :manifest="key"
+                                    :missingDependencies="getMissingDependencies(key)"
+                                    :funkyMode="settings.funkyModeEnabled"
+                                    @update:enabled="toggleMod(key, $event)"
```
ebkr, line 128: Could you change the event name to update-enabled to be more inline with other event names?

src/components/views/LocalModList.vue
```diff
@@ -393,13 +338,27 @@
             this.showingDependencyList = true;
         }
 
+        /**
+         * @returns whether the mod was uninstalled. If the mod was not uninstalled, an uninstall comfirmation is shown
+         */
```
ebkr, line 343: Almost.
So what happens here is that it checks if there are any dependants. If there are any dependants, it'll show a confirmation popup. If there are no dependants, then it just uninstalls the mod as nothing else would be affected.

src/components/views/LocalModList.vue
```diff
@@ -436,6 +395,14 @@
                 Logger.Log(LogSeverity.ACTION_STOPPED, `${err.name}\n-> ${err.message}`);
             }
         }
+        
+        toggleMod(vueMod: any, enabled: boolean) {
+            if (enabled) {
+                this.enableMod(vueMod);
+            } else {
+                this.disableModRequireConfirmation(vueMod);
```
ebkr, line 403: Could be good to have it similar to uninstalling by checking the dependant count

src/css/components/Card.scss
```diff
@@ -0,0 +1,347 @@
+@function str-replace($string, $search, $replace: '') {
```
ebkr, line 1: Is this a replace all?
Might be good to have a comment here

src/css/components/Icon.scss
```diff
@@ -0,0 +1,11 @@
+.icon {
```
ebkr, line 1: Comment explaining this would be good

src/css/custom.scss
```diff
@@ -90,55 +90,8 @@ select {
...
-}
-
-.card-byline, .card-timestamp {
-    color:$scheme-text-secondary;
-}
-
-.card-timestamp strong {
-    color: inherit;
-}
+@import "components/Card";
```
ebkr, line 93: Minor nitpick but SCSS filenames are currently lowercase

src/components/views/LocalModList.vue
ChiriCuddles: Isn't that what `disableModRequireConfirmation` is doing?

src/css/custom.scss
ChiriCuddles: Ah, yeah, that's my bad, I'm used to my own way of doing things which is camelCase for "general" files and PascalCase for singular components. I'm guessing you'll prefer kebab-case? I'll remember to use that instead

ChiriCuddles:
- [ ] Add some kind of handle to indicate that local mod cards can be dragged to sort
- [ ] Add button for changing installed version of a mod (online-tab-only for now)
- [ ] Make title (and switch?) more smol
- [ ] Experiment with compact display mode
- [ ] Unindent card content
- [ ] See if it's possible to remove the padding above the card header

PR #352, opened 2020-12-06
Merge feature/310 with non-Windows platforms support (AKA Linux)

### LINKED ISSUE #177: Native Linux Support
For the most part, it's just a case of replacing all `powershell` calls with something that can run under Linux.

Best solution is to create a new class that can divert calls to platform specific applications.

**Potential issues:** 
- I might have to replace the launch method to call the EXE directly.
- Symlinks aren't guaranteed to work. Best solution is to remove the r2modman folder from the Risk of Rain 2 install directory, and point doorstop to use the absolute profile path.
- Anything that opens `explorer` can potentially be an issue. `xdg-open` is a viable replacement, however still lacks support on all distros.

AryToNeX:
This closes #350 (as outdated) and it also closes #177 by proxy.

# IMPORTANT:
For the game to be able to launch modded, you need to set your launch options on Risk of Rain 2 as follows: `WINEDLLOVERRIDES="winhttp=n,b" %command%`

## THIS IS ALPHA QUALITY SOFTWARE

As such, no prebuilts are given. If you're a Linux user, use the "stable" AppImage over at #350.


src/r2mm/manager/GameDirectoryResolver.ts
```diff
@@ -5,38 +5,58 @@ import VdfParseError from '../../model/errors/Vdf/VdfParseError';
 import child from 'child_process';
 import * as vdf from '@node-steam/vdf';
 import * as path from 'path';
+import { homedir } from 'os';
 import ManagerSettings from './ManagerSettings';
 import FsProvider from "../../providers/generic/file/FsProvider";
 
-const installDirectoryQuery = 'Get-ItemProperty -Path HKLM:\\SOFTWARE\\WOW6432Node\\Valve\\Steam -Name "InstallPath"';
+const win32_installDirectoryQuery = 'Get-ItemProperty -Path HKLM:\\SOFTWARE\\WOW6432Node\\Valve\\Steam -Name "InstallPath"';
```
ebkr, line 12: Would prefer this to be `windowsInstallDirectoryQuery` as no underscores are used in the code style.

src/r2mm/manager/GameDirectoryResolver.ts
```diff
@@ -53,20 +73,16 @@ export default class GameDirectoryResolver {
...
-            const queryResult: string = child.execSync(`powershell.exe "${installDirectoryQuery}"`).toString().trim();
-            const installKeyValue = queryResult.split('\n')[0].trim();
-            // Remove key (InstallPath) from string
-            const installValue = installKeyValue.substr(('InstallPath').length)
-                .trim()
-                // Remove colon
-                .substr(1)
-                .trim();
+            const installValue = await this.getSteamDirectory();
+            if(installValue instanceof R2Error)
```
ebkr, line 77: Missing space

src/r2mm/manager/GameDirectoryResolver.ts
```diff
@@ -53,20 +73,16 @@ export default class GameDirectoryResolver {
...
+                throw installValue;
+
             const dir = await this.findAppManifest(installValue);
             return dir;
         } catch(e) {
             const err: Error = e;
             return new R2Error(
-                'Unable to resolve steam install directory',
+                'Unable to resolve the Risk of Rain 2 install directory',
```
ebkr, line 85: Thanks 👀

src/r2mm/manager/GameRunner.ts
```diff
@@ -10,17 +9,37 @@ import Profile from '../../model/Profile';
...
+            default:
+                return 'steam';
+        }
+    }
+
     public static async playModded(ror2Directory: string, onComplete: (err: R2Error | null) => void) {
         LoggerProvider.instance.Log(LogSeverity.INFO, 'Launching modded');
         const settings = await ManagerSettings.getSingleton();
         const steamDir: string | R2Error = await GameDirectoryResolver.getSteamDirectory();
+        const preloaderDir: string =
```
ebkr, line 27: Close, but not quite.
The `BepInEx.Preloader.dll` is stored in the `<profile>/BepInEx/core/` directory. This will always be resolvable unless BepInEx has not been installed, in which case the command will be ignored.

This has nothing to do with Wine/Proton.

AryToNeX: Trust me it needs to be there to work properly.

You're used to this kind of scenario where you pass the path of BepInEx as an argument and call it a day; and it would work fine if we're strictly talking Windows.

On Linux, every Windows game is ran in a sandbox and you cannot just pass Linux paths to it. The `path` module alone won't ever return a sandbox-compatible path, and that is why I needed to add this constant and prefix the path with Z: if we're not on Windows, again, because
- The sandbox behaves like Windows would, if we're seeing it from the game's perspective
- The *nix root file system is mounted at Z: in the sandbox
- The *nix file system doesn't have letters as mountpoints, but it has the root (/) and subsequent directories

Remember a key thing: r2modman will run natively on Linux so it has to account for Wine/Proton. This has nothing to do with the DLL itself existing or not.

ebkr:
I know this got mentioned on the Discord server but I'll leave this here just in case.
1. Migrate to a provider implementation.
2. Fix icons (hopefully) from the Quasar upgrade.

I've added GameDirectoryResolver since too. (#369)
The providers should give you a decent base that allow you to essentially hijack and overload the base functionality with something Linux specific.

For the provider implementation:
1. Create a new class file that extends GameRunnerProvider.
2. Conditionally provide the correct impl based on the platform in App.vue.
3. Repeat for GameDirectoryResolver.

See provider usage below:
```ts
// App.vue
const linuxGameRunnerProvider = new LinuxGameRunnerProvider();
GameRunnerProvider.provide(() => linuxGameRunnerProvider);

// Other.ts
GameRunnerProvider.instance.method()
```
There should be no reason for you to ever have to call `GameRunnerProvider.instance.method()` as the underlying code should already support it. Let me know if there's something you can't do with a provider and I'll resolve it unless you feel like taking it on.

PR #398, opened 2021-02-12
Linux: improve default Steam dir autoconfiguration

AryToNeX:
Use the Steam catch-all just in case, BUT resolve links to their real path

This specifically fixes the issue of the guy over at Discord, though it is SUPER weird they got that bug in the first place. Maybe Ubuntu is doing stuff the non standard way again.

(guess you'll have to pick this on the dsp branch also}


riking (approved):
Looks good to me, should fix my problems.

ebkr (approved):

PR #399, opened 2021-02-13
Add missed dspDirectory change

riking:
When loading the settings from IndexedDB, the saved Dyson Sphere Program directory value was ignored.


ebkr:
Closing this because it's fixed in master currently.

PR #426, opened 2021-03-03
Better support for linux native games

jrobsonchase:
Pretty much everything else "just works" aside from the `Z:` prefix on the preloader path and unnecessary wine things.

Remaining problems:
* Users will need to set the launch options in steam to `./start_game_bepinex.sh %command%`. Not sure if there's any way around this apart from skipping `steam.sh` altogether and calling `start_game_bepinex.sh` directly in the game runner.
* Should probably add an override in the profile settings in the event that users want to force it to run through proton. Steam has a checkbox for it, so r2mm might as well too.


jrobsonchase:
Looks like #429 does this more completely

PR #455, opened 2021-03-29
Add mention of Scoop package to README

cedws:
A Scoop [package](https://github.com/Calinou/scoop-games/commit/14f791b24d0d83f8a1975511554c53158cb215f8) for `r2modman` is available in the unofficial games bucket.
I can't guarantee I'll be able to maintain this forever, but it is fairly easy. The JSON manifest can automatically update using the `checkver.ps1` tool shipped with Scoop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,11 @@ You can choose to update, enable/disable or even uninstall mods with a simple cl
 ## Installing
 
 ### First time installing
+#### Packages
+If you use the [Scoop](https://scoop.sh/) package manager, you can install by running:
+1. `scoop bucket add games`
+2. `scoop install r2modman`
+
 #### Windows
 1. Click "Manual Download" on Thunderstore.
 2. Inside the downloaded **.zip** file. Run the "r2modman Setup X.X.X.exe" (where X.X.X is the current version).


ebkr:
Hey, thanks for the PR!

I'd prefer to host a scoop package myself in terms of what goes in the README. At the same time, I'm not really sure how many people would end up using it in terms of it being worthwhile maintaining.

ebkr:
To add to this, I could potentially host my own if that's something you're interested in

cedws:
There are benefits to having the package in a bucket (Scoop's term for package source) someone else maintains, mainly that you don't need to think about how to host it, or whether it's working or not.

The other benefit to having it in this `games` bucket is that it's more readily available. Anyone using Scoop who might want to download `r2modman` can `scoop search` it without having to check if there is a separate bucket for it.

If you concerns are along the lines of malicious takeover, the bucket maintainers do appear to vet manifest PRs to some degree.

Your concerns are understandable, so if you want to close this, I won't complain. I'll continue to maintain the manifest in the `games` bucket as long as there isn't a duplicate anywhere else. Don't want to create confusion.

PR #552, opened 2021-06-04
Rounds support

ebkr:


ebkr:
Also includes improved local mod importing

PR #673, opened 2021-11-21
Added Open File option in config editor

MSchmoecker:
This adds the option to open a config in the file system without having to navigate to the file.
Why? Sometimes it is just easier to edit the file directly instead of using the UI.

grafik


ebkr:
Oh nice, thanks!
I'll check this out some time tomorrow.

src/components/config-components/ConfigSelectionLayout.vue
```diff
@@ -67,6 +68,7 @@
     import ConfigSort from '../../r2mm/configs/ConfigSort';
     import FsProvider from '../../providers/generic/file/FsProvider';
     import ManagerInformation from '../../_managerinf/ManagerInformation';
+    import LinkProvider from 'src/providers/components/LinkProvider';
```
ebkr, line 71: Had a quick look at it all looks good (just need to test manually).
Could this line be relative like the other imports?

ebkr (changes requested):

src/components/config-components/ConfigSelectionLayout.vue
MSchmoecker: Sure, is changed :)

ebkr (approved):

PR #674, opened 2021-11-21
WIP: Rollbackable settings

ebkr:
Currently people cannot downgrade to older versions if a new game was added.
This will only prevent downgrades should schema structure changes occur.

Additionally added a setting in preparation of launching via Virtual Desktop.

TODO:
- [X] Unique migration versions
- [x] Slow rollout
- [x] Tests


ebkr (commented):
Need to add tests

MythicManiac:
Some preliminary questions:
- Is this able to exist in parallel with the older settings version, i.e. if we need to rollback **this** migration?
- Can we make this so we can take this into use gradually, e.g. only for some games at first to limit potential issues? Perhaps even toggleable remotely if we notice issues with the deployment.
- Tests should be included as you mentioned. Perhaps at least for the following scenarios
  - Migration from v2 settings to v3 settings
  - Migration from v3 settings to v2 settings
  - Migration from v3 settings version to an upper v3 settings version
  - Rolling back a settings version in v3
- This seems to lack validation that ensures migration IDs are unique and there can't be duplicate declarations of the same migration ID. Adding a validation step for it would be a good insurance to have upfront.

I'd be most comfortable rolling out a solution that can be gradually taken into use (or rolled back if issues arise), making it less risky to make a major change such as this.

MythicManiac (commented):
Overall looks good as far as I can tell, but still needs a bit of work to make it production ready (refs. previous comment).

ebkr:
Taken a look and here are some answers:
> Is this able to exist in parallel with the older settings version, i.e. if we need to rollback this migration?

Yeah, the old settings are still preserved and so code rollbacks will just use the old method.

> Can we make this so we can take this into use gradually, e.g. only for some games at first to limit potential issues? Perhaps even toggleable remotely if we notice issues with the deployment.

Maybe by using some switch statement, but probably a bit awkward and feature flags wouldn't be super reliable given people could always be offline.

Happy to restrict it to certain games though. Perhaps Northstar could be a low impact one to switch initially?

> This seems to lack validation that ensures migration IDs are unique and there can't be duplicate declarations of the same migration ID. Adding a validation step for it would be a good insurance to have upfront.

Good spot

PR #701, opened 2021-12-30
Added config slider

MSchmoecker:
This adds a slider inside the config settings for values that have a `# Acceptable value range: From [a] to [b]` setting.
It is only applying integer values even when the field is set to another type.

grafik

### DIFF
--- a/src/components/config-components/ConfigEditLayout.vue
+++ b/src/components/config-components/ConfigEditLayout.vue
@@ -38,14 +38,24 @@
                             <span class="pre" v-else>{{getCommentDisplay(line.comments)}}</span>
                         </p>
                         <br/>
-                        <template v-if="line.allowedValues.length > 0">
-                            <select class="select select--full" v-model="line.value">
-                                <option v-for="(opt, index) in line.allowedValues" :key="`opt-${key}-${configFile.getName()}-${variable}-${index}`">
-                                    {{opt}}
-                                </option>
-                            </select>
-                        </template>
-                        <input class="input" v-model="line.value" v-else/>
+                        <div class='settings-input-container'>
+                            <template v-if='line.hasRange()'>
+                                <q-slider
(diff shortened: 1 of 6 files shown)


src/components/config-components/ConfigEditLayout.vue
```diff
@@ -38,14 +38,24 @@
...
-                            <select class="select select--full" v-model="line.value">
-                                <option v-for="(opt, index) in line.allowedValues" :key="`opt-${key}-${configFile.getName()}-${variable}-${index}`">
-                                    {{opt}}
-                                </option>
-                            </select>
-                        </template>
-                        <input class="input" v-model="line.value" v-else/>
+                        <div class='settings-input-container'>
+                            <template v-if='line.hasRange()'>
+                                <q-slider
```
ebkr, line 43: Do you think you could use a different component to q-slider?

Think you can just do:
```
<input :input="..." v-model="line.value" :max="line.getMinRange()" ...>
```
Not sure how parseFloat() is affected this way though.

Generally avoiding quasar components because of TMM porting, and I don't think I've used them anywhere else anyway.

ebkr (changes requested):

src/components/config-components/ConfigEditLayout.vue
MSchmoecker: Yes it is possible to use a `<input type="range" ...` I have just avoided it because the styling is a pain. But because of electron it only has to be configured for Chrome I guess?

ebkr: Yeah, although I think we may already have styling: https://wikiki.github.io/form/slider/

If not, feel free to add it to package.json

ebkr:
Sorry this has taken so long, I plan on getting around to this in a few days!

ebkr:
Just got around to reviewing this, looks good. Thanks

PR #718, opened 2022-02-10
Search by key instead of whole string

MSchmoecker:
The search is sometimes annoying as it has to be typed exactly like it is named, without making an error with spaces. I changed it to split the search between spaces and then every key must be present. It feels more natural to use.

May also related to #714 as underscores are not really typed by a user

### DIFF
--- a/src/components/views/LocalModList.vue
+++ b/src/components/views/LocalModList.vue
@@ -230,6 +230,7 @@ import Game from '../../model/game/Game';
 import ConflictManagementProvider from '../../providers/generic/installing/ConflictManagementProvider';
 import Draggable from 'vuedraggable';
 import DonateButton from '../../components/buttons/DonateButton.vue';
+import SearchUtils from '../../utils/SearchUtils';
 
 @Component({
         components: {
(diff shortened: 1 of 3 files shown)


ebkr:
Hey, sorry this has taken so long, it completely slipped my mind. I'll try to get around to this at the weekend.

xiaoxiao921:
Narrator: He in fact, did not, got to this that weekend.

MSchmoecker:
Ah yes, this got really old now. Still think it's a good QoL addition as you don't have to type the exact mod name. If wanted I can update it to the current version @ebkr

ebkr:
Ah yeah, oops. That'd be great if you could, thanks!

MythicManiac:
Thanks for the contribution 👍 the only potential downside I can think of with fuzzy matching is a lack of relevance ordering making it harder to find what you're looking for even if you write it as an exact match on the query.

That said, this is probably an overall improvement, so perhaps relevance ordering can be included on a later PR. We have a roadmap item for search & discoverability updates on the website, so chances are we'll be able to contribute some of that here when we get around to it.

MSchmoecker:
Finally got time to rebase this PR, the code had to be moved to the new component. It is ready and working again.

One additional point in favor of this change, the current search behavior is not really affected if you search for the exact name. You will will display only get the matching mod, since there are no spaces possible in mod names.

MSchmoecker:
@ebkr @MythicManiac any updates on this?

ebkr (approved):

ebkr:
Apologies for the delay, looks good to me and works as expected when testing locally. Thanks

PR #743, opened 2022-03-13
Added UnityDoorstop v4 support

bb010g:
The setting should probably be a dropdown instead of a toggle, but I'm not too familiar with Vue, so feel free to change that. Also, factoring at least part of the common UnityDoorstop flags logic might be a good idea.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "minimist": "^1.2.2",
     "mocha": "^8.0.1",
     "mock-require": "^3.0.3",
-    "node-sass": "^4.13.1",
+    "node-sass": "^7.0.1",
     "sinon": "^11.1.1",
     "ts-node": "^8.10.2",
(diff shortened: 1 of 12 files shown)


ebkr:
Hey, thanks for the PR!

I've got in touch with ghorsington regarding this and the outcome currently is that BepInEx releases will include some metadata file that we can read to determine the version. This means we'd be able to dynamically determine the version as opposed to having the user toggle it.

I'd also say that having the user toggle the setting isn't a great idea because we'd end up in a scenario where someone needs to and it's not immediately obvious. I mean I didn't even know it was currently v3 🤔. Expecting the user to know doesn't seem correct. We also get users who just change settings without knowing what they do/clicked (see: funky mode).

I'll close this for now as it's not actually released yet and there'll hopefully be a better solution when it comes around to it. More than happy to respond to any points/concerns/questions though.

Issue #759, opened 2022-04-04
[BUG] - App won't go past init screen on Linux


lonegamedev:
**Describe the bug**
App gets stuck on init screen. Clicking `Continue offline` doesn't help.

**To Reproduce**
Open app on Linux.

**Expected behavior**
App moves to game screen.

**Screenshots**
![Screenshot_20220404_181257](https://user-images.githubusercontent.com/13684645/161586864-b95bac91-6b26-4d7d-9f6a-298ae0a69375.png)

**Additional context**
Tried both pacman and AppImage - same issue.

lonegamedev:
Created PR https://github.com/ebkr/r2modmanPlus/pull/760

PR #760, opened 2022-04-04
fix-for-issue-759

lonegamedev:
Looks like that json structure for steam apps has changed again (at least for Linux). This patch extends previous solution to check for all known path variants. Possibly we could even get away with doing case-insensitive string comparison - since swapping upper and lower case is their thing :upside_down_face: 

In case nothing is found  `Steam.Apps not found !` warning will be outputted to the console - so hopefully it will be easier to find/report/fix in the future.

Since project is not using lodash or something similar I have added simple util function for finding object property by path.

https://github.com/ebkr/r2modmanPlus/issues/759

### DIFF
--- a/src/utils/Common.ts
+++ b/src/utils/Common.ts
@@ -1,0 +1,27 @@
+type Mappable = {
+  [key: string]: any;
+};
+
+export const getPropertyFromPath = (object: Mappable, path: string | string[]): any => {
+    if (!object) {
+    	return undefined;
+    }
+    if (typeof path === 'string') {
+  	    const parts = path.split('.');
+        for (let i = 0; i < parts.length; i++) {
+            if (!object) {
+                return undefined;
+            }
+            const key = parts[i];
+            object = object[key];
+        }
+        return object;
+    }
+    for (let i = 0; i < path.length; i++) {
+        const res = getPropertyFromPath(object, path[i]);
+        if (res) {
+            return res;
+        }
+    }
+    return undefined;
+};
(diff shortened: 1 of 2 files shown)


ebkr:
Thanks! I’ll take a look when I’m next available.
I’ll also see if I can find a way to log that discovering the key has failed so that this’ll flag up next time

src/utils/Common.ts
```diff
@@ -0,0 +1,27 @@
+type Mappable = {
+  [key: string]: any;
+};
+
+export const getPropertyFromPath = (object: Mappable, path: string | string[]): any => {
```
ebkr, line 5: This is quite a condensed method. I think it'd help to both rename iterators from `i` and to add some comments explaining why it's structured like this.

src/utils/Common.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (!object) {
```
ebkr, line 6: I'd also not worry about this check. TS won't allow a potential undefined/null value to be passed and we should be accounting for it outside of this method.

It wouldn't be a good idea to pass around potential null/undefined values in the first place, so we should be handling that prior to making any calls against the object anyway.

src/utils/Common.ts
ebkr: Another thing is that this method could likely be handled slightly nicer recursively by doing something like:
- Check if first index of path exists
  - If the path doesn't exist, the property is unnavigable. Return undefined.
  - If it does, recurse into the property and pass the new path (original being path[0..n], new being path[1..n])
- Once `path` is empty, return the object.

PR #769, opened 2022-04-18
Added support for manifest donation_link

ebkr:
**Local mod list**
image

image

**Online mod list (with and without donation_link)**
image


MythicManiac:
Perhaps it could be made visible in the collapsed state already, given that the goal is to give mod authors visibility?

PR #789, opened 2022-06-02
Added new PersistentData installation rule. Fixed several broken tests and added test for new rule.

ebkr:


MythicManiac (approved):
I don't think this is the thing GTFO should be doing, but adding support is fine, up to them how to use it

ebkr:
This isn't needed anymore. Will close.

PR #790, opened 2022-06-15
Serverlistui

VilppeRiskidev:
ServerList support added


src/components/views/ServerList.vue
```diff
@@ -0,0 +1,69 @@
...
+    }
+})
+
+export default class ServerList extends Vue {
+
+    @Prop({required: false})
+    private data!: Object;
+
+    private openServerDetails() {
+        alert('open server details clicked');
```
harbingerofme, line 63: Debug code?

MythicManiac (changes requested):
This should probably be structured in a way where it can be merged & deployed without it doing anything unless a server list provider is supplied, which should supply all of the implementations.

So in other words changes such as the `serverlistData` shouldn't actually exist where it does in this PR, but instead the provider interface or such.

In an ideal scenario this PR would do nothing but add capabilities for _providing_ a server list implementation, and only if provided, would it affect anything. An easy rule of thumb is to minimize modifications to existing files and instead move them to new files.

src/pages/Manager.vue
```diff
@@ -356,13 +367,15 @@ import GameRunnerProvider from '../providers/generic/game/GameRunnerProvider';
...
 import GameInstructions from '../r2mm/launching/instructions/GameInstructions';
+import ServerListProvider from "../providers/components/loaders/ServerListProvider";
 
 @Component({
 		components: {
             LocalFileImportModal,
             OnlineModList: OnlineModListProvider.provider,
             LocalModList: LocalModListProvider.provider,
             NavigationMenu: NavigationMenuProvider.provider,
+            'serverlist': ServerListProvider.provider,
```
MythicManiac, line 378: Why is the key a string here? Can we use the same formatting as the others (consts?)?

src/components/navigation/NavigationMenu.vue
```diff
@@ -122,6 +129,10 @@ import { PackageLoader } from '../../model/installing/PackageLoader';
             return this.activeGame.packageLoader === PackageLoader.BEPINEX;
         }
 
+        get canShowServerList(): boolean {
```
MythicManiac, line 132: This should probably read from the provider instead of be implemented here

MythicManiac: If the provider is simply a vue component, perhaps we should modularize the structure a bit more in a way where it can also include this logic

MythicManiac (changes requested):

src/pages/Manager.vue
anttimaki: Also PascalCase since it's used with the other keys.

src/components/navigation/NavigationMenu.vue
VilppeRiskidev: I don't quite understand how the provider works, how can this check be moved to the provider?🤔
Also, hard coding a game name seems like a bad practice, but it might be solved in the future?

Issue #729, opened 2022-02-23
[FEATURE REQUEST] - Flatpak Support


quasigod-io:
**What problem does the feature request solve?**
With the Steam Deck releasing soon, I think it would be great to be able to install this without needing developer mode enabled.

**Example mock-up (optional)**
I have no clue how to package a flatpak, but I will be looking into learning and might be able to do it myself.

zensei-n7:
1+ for this. 
For me, the problem would solve the compatibility for Flatpak version of Steam: It seems like the appimage version just doesn't work with Flatpak version of Steam and instead launches another instance of steam with login screen. I have flatpak version of Steam because I want to avoid multilib arch's repo and also because of Flatpak's sandboxing. 

I think the issue for this specifically can be solved by perhaps adding permission or some kind of link to Steam without avoiding the sandbox, but I have no idea to do that.

GeckoEidechse:
As r2modman ships its Linux build as an AppImage, the easiest solution would be to unwrap it inside a Flatpak sandbox and ship the resulting image as a Flatpak on Flathub. I recently did exactly this for a Northstar installer called [Viper](https://github.com/0neGal/viper/).

The Github repo for the Flathub version can be found here: https://github.com/flathub/com.github._0negal.Viper/

The initial PR to get it added to Flathub is here: https://github.com/flathub/flathub/pull/2991

GeckoEidechse:
Actually, thinking about it, just unpacking the tar archive inside the Flatpak sandbox is probably a better idea to get closer to native.

Lunchtimeme:
So I have a Steam Deck and I'd also like to be able to use r2modman on it (in gaming mode) so that's +1 from me.
In the meantime is there a workaround I could do myself to make the appimage launch through Steam?

wash2:
> Actually, thinking about it, just unpacking the tar archive inside the Flatpak sandbox is probably a better idea to get closer to native.

I more or less copied what you did for viper [here](https://github.com/wash2/com.github.ebkr.r2modman), and after building and installing as a user r2modman launches, and i can select mods for risk of rain, but it can't seem to launch steam and fails with `You are missing the following 32-bit libraries, and Steam may not run: libc.so.6`. I also had to explicitly give access to `home/.var/app/com.valvesoftware.Steam` in addition to `filesystem=home` to see the flatpak installation of steam along with its games. Did you ever have the same issue?

wash2:
[...]

PR #798, opened 2022-06-20
feat: flatpak support #729

wash2:
This PR adds support for steam as a flatpak, and also for r2modmanplus to be packaged as a flatpak but detecting if it is a flatpak and launching steam using flatpak-spawn. 

I tested that user installation of flatpak steam as a user works, both user + system installation of r2modmanplus as a user works, and that they work together. I also tested that other use cases didn't break to an extent. I did not test system installations of steam as a flatpak though which might need to be tested.

If users have a more custom setup, they may need to make adjustments to flatpak permissions with flatseal.

On an unrelated note, is there a way to launch r2modman such that auto-updates are disabled? I think that might also be a good idea for the flatpak, which manages updates.


Shupsta:
Is there a way I can help test this? I can try to run it on my Steam Deck which is I know would help a lot of people.

PR #800, opened 2022-06-22
Add convenience methods for ThunderstoreCombo

anttimaki:


anttimaki:
They would've been used in a PR for `thunderstore-mod-manager`. I'll move this PR to a fork under `thunderstore-io` org.

PR #808, opened 2022-07-04
DRAFT: Changes from Thunderstore Mod Manager

MythicManiac:
The Thunderstore Mod Manager fork has some changes that would be a good idea to eventually merge here, and before doing so should also be reviewed.

Keeping this PR as draft until the whole feature set we're working on is complete, but it seems like a good idea to open it for comments already. It's also possible to do partial merges earlier in the process if it makes sense.


src/components/navigation/NavigationMenu.vue
```diff
@@ -19,11 +19,18 @@
...
-                    <li><a href="#" @click="launchModded"><i class="fas fa-play-circle icon--margin-right"/>Start modded</a>
+                    <li v-if="canShowServerList">
+                        <a href="#" data-ref="serverlist" @click="emitClick($event.target)"
+                           class="tagged-link" :class="[view === 'serverlist' ? 'is-active' : '']">
+                            <i class="fas fa-server tagged-link__icon icon--margin-right" data-ref="serverlist" @click.prevent.stop="emitClick($event.target)"/>
+                            <span class="tagged-link__content margin-right margin-right--half-width" data-ref="serverlist" @click.prevent.stop="emitClick($event.target)">Server List</span>
+                        </a>
+                    </li>
+                    <li>
+                        <a href="#" @click="launch(true)"><i class="fas fa-play-circle icon--margin-right"/>Start modded</a>
```
ebkr, line 30: We should be avoiding passing what's essentially a magic value where possible. A better approach would be to either pass an enum (EG: LaunchType.MODDED), or to keep the multiple methods separation. 

Removing the method separation should be alright now because it of the argument resolvers. Originally they behaved differently.

src/providers/components/loaders/ServerListProvider.ts
```diff
@@ -0,0 +1,10 @@
+export default class ServerListProvider {
```
ebkr, line 1: These usually throw errors if they're accessed and not defined. I think it'd be best to follow that format.

They're usually for explicit on-screen warnings so that they don't fail silently for the user, and so you don't need to run checks to see if they're defined in other areas. In theory we should always have a provider set.

src/utils/LaunchUtils.ts
```diff
@@ -0,0 +1,70 @@
+import Profile from "../model/Profile";
+import R2Error from "../model/errors/R2Error";
+import Game from "../model/game/Game";
+import FsProvider from "../providers/generic/file/FsProvider";
+import GameRunnerProvider from "../providers/generic/game/GameRunnerProvider";
+import GameDirectoryResolverProvider from "../providers/ror2/game/GameDirectoryResolverProvider";
+import ManagerSettings from "../r2mm/manager/ManagerSettings";
+import ModLinker from "../r2mm/manager/ModLinker";
+
+export const launch = async (game: Game, profile: Profile, modded: boolean): Promise<void> => {
```
ebkr, line 10: I'd recommend, with @MythicManiac to approve, that these are moved into a class and defined as public static methods.

The reasoning behind this is that when used, these look like they'd be defined within the same file as a method, whereas these are actually just imported. We can also tell the exact area where the action is coming from without having to refer back to the import statement.

Another useful thing is that when it's class based, we can likely mock it in tests pretty easily.

PR #822, opened 2022-07-26
Fix elements peaking out above sticky-top

benediktwerner:
Screenshot 2022-07-26 192348

Not sure why this happens, it's kinda weird because using dev tools it appears that the sticky-top is really at the top but for some reason, the games below still peak through. I suspect Windows upscaling is messing something up weirdly because it doesn't happen on my 1080p monitor which doesn't have upscaling.

Also, not sure if there's a better fix but this works and shouldn't make a perceptible difference (probably not even any difference at all on 1080p monitors, at least I didn't notice one when toggling ig back and forth via dev tools).

### DIFF
--- a/src/css/custom.scss
+++ b/src/css/custom.scss
@@ -77,7 +77,7 @@ select {
 
 .sticky-top {
     position: sticky;
-    top: 0;
+    top: -0.1px;
     z-index: 100;
     padding: 1em;
     box-shadow: 1px 1px 3px 0 $scheme-shadow;


benediktwerner:
It would only cause issues at the bottom if there's another sticky element positioned absolutely below it which doesn't seem like good idea. Are there any such elements? The logical layout would be to put all the sticky elements together into a parent that's `sticky-top`. Using a 1px solid border at the top and then moving it up 1px also works but then you need to know the bg color.

MythicManiac:
Well, the currently proposed solution is probably fine, it just feels a bit off to offset the content upwards without creating the same amount of extra height for the element, but I'm unsure what would be the cleanest way to achieve both.

PR #849, opened 2022-10-10
Refactor GameSelectionScreen.created()

anttimaki:


src/migrations/FolderMigration.ts
```diff
@@ -1,25 +1,37 @@
...
-        console.log("Started migration");
+    public static async runMigration() {
+        if (!await this.needsMigration()) {
+            return;
+        }
+
+        console.log("Started legacy directory migration");
         const fs = FsProvider.instance;
+        const rorPath = path.join(PathResolver.ROOT, "RiskOfRain2");
         await CacheUtil.clean();
```
anttimaki, line 27: I was wondering why is cache clean run as a part of the migration? Not that it matters that much, I guess.

src/migrations/FolderMigration.ts
```diff
@@ -1,25 +1,37 @@
...
+
+        if (await fs.exists(rorPath)) {
+            await FileUtils.emptyDirectory(rorPath);
+            await fs.rmdir(rorPath);
         }
-        await fs.rename(path.join(PathResolver.ROOT, "mods"), path.join(PathResolver.ROOT, "RiskOfRain2"));
-    }
 
+        await fs.rename(path.join(PathResolver.ROOT, "mods"), rorPath);
+        console.log("Directory migration done");
```
anttimaki, line 35: What's the policy with console.logging in this project? My initial reaction was to clean them out. Then I thought they might help users to debug stuff. But then again, they don't have access to the console in production, do they?

src/pages/GameSelectionScreen.vue
```diff
@@ -385,61 +385,31 @@ export default class GameSelectionScreen extends Vue {
...
-
-                        this.proceed();
-                        return;
-                    }
-                }
-            });
-        })
+
+        try {
+            await FolderMigration.runMigration();
```
anttimaki, line 392: I considered moving the migration to e.g. App.vue so it wouldn't get checked every time this view mounts (which can happen multiple times now that the app doesn't restart upon game change). But `this.runningMigration` is now used to prevent game selection when the migration is running, and handling that might become complicated if the migration is run in the App. Any thoughts where this should be called (or if it's still needed at all)?

src/r2mm/manager/ManagerSettings.ts
```diff
@@ -27,6 +28,17 @@ export default class ManagerSettings {
         return this.LOADED_SETTINGS;
     }
 
+    /**
+     * Return default game selection needed to skip the game selection screen.
+     */
+    public static async getDefaults(settings?: ManagerSettings) {
```
anttimaki, line 34: Is it ok to import GameManager in ManagerSettings, or should the files be kept loosely coupled? I personally think the helper methods fits nicely here, but it could be moved to another file that imports both GameManager and ManagerSettings, I guess?

anttimaki:
Left some questions as comments. Unfortunately they're not visible here, but they can be seen in the "Files Changed" tab.

src/migrations/FolderMigration.ts
ebkr: Originally it copy/pasted the directory, however there were reports that it was failing silently in some cases causing loss of profiles.

It got reworked to rename instead and that was just leftover.

Doesn't hurt to leave in, however no preference if you want to remove it or not

src/migrations/FolderMigration.ts
ebkr: Logging is accessible during production. `Ctrl + Shift + I` for r2modman, and accessible via logs in TMM.

Generally the rule I've mostly followed is that if a user can't see it, we should still be able to track progress via logs.

src/pages/GameSelectionScreen.vue
ebkr: Could belong to Vuex? If not yet ran, run and set state.

src/r2mm/manager/ManagerSettings.ts
ebkr: I'd say no mostly because then we have a two-way binding on the files. It'd be nice if we could keep settings pretty independent of everything.

What does this do when there's no defaultGame?

src/migrations/FolderMigration.ts
anttimaki: I guess I'll leave it in just in case, to avoid any regressions.

src/pages/GameSelectionScreen.vue
anttimaki: I'll look into it 👍🏻

src/r2mm/manager/ManagerSettings.ts
anttimaki: I'll see if there's a suitable helper file to move this into, and create one if there isn't.

If there's no default game, the returned `defaultGame` and `defaultPlatform` will be undefined (the latter can be undefined independently too). If either is undefined, the calling Vue component will not set `this.selectedGame`/`this.selectedPlatform` nor call `this.proceed` (i.e. works like it did before, I believe).

src/migrations/FolderMigration.ts
anttimaki: So I ended up removing it anyway, because I noticed that it caused unhandled exceptions. [Reading the profile folder](https://github.com/ebkr/r2modmanPlus/blob/develop/src/r2mm/mods/CacheUtil.ts#L20) fails if the migrations are not run yet, because it makes wrong assumptions about the directory structure. I'm _quite_ sure that this is not caused by my changes, but was a pre-existing condition.

src/pages/GameSelectionScreen.vue
anttimaki: This seems to work nicely in the main Vuex store.

src/r2mm/manager/ManagerSettings.ts
anttimaki: Created [ManagerUtils ](https://github.com/ebkr/r2modmanPlus/pull/849/files#diff-5f7b44feef6c321d50703c5cd6d8cce0441680e1cdfde8d25f96082187bccd51)

src/r2mm/manager/SettingsDexieStore.ts
```diff
@@ -206,7 +206,7 @@ export interface ManagerSettingsInterfaceGlobal_V2 {
     favouriteGames: string[] | undefined;
     defaultGame: string | undefined;
     defaultStore: StorePlatform | undefined;
-    gameSelectionViewMode: string | undefined;
+    gameSelectionViewMode: GameSelectionViewMode;
```
MythicManiac, line 209: Shouldn't the new type also allow for `undefined`? I don't see anything in the code changes that guarantees this to never be undefined

src/pages/GameSelectionScreen.vue
```diff
@@ -389,41 +390,20 @@ export default class GameSelectionScreen extends Vue {
...
-            this.favourites = value.getContext().global.favouriteGames || [];
-            if (value.getContext().global.defaultGame !== undefined) {
-                if (value.getContext().global.defaultStore !== undefined) {
-                    const game = GameManager.gameList
-                        .find(value1 => value1.internalFolderName === value.getContext().global.defaultGame)!;
-
-                    const platform = game.storePlatformMetadata.find(value1 => value1.storePlatform === value.getContext().global.defaultStore)!;
+        this.settings = await ManagerSettings.getSingleton(GameManager.unsetGame());
+        const globalSettings = this.settings.getContext().global;
+        this.viewMode = globalSettings.gameSelectionViewMode;
```
MythicManiac, line 395: Continuing the previous comment, this should probably default to `GameSelectionViewMode.CARD` in case of `undefined` value.

MythicManiac (commented):
Also something that came to my mind, did you test if setting a default game and then going back to the game selection screen works as expected? As far as I could tell the changes in this PR keep the overall logic the same, but I'd still check that just to be sure.

src/r2mm/manager/SettingsDexieStore.ts
anttimaki: As far as I can tell there's no code changes needed for such guarantee. Again, *AFAICT*, currently the value can only be initiated by `SettingsDexieStore.createNewSettingsInstance()` and `.mapLegacyToV2()`. To answer your second comment, they both set the value to `GameSelectionViewMode.CARD` so it's already a default value. Once initiated, the value can only be set via `ManagerSettings.setGameSelectionViewMode()`, which doesn't accept `undefined` as a value.

anttimaki:
> did you test if setting a default game and then going back to the game selection screen works as expected

Yes and it does.

MythicManiac (approved):
Good to go as far as I'm concerned

src/utils/ManagerUtils.ts
```diff
@@ -0,0 +1,13 @@
...
+
+
+/**
+ * Return default game selection needed to skip the game selection screen.
+ */
+export const getDefaults = (settings: ManagerSettings) {
+    const globals = settings.getContext().global;
+    const defaultGame = GameManager.findByFolderName(globals.defaultGame);
+    const defaultPlat = defaultGame?.storePlatformMetadata.find(x => x.storePlatform === globals.defaultStore);
+    return {defaultGame, defaultPlatform: defaultPlat?.storePlatform};
```
ebkr, line 12: Could you remove the optional chaining operator for now?

ebkr (changes requested):

src/utils/ManagerUtils.ts
anttimaki: Replaced two occurrences of optional chaining with good old ternary operator.

ebkr (approved):

PR #850, opened 2022-10-11
Remove unnecessary layout spacing

MythicManiac:
Remove unnecessary layout spacing in order to enable more compact layout
to render nicely e.g. on smaller resolutions.

Most of the changes apply to the game select view, which can be compared here:

Old: https://streamable.com/5eezo5
New: https://streamable.com/6cacf0

Or the most notable section, being the very top of the list:
image
image

Note that the hero vertical padding change shown above hasn't been applied to this PR, but otherwise the results are the same


MythicManiac (commented):
@ebkr approved via Discord, so merging

PR #853, opened 2022-10-17
Refactor GameSelectionScreen's "proceed" methods

anttimaki:


MythicManiac (approved):
Still seems to more or less produce the same functionality as before, and the refactors make sense. Looks good to me 👍

ebkr (approved):

PR #856, opened 2022-10-18
PC Game Pass support

starfi5h:
Not sure about the platform naming, feel free to edit it.

### DIFF
--- a/src/model/game/StorePlatform.ts
+++ b/src/model/game/StorePlatform.ts
@@ -4,5 +4,6 @@ export enum StorePlatform {
     EPIC_GAMES_STORE = "Epic Games Store",
     OCULUS_STORE = "Oculus Store",
     ORIGIN = "Origin / EA Desktop",
+    PC_GAME_PASS = "PC Game Pass",
     OTHER = "Other",
 }
--- a/src/model/game/StorePlatform.ts
+++ b/src/model/game/StorePlatform.ts
@@ -4,6 +4,6 @@ export enum StorePlatform {
     EPIC_GAMES_STORE = "Epic Games Store",
     OCULUS_STORE = "Oculus Store",
     ORIGIN = "Origin / EA Desktop",
-    PC_GAME_PASS = "PC Game Pass",
+    XBOX_GAME_PASS = "Xbox Game Pass",
     OTHER = "Other",
 }
--- a/src/providers/generic/game/directory_resolver/win/XGPDirectoryResolver.ts
+++ b/src/providers/generic/game/directory_resolver/win/XGPDirectoryResolver.ts
@@ -1,0 +1,39 @@
+import GameDirectoryResolverProvider from '../../../../ror2/game/GameDirectoryResolverProvider';
+import Game from '../../../../../model/game/Game';
+import R2Error from '../../../../../model/errors/R2Error';
+import ManagerSettings from '../../../../../r2mm/manager/ManagerSettings';
+import FsProvider from '../../../../../providers/generic/file/FsProvider';
+import child from 'child_process';
+
+export default class XGPDirectoryResolver extends GameDirectoryResolverProvider {
+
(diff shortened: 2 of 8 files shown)


src/providers/generic/game/directory_resolver/win/XGPDirectoryResolver.ts
```diff
@@ -0,0 +1,39 @@
+import GameDirectoryResolverProvider from '../../../../ror2/game/GameDirectoryResolverProvider';
+import Game from '../../../../../model/game/Game';
+import R2Error from '../../../../../model/errors/R2Error';
+import ManagerSettings from '../../../../../r2mm/manager/ManagerSettings';
+import FsProvider from '../../../../../providers/generic/file/FsProvider';
+import child from 'child_process';
+
+export default class XGPDirectoryResolver extends GameDirectoryResolverProvider {
```
ebkr, line 8: Minor but I'd rather this be renamed to XboxGamePassDirectoryResolver

The reasoning is that we don't really benefit from condensing the name. It's okay to be a little verbose as it helps understand the context. A random person going through the source code may not know what XGP means.

src/model/game/StorePlatform.ts
```diff
@@ -4,5 +4,6 @@ export enum StorePlatform {
     EPIC_GAMES_STORE = "Epic Games Store",
     OCULUS_STORE = "Oculus Store",
     ORIGIN = "Origin / EA Desktop",
+    PC_GAME_PASS = "PC Game Pass",
```
ebkr, line 7: PC_GAME_PASS should be XBOX_GAME_PASS

src/providers/generic/game/platform_interceptor/PlatformInterceptorImpl.ts
```diff
@@ -79,6 +80,13 @@ const RUNNERS: RunnerType = {
           "darwin": new DirectGameRunner(),
       }
     },
+    [StorePlatform.PC_GAME_PASS]: {
+        [PackageLoader.BEPINEX]: {
+            "win32": new DirectGameRunner(),
```
ebkr, line 85: I don't think these should be calling DirectGameRunner as we want to execute `gamelaunchhelper.exe` right?

A separate runner would be better `XboxGamePassGameRunner` (probably?) which instead of calling the first found executable in the GameManager.ts executables list (like DirectGameRunner does), it'd instead look for and call the `gamelaunchhelper` executable.

We also wouldn't need to declare the other executable names now. Although this is all worth making sure that args are passed through to the game when calling that executable.

Also a minor thing but since game pass isn't available on neither Linux nor MacOS, I'd say you don't need them defined. We can add that at a later date if necessary.

src/providers/generic/game/platform_interceptor/PlatformInterceptorImpl.ts
```diff
@@ -114,6 +122,11 @@ const RESOLVERS: ResolverType = {
         "linux": new DRMFreeDirectoryResolver(),
         "darwin": new DRMFreeDirectoryResolver()
     },
+    [StorePlatform.PC_GAME_PASS]: {
+        "win32": new XGPDirectoryResolver(),
+        "linux": new DRMFreeDirectoryResolver(),
```
ebkr, line 127: Same minor thing here about unnecessary definitions

ebkr (changes requested):

ebkr:
Sorry for the long wait time for getting around to this PR. I'm generally only available during weekends for them

src/model/game/StorePlatform.ts
starfi5h: Although Microsoft renamed "Xbox Game Pass for PC" to "PC Game Pass", I guess Xbox Game Pass is still a more recognizable name for players.

ebkr (approved):
Just tested and this all works perfectly.

Thanks for the contribution!

PR #857, opened 2022-10-23
Split installed and implicitly installed mods

ebkr:
# Purpose
One of the goals are to improve install behaviour by installing only the highest specified version of a dependency.

This is a preliminary step that allows item grouping in to different tabs. Dependencies can easily be located and (in the future) managed from this tab.

# Behaviour
Migration from old profiles will keep all existing mods in the "installed" tab by default. Only new dependencies will be moved to the "dependencies" tab.

Installing a mod with dependencies will place the selected mod into the "installed" tab. All dependencies will be placed in the "dependencies" tab except for dependencies already included in the "installed" tab.

# Future work
- Implement the ability to switch a mod to an "installed" variant
  - May need further work to make this less confusing in terms of terminology for the users. (EG: A dependency is installed, but it's not in the installed tab. Is there a condensed version of "Move to installed tab"?)
- Implement the ability to switch a mod back to a "dependency" variant.
  - This has the effect of allowing the manager to keep a mod updated when necessary should a user no longer wish to keep it updated at the latest version (or stuck at a fixed version).
- Download only the highest listed version of a dependency across all mods listed in the ALL tab.

# Screenshots

## Installed tab
image

## Dependencies tab
image

## All tab
image


src/components/views/DownloadModModal.vue
```diff
@@ -293,9 +295,12 @@ let assignId = 0;
                 }
             }, async (downloadedMods: ThunderstoreCombo[]) => {
                 ProfileModList.requestLock(async () => {
+                    const tsOutdatedMods = outdatedMods.filter(value => !value.getImplicitlyInstalled())
+                        .map(value => ModBridge.getThunderstoreModFromMod(value, tsMods))
+                        .filter(value => value !== undefined) as ThunderstoreMod[];
```
anttimaki, line 300: Something like `.filter((value): value is ThunderstoreMod => value !== undefined);` might work here without explicit type casting. Dunno if it matters much.

src/components/views/LocalModList.vue
```diff
@@ -36,10 +36,22 @@
                 </div>
             </div>
             <div class="tabs">
-                <ul>
-                    <li :class="[{'is-active': activeTab === 0}]" @click="changeTab(0)"><a>All ({{ getTabCountForAll }})</a></li>
-                    <li :class="[{'is-active': activeTab === 1}]" @click="changeTab(1)"><a>Tracked ({{ getTabCountForInstalled }})</a></li>
-                    <li :class="[{'is-active': activeTab === 2}]" @click="changeTab(2)"><a>Dependencies ({{ getTabCountForDependencies }})</a></li>
+                <ul class="margin-bottom">
+                    <li :class="[{'is-active': activeTab === 0}]" @click="changeTab(0)">
```
anttimaki, line 40: IMO `:class` and `@click` should use LocalModTabs enums instead of magic numbers.

anttimaki (approved):
Couple of small comments, no blockers.

src/components/views/LocalModList.vue
ebkr: We can't reference enums directly in here so the only alternative would be a method per option

src/components/views/DownloadModModal.vue
ebkr: Reason for this is that `.filter` returns `ThunderstoreMod[] | undefined` and `!` doesn't appear to resolve the issue. `!` may only apply to nulls rather than undefined?

anttimaki: At least on my IDE using the `value is ThunderstoreMod` type guard is enough to tell TS that this particular `filter` now only returns `ThunderstoreMod`s, making the type cast unnecessary.

src/components/views/LocalModList.vue
anttimaki: If we define something like this in the component:
```
        private activeTab: LocalModTabs = LocalModTabs.INSTALLED;
        private tabs = LocalModTabs;
```

Then the enum options can be accessed in the template via `tabs`. Not sure if this is bad practice for some reason?

MSchmoecker:
One bug/unexpected behaviour is that updating a mod pack moves dependencies to the installed tab, despite them still being dependencies. Only mods that have been changed (removed/added/updated) are left in the dependencies tab and not moved. This happens when using the update button, as well when re-downloading a higher version.

Also, I think this change is the perfect opportunity to remove dependencies when no installed mods are using them anymore. Removing mods in a pack is a real pain at the moment, as every user has to create a new profile or remove the mods manually.

ebkr:
@MSchmoecker Yeah this is just one part of the feature, and auto removing the dependencies are actually planned too.

This is basically opening up the opportunities for:
- Auto resolving dependency versions (and removing)
- Cleaner mod lists
- Allowing modpack updates

None of this will merge into actual develop until the auto resolved dependency versions are implemented. Auto removing is less essential but still good to have. Fortunately we can add that bit later without much impact

src/components/views/DownloadModModal.vue
```diff
@@ -115,6 +115,7 @@ import { Progress } from '../all';
 import Game from '../../model/game/Game';
 import GameManager from '../../model/game/GameManager';
 import ConflictManagementProvider from '../../providers/generic/installing/ConflictManagementProvider';
+import ThunderstorePackages from 'src/r2mm/data/ThunderstorePackages';
```
anttimaki, line 118: This import will break tsmm.

src/components/views/DownloadModModal.vue
```diff
@@ -417,6 +426,18 @@ let assignId = 0;
...
                         }
                     }
+                } else if (modAlreadyInstalled) {
+                    manifestMod.setImplicitlyInstalled(false);
+                    const newModList: ManifestV2[] | R2Error = await ProfileModList.addMod(manifestMod, profile);
+                    if (newModList instanceof R2Error) {
+                        return reject(newModList);
+                    }
+                    if (!modAlreadyInstalled.isEnabled()) {
+                        await ProfileModList.updateMod(manifestMod, profile, mod => {
```
anttimaki, line 436: Tsmm's build complains that this should be `async mod => {` (like on L423).

anttimaki (changes requested):
- Changes look fine
- Two comments regarding errors when building tsmm with these changes
- While testing r2mm locally noticed this (maybe related to what was already mentioned for updating a mod pack):

1. Create empty profile for Valheim
2. Install `Ready_To_Die_MMO` mod pack. Installed tab will show: All 58, Installed 1, Dependencies 57
3. Install `Ready_To_Die_MMO_Season_2` mod pack. Installed tab will show: All 64, Installed 59, Dependencies 5

PR #860, opened 2022-10-24
Remove duplicate code from install rules

MythicManiac:
Re-use the base bepinex install rules for most games and remove the copypasted duplicate code it replaces.


ebkr (approved):
Looks good and all verified all files removed had the same duplicate rules - any custom were kept.

Not a concern for now but it might be nice to extend the BepInEx base rule to add extension `rules` too so we don't need to duplicate the base bepinex file paths.

~~I'll resolve the codacity stuff~~ or not, apparently there's no errors but it fails scanning 🤔

PR #862, opened 2022-11-01
Add GitHub Actions workflow for building the app

anttimaki:
The workflow attemps to build Electron apps based on the Quasar config. Apps are built on Windows, Linux and Mac platforms. Hopefully this will catch changes that break things before they're merged to master/develop branches.

Refs TS-1218

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,61 @@
+name: Build
+
+on: [push]
+
+jobs:
+  build:
+    name: Build for ${{ matrix.platform }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - kind: linux
+            os: ubuntu-latest
+            platform: linux
+          - kind: windows
+            os: windows-latest
+            platform: win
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,61 @@
...
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y libarchive-tools
+
+      - name: Build project
+        id: build
+        uses: StarUbiquitous/command-output@v1.0.1 # Store stdout/stderr to outputs.
+        # You might want to turn this on if there's problems with electron-builder.
+        # env:
+        #   DEBUG: electron-builder
```
ebkr, line 46: What do we gain by having this off?
Is it storage related? Faster builds?

ebkr (approved):
One minor comment but otherwise looks good and appears to work

.github/workflows/build.yml
anttimaki: It was just quite verbose, outputting some 600 lines of debug info. I thought it might actually make it more difficult to spot the actual problem amidst all the noise.

PR #868, opened 2022-11-14
Add standalone GameRunningModal

anttimaki:
The modal was originally built in the navigation menu since that's where the games are launched from, but decoupling these two will hopefully make it easier to customize the NavigationMenu in different views and mod manager variants.

Since the modal and the NavigationMenu are no longer a direct parent- child relation, the status of modal visibity is stored in Vuex store. A separate store module was created for this purpose - the idea is that in the future it could be used to manage more modals than just the one it currently does.

### DIFF
--- a/src/components/modals/GameRunningModal.vue
+++ b/src/components/modals/GameRunningModal.vue
@@ -1,0 +1,40 @@
+<template>
+    <div id="gameRunningModal" :class="['modal', {'is-active': isOpen}]">
+        <div class="modal-background" @click="close"></div>
+        <div class="modal-content">
+            <div class='notification is-info'>
+                <h3 class="title" v-if="isSteamGame">{{ activeGame.displayName }} is launching via Steam</h3>
+                <h3 class="title" v-else>{{ activeGame.displayName }} is starting</h3>
+                <h5 class="title is-5">Close this message to continue modding.</h5>
+                <div v-if="isSteamGame">
+                    <p>If this is taking a while, it's likely due to Steam starting.</p>
+                    <p>Please be patient, and have fun!</p>
+                </div>
+            </div>
+        </div>
+        <button class="modal-close is-large" aria-label="close" @click="close"></button>
+    </div>
+</template>
+
+<script lang="ts">
+import { Component, Prop, Vue } from "vue-property-decorator";
+
+import Game from "../../model/game/Game";
+import { StorePlatform } from "../../model/game/StorePlatform";
+
+@Component
+export default class GameRunningModal extends Vue {
+    @Prop({required: true})
+    readonly activeGame!: Game;
+
+    isSteamGame = this.activeGame.activePlatform.storePlatform === StorePlatform.STEAM;
+
+    close() {
+        this.$store.commit('closeGameRunningModal');
+    }
+
+    get isOpen(): boolean {
+        return this.$store.state.modals.isGameRunningModalOpen;
+    }
+}
+</script>
--- a/src/components/navigation/NavigationMenu.vue
+++ b/src/components/navigation/NavigationMenu.vue
@@ -1,20 +1,5 @@
 <template>
     <div class="full-height">
-        <div id='gameRunningModal' :class="['modal', {'is-active':(gameRunning !== false)}]">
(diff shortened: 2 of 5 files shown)


src/components/modals/GameRunningModal.vue
```diff
@@ -0,0 +1,38 @@
...
+import GameManager from "../../model/game/GameManager";
+import { StorePlatform } from "../../model/game/StorePlatform";
+
+@Component
+export default class GameRunningModal extends Vue {
+    activeGame = GameManager.activeGame;
+    isSteamGame = this.activeGame.activePlatform.storePlatform === StorePlatform.STEAM;
+
+    close() {
+        this.$store.commit('closeModals');
```
ebkr, line 31: Nit but can we rename this event to `closeGameRunningModal`
Feels like we shouldn't be closing all in one go and instead should close whichever has the priority overlay. (Thinking errors vs game running modals sharing the same event)

src/components/modals/GameRunningModal.vue
```diff
@@ -0,0 +1,38 @@
...
+
+<script lang="ts">
+import { Component, Vue } from "vue-property-decorator";
+
+import GameManager from "../../model/game/GameManager";
+import { StorePlatform } from "../../model/game/StorePlatform";
+
+@Component
+export default class GameRunningModal extends Vue {
+    activeGame = GameManager.activeGame;
```
ebkr, line 27: Does this change whenever the game is changed via settings?
As far as I'm aware it didn't and that's why binds are in created hooks instead, could be wrong though.

src/store/ModalsModule.ts
```diff
@@ -0,0 +1,20 @@
+interface State {
+    isGameRunningModalOpen: boolean;
+}
+
+export default {
```
ebkr, line 5: Potentially nicer to be under `/src/store/modules/ModalsModule.ts

src/components/modals/GameRunningModal.vue
anttimaki: I agree, on the hindsight closing all modals at once doesn't sound that appealing. Renamed the commit name.

src/components/modals/GameRunningModal.vue
anttimaki: It does change, since the modal gets recreated/remounted every time the user travels to `GameSelectionScreen` and returns to `Manager` screen, and when it's created the value of `activeGame` has already been updated by other components. If we ever change it so the game can be reselected directly from the `Manager` screen, we probably need to rethink this as well.

src/store/ModalsModule.ts
anttimaki: Moved to subdir.

anttimaki:
Btw should something be done about the Codacy code analysis failing? It says

> It seems like the branch navigation-menu-refactor is not being analyzed. To review this pull request[ enable the branch in the repository settings](https://app.codacy.com/gh/ebkr/r2modmanPlus/settings/branches).

Would this issue get automatically resolved once the earlier PR in the chain is merged and this points to `develop` branch?

src/components/navigation/NavigationMenu.vue
```diff
@@ -139,11 +122,11 @@ import {
                     await linkProfileFiles(this.activeGame, this.contextProfile!);
                 }
 
-                this.gameRunning = true;
+                this.$store.commit("openGameRunningModal");
                 await launch(this.activeGame, this.contextProfile!, mode);
             } catch (error) {
                 if (error instanceof R2Error) {
-                    this.gameRunning = false;
+                    this.$store.commit("closeModals");
```
MythicManiac, line 129: This should probably be calling `closeGameRunningModal` now that it was renamed

src/components/modals/GameRunningModal.vue
MythicManiac: We could also decouple the component entirely from the `GameManager` and simply pass in the active game name as a prop to the component. I'd feel most comfortable with that solution as there's less coupling that way.

MythicManiac (changes requested):
Seems good overall, but spotted a bug

MythicManiac (approved):
Looks good to me 👍

PR #869, opened 2022-11-17
Change profile codes to rely on the Thunderstore API

Oksamies:
I can in no way be held responsible for the quality of this stuff. Forgive me emperor I've become a heretic.

That said, I just kinda slapped together what I think could work and everything seems to work on my dev env. Do let me know if there is something I've missed.

### DIFF
--- a/src/model/errors/R2Error.ts
+++ b/src/model/errors/R2Error.ts
@@ -5,11 +5,24 @@ export default class R2Error extends Error {
     public solution: string;
     public errorReferenceString: string | undefined;
 
-    public constructor(name: string, message: string, solution: string | null) {
+    public constructor(name: string, message: string, solution: string | null = null) {
         super(message);
         this.name = name;
         this.message = message;
         this.solution = solution || '';
         Object.setPrototypeOf(this, R2Error.prototype);
+    }
+
+    public static fromThrownValue(
+        error: R2Error | Error | unknown,
+        defaultName: string = "An unhandled error occurred",
+    ): R2Error {
+        if (error instanceof R2Error) {
+            return error
+        } else if (error instanceof Error) {
+            return new R2Error(defaultName, error.message);
+        } else {
+            return new R2Error(defaultName, `${error}`);
+        }
     }
 }
--- a/src/pages/Profiles.vue
+++ b/src/pages/Profiles.vue
@@ -246,7 +246,6 @@ import Component from 'vue-class-component';
 import { Hero, Progress } from '../components/all';
 import sanitize from 'sanitize-filename';
 import ZipProvider from '../providers/generic/zip/ZipProvider';
-import Axios from 'axios';
 
 import Profile from '../model/Profile';
 import VersionNumber from '../model/VersionNumber';
@@ -274,6 +273,7 @@ import ManagerInformation from '../_managerinf/ManagerInformation';
 import GameDirectoryResolverProvider from '../providers/ror2/game/GameDirectoryResolverProvider';
 import GameManager from '../model/game/GameManager';
 import Game from '../model/game/Game';
+import { ProfileApiClient } from '../r2mm/profiles/ProfilesClient';
 
 let settings: ManagerSettings;
 let fs: FsProvider;
@@ -478,22 +478,24 @@ export default class Profiles extends Vue {
         });
     }
 
-    importProfileUsingCode() {
-        Axios.get(`https://r2modman-hastebin.herokuapp.com/raw/${this.profileImportCode}`)
-            .then(resp => resp.data)
-            .then(async resp => {
-                if (resp.startsWith("#r2modman")) {
-                    const buf = Buffer.from(resp.substring(9).trim(), 'base64');
-                    await FileUtils.ensureDirectory(path.join(PathResolver.ROOT, '_import_cache'));
-                    await fs.writeFile(path.join(PathResolver.ROOT, '_import_cache', 'import.r2z'), buf);
-                    await this.importProfileHandler([path.join(PathResolver.ROOT, '_import_cache', 'import.r2z')]);
-                } else {
-                        throw new Error('Code invalid, no profile is associated with this code');
-                }
-            }).catch(e => {
-                const err = new R2Error('Failed to find profile', e.message, null);
-                this.showError(err);
-            })
+    async parseProfileResponse(profileData: string) {
+        if (profileData.startsWith("#r2modman")) {
+            const buf = Buffer.from(profileData.substring(9).trim(), 'base64');
+            await FileUtils.ensureDirectory(path.join(PathResolver.ROOT, '_import_cache'));
+            await fs.writeFile(path.join(PathResolver.ROOT, '_import_cache', 'import.r2z'), buf);
+            await this.importProfileHandler([path.join(PathResolver.ROOT, '_import_cache', 'import.r2z')]);
+        } else {
+            throw new Error('Code invalid, no profile is associated with this code');
+        }
+    }
+
+    async importProfileUsingCode() {
+        try {
+            const resp = await ProfileApiClient.getProfile(this.profileImportCode);
+            await this.parseProfileResponse(resp.data);
+        } catch (e: any) {
+            this.showError(R2Error.fromThrownValue(e, "Failed to find profile"));
+        }
     }
 
     async importProfileHandler(files: string[] | null) {
--- a/src/r2mm/mods/ProfileModList.ts
+++ b/src/r2mm/mods/ProfileModList.ts
@@ -13,7 +13,6 @@ import ExportFormat from '../../model/exports/ExportFormat';
 import ExportMod from '../../model/exports/ExportMod';
 import PathResolver from '../manager/PathResolver';
 import ZipProvider from '../../providers/generic/zip/ZipProvider';
-import Axios from 'axios';
 import FileUtils from '../../utils/FileUtils';
 import ManagerInformation from '../../_managerinf/ManagerInformation';
 import LinkProvider from '../../providers/components/LinkProvider';
@@ -23,6 +22,7 @@ import { MOD_LOADER_VARIANTS } from '../installing/profile_installers/ModLoaderV
 import FileTree from '../../model/file/FileTree';
 import ZipBuilder from '../../providers/generic/zip/ZipBuilder';
 import InteractionProvider from '../../providers/ror2/system/InteractionProvider';
+import { ProfileApiClient } from '../profiles/ProfilesClient';
 
 export default class ProfileModList {
 
@@ -249,12 +249,12 @@ export default class ProfileModList {
         } else {
             await exportBuilder.createZip(exportPath);
             const profileBuffer = '#r2modman\n' + (await fs.base64FromZip(exportPath));
-            Axios.post('https://r2modman-hastebin.herokuapp.com/documents', profileBuffer)
-                .then(resp => callback(resp.data.key, null))
-                .catch(e => {
-                    const err: Error = e;
-                    callback('', new R2Error('Failed to export profile', err.message, null));
-                });
+            try {
+                const storageResponse = await ProfileApiClient.createProfile(profileBuffer);
+                callback(storageResponse.data.key, null);
+            } catch (e: R2Error | unknown) {
+                callback('', R2Error.fromThrownValue(e, "Failed to export profile"));
+            }
         }
     }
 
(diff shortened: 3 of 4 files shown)


Oksamies:
> This change doesn't actually match what was specced

Alright, now the changes should reflect the specs pretty good. I didn't separate the changes into a new file since the there are not that much changes and on top of that the wrappers were not needed, cause Axios gives the Promise straight away 👍

src/r2mm/profiles/ProfilesClient.ts
```diff
@@ -0,0 +1,34 @@
+import Axios, { AxiosResponse } from 'axios';
+import R2Error from 'src/model/errors/R2Error';
+
+export default class ProfilesClient {
```
MythicManiac, line 4: There's no need to use a class here since the file itself is already handling the encapsulation. Classes can be instantiated which doesn't really make sense here.

Also generally speaking, default exports are worse than named exports, so use them instead unless default exports are explicitly required (like with NextJS)

src/r2mm/profiles/ProfilesClient.ts
```diff
@@ -0,0 +1,34 @@
...
+                endpoint,
+                payload,
+                {
+                    headers: {'Content-Type': 'application/octet-stream'}
+                }
+            )
+        } catch (e: any) {
+            if (Axios.isAxiosError(e)) {
+                if (e.response!.status == 429) {
+                    return new R2Error('Too many exports in a short period of time', ':PeepoShake:', 'Slow down there champ');
```
MythicManiac, line 21: `:PeepoShake:`?

MythicManiac, line 21: We should probably just throw these rather than returning them, and add a try/catch to the call site

src/r2mm/profiles/ProfilesClient.ts
```diff
@@ -0,0 +1,34 @@
...
+                } else {
+                    return new R2Error('Failed to export profile', e.message, `${e.response!.status}`);
+                }
+            } else {
+                return new R2Error('Unexpected error', e.message, null);
```
MythicManiac, line 26: And since we're throwing them, you can just throw the original error here if we don't do any special handling for it and let the call site deal with it.

src/r2mm/profiles/ProfilesClient.ts
```diff
@@ -0,0 +1,34 @@
...
+            }
+        }
+    }
+
+    public static async retrieveProfile(endpoint: string): Promise<AxiosResponse> {
+        return await Axios.get(endpoint)
```
MythicManiac, line 32: Semicolon missing, but also you can remove `async` from the function and `await` from this line and it'll work exactly as it does now (since Axios already returns a `Promise<AxiosResponse>` and you can just pass it forward). Semicolon also missing.

src/r2mm/profiles/ProfilesClient.ts
```diff
@@ -0,0 +1,34 @@
+import Axios, { AxiosResponse } from 'axios';
+import R2Error from 'src/model/errors/R2Error';
+
+export default class ProfilesClient {
+    public static TSCreateAPI = 'https://thunderstore.io/api/experimental/legacyprofile/create/'
+    public static TSRetrieveAPI = (profileImportCode: string) => {return `https://thunderstore.io/api/experimental/legacyprofile/get/${profileImportCode}/`}
+    public static HerokuRetrieveAPI = (profileImportCode: string) => {return `https://r2modman-hastebin.herokuapp.com/raw/${profileImportCode}`}
+
+    public static async sendProfile(endpoint: string, payload: string): Promise<AxiosResponse | R2Error> {
```
MythicManiac, line 9: Make sure to remove `| R2Error` from this type once you change the error handling to use throws instead of returns.

src/r2mm/profiles/ProfilesClient.ts
```diff
@@ -0,0 +1,34 @@
+import Axios, { AxiosResponse } from 'axios';
+import R2Error from 'src/model/errors/R2Error';
+
+export default class ProfilesClient {
+    public static TSCreateAPI = 'https://thunderstore.io/api/experimental/legacyprofile/create/'
+    public static TSRetrieveAPI = (profileImportCode: string) => {return `https://thunderstore.io/api/experimental/legacyprofile/get/${profileImportCode}/`}
+    public static HerokuRetrieveAPI = (profileImportCode: string) => {return `https://r2modman-hastebin.herokuapp.com/raw/${profileImportCode}`}
```
MythicManiac, line 7: You don't need to use return here at all if you remove the `{}`. Semicolons missing

src/r2mm/profiles/ProfilesClient.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    public static async sendProfile(endpoint: string, payload: string): Promise<AxiosResponse | R2Error> {
+        try {
+            return await Axios.post(
+                endpoint,
+                payload,
+                {
+                    headers: {'Content-Type': 'application/octet-stream'}
+                }
+            )
```
MythicManiac, line 17: semicolon missing

src/r2mm/profiles/ProfilesClient.ts
MythicManiac, line 7: These could be consts when not in a class

src/r2mm/mods/ProfileModList.ts
```diff
@@ -248,13 +249,12 @@ export default class ProfileModList {
...
         } else {
             await exportBuilder.createZip(exportPath);
-            const profileBuffer = '#r2modman\n' + (await fs.base64FromZip(exportPath));
-            Axios.post('https://r2modman-hastebin.herokuapp.com/documents', profileBuffer)
-                .then(resp => callback(resp.data.key, null))
-                .catch(e => {
-                    const err: Error = e;
-                    callback('', new R2Error('Failed to export profile', err.message, null));
-                });
+            const storage_response_or_error = await ProfilesClient.sendProfile(ProfilesClient.TSCreateAPI, '#r2modman\n' + (await fs.base64FromZip(exportPath)))
```
MythicManiac, line 252: We're not writing python so `camelCase` is preferred.

MythicManiac (changes requested):
Better but not quite there.

To summarize:
- Move as much logic away from the vue components as possible. It's likely at least a part of the error handling will have to remain in the components however.
- Add missing semicolons
- Use proper casing with variable names
- Some other refactoring suggestions

src/pages/Profiles.vue
```diff
@@ -478,22 +479,30 @@ export default class Profiles extends Vue {
...
-                if (resp.startsWith("#r2modman")) {
-                    const buf = Buffer.from(resp.substring(9).trim(), 'base64');
-                    await FileUtils.ensureDirectory(path.join(PathResolver.ROOT, '_import_cache'));
-                    await fs.writeFile(path.join(PathResolver.ROOT, '_import_cache', 'import.r2z'), buf);
-                    await this.importProfileHandler([path.join(PathResolver.ROOT, '_import_cache', 'import.r2z')]);
-                } else {
-                        throw new Error('Code invalid, no profile is associated with this code');
-                }
-            }).catch(e => {
+    async parseProfileResponse(response_data: string) {
```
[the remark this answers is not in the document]
Oksamies: With a quick glance, seems like doing this would mean splitting right about the whole file down into separate modules. I'll leave it out of this PR, but I'll drag it through the post-mortem and wrap-up so we'll have task for refactoring the mod manager code into the modules; so that it can be used more easily in the future.

src/pages/Profiles.vue
```diff
@@ -478,22 +479,30 @@ export default class Profiles extends Vue {
...
+        } catch (e: any) {
+            try {
+                const resp = await ProfilesClient.retrieveProfile(ProfilesClient.HerokuRetrieveAPI(this.profileImportCode))
+                this.parseProfileResponse(resp.data)
+            } catch (e: any) {
                 const err = new R2Error('Failed to find profile', e.message, null);
                 this.showError(err);
-            })
+            }
+        }
```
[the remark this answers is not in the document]
Oksamies: I kind of cant move anything from `importProfileUsingCode` to the `ProfileClient.ts` module, without dragging the whole `Profiles.vue` file with it. Or might be that I don't know how to. Anyways going bit out of scope, from getting the heroku fix done.

I suggest we move forward with this (if there are no big issues with this) and move the "refactoring" to the next milestone.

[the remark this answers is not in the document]
MythicManiac: As mentioned on Discord, as long as the API endpoint fallback logic is in the new module this is fine, so the bare minimum is to move that portion of the logic in the API module.

On another note, don't try catch but simply check if the first returns a 404 status code and use that as a signal to test the fallback. Falling back to the old API if the new API is returning some other error than 404 is wrong behavior and could lead to some edge case bugs (not really in this case but at l east invalid error messages). In fact you probably shouldn't even use try/catch in the API module at all, just check for status codes and return the data as is.

anttimaki:
Well I don't have too much to add to what Mythic already mentioned. Mildly annoyed by the code style things (missing semicolons, mixing single and double quotes). I guess those should be handled by a linter but in the meanwhile wouldn't hurt to be meticulous about them.

src/r2mm/profiles/ProfilesClient.ts
ebkr: The general pattern so far has been to use default exports as most things are classes. Named exports tend to be type declarations.

Reasoning is that it's easier to remember a class name than it is to remember a method name which ends up being more productive.

src/r2mm/profiles/ProfilesClient.ts
```diff
@@ -0,0 +1,33 @@
...
+        payload,
+        {
+            headers: {'Content-Type': 'application/octet-stream'}
+        }
+    );
+}
+
+export async function retrieveProfile(profileImportCode: string): Promise<AxiosResponse> {
+    const instance = Axios.create({
+        validateStatus: function(status) {
```
ebkr, line 19: Is this needed?

ebkr (changes requested):

ebkr:
@anttimaki The project used to have ESLint however it was incredibly slow.
It would freeze up VSC for ages, so it was disabled.

src/r2mm/profiles/ProfilesClient.ts
Oksamies: Yes if we don't want to use try catch (mythics suggestion in comments above). Without this Axios will throw an AxiosError and the latter code wont be run. Now that I think about it, if Axios gets something else than a 404 like 500, it probably wont work. I'll change it back to the try catch since it's more reliable.

MythicManiac: Try catch is fine to use but the point was not everything has to be wrapped in it since you can simply let the promise fail if something unexpected happens.

src/pages/Profiles.vue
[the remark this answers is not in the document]
MythicManiac: Opened #883 regarding this

ebkr (approved):
Minor comments but approved to not block once fixed

src/model/errors/R2Error.ts
```diff
@@ -5,11 +5,24 @@ export default class R2Error extends Error {
     public solution: string;
     public errorReferenceString: string | undefined;
 
-    public constructor(name: string, message: string, solution: string | null) {
+    public constructor(name: string, message: string, solution: string | null = null) {
```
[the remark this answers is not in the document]
MythicManiac: This was set to null as that's what it's typed as and I don't want to risk breaking something by changing the signature

src/model/errors/R2Error.ts
```diff
@@ -5,11 +5,24 @@ export default class R2Error extends Error {
...
         super(message);
         this.name = name;
         this.message = message;
         this.solution = solution || '';
         Object.setPrototypeOf(this, R2Error.prototype);
     }
+
+    public static FromThrownValue(
```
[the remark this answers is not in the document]
MythicManiac: It was intentionally uppercased to differentiate between static class-scoped methods vs. object instance scoped methods (a fairly common naming convention), but I can be change it if you prefer not to make that distinction visible in naming.

src/model/errors/R2Error.ts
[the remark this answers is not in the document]
MythicManiac: Marking as resolved as discussed on Discord

MythicManiac (approved):

PR #881, opened 2022-11-24
Refactor DownloadModModal

anttimaki:
* Remove the prop that controls the modal visibility. Manage it via Vuex store instead
* Remove the prop that defines which mod to download/update. The value is passed to Vuex store as a part of the open modal command instead
* Remove "update all mods" prop. The value can be derived from whether there's a mod to download/update set or not
* Since the props were removed, there's no longer a need to render duplicate modals in LocalModList and OnlineModList components

The motivation for the refactoring is to support breaking the Manager page into smaller, more manageable pieces, which in turn allows easier customization.

Refs TS-1321

### DIFF
--- a/src/pages/Manager.vue
+++ b/src/pages/Manager.vue
@@ -129,16 +129,8 @@
 
         <CategoryFilterModal :onClose="sortThunderstoreModList" />
         <LocalFileImportModal :visible="importingLocalMod" @close-modal="importingLocalMod = false" @error="showError($event)"/>
-
-        <DownloadModModal
-            :show-download-modal="showUpdateAllModal"
-            :update-all-mods="true"
-            :thunderstore-mod="null"
-            @closed-modal="showUpdateAllModal = false;"
-            @error="showError($event)"
-        />
-
-		<GameRunningModal :activeGame="activeGame" />
+        <DownloadModModal @error="showError($event)" />
+        <GameRunningModal :activeGame="activeGame" />
 
 		<div class='columns' id='content'>
 			<div class="column non-selectable" :class="navbarClass">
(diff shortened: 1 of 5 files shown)


src/pages/Manager.vue
```diff
@@ -938,7 +929,7 @@ import GameRunningModal from '../components/modals/GameRunningModal.vue';
                     this.setAllModsEnabled(false);
                     break;
                 case "UpdateAllMods":
-                    this.showUpdateAllModal = true;
+                    this.$store.commit("openDownloadModModal");
```
MythicManiac, line 932: I'm not a big fan of implicitly assuming the context is to update all mods if no specific mod is set, mostly due to it not being very obvious from the code if you don't check. I'm not sure what would I expect a mod download modal to do if it doesn't get a mod assigned to it, but updating all mods is not it at least.

I understand that logic from the state-machine standpoint, as it ensures the modal can't end up in an invalid state, but I'd suggest exploring some ways to make it more obvious to the developer (naming? idk).

Perhaps using two separate transactions at least, one for opening the modal in update view and one for a specific mod?

Also shouldn't the commit use a namespaced transaction name, i.e. `"modalsModule/openDownloadModModal"` or whatever the namespace was?

MythicManiac (commented):
Clarity improvement suggestions stand, looks good otherwise. Requires rebasing before approval.

src/pages/Manager.vue
anttimaki: Changed to have `openDownloadModModal` for a single mod and `openUpdateAllModsModal` for updating all.

> Also shouldn't the commit use a namespaced transaction name, i.e. "modalsModule/openDownloadModModal" or whatever the namespace was?

Vuex supports choosing whether to namespace submodules on a module basis. In this case I chose not to, since I figured modal management is something that happens globally and the calls aren't limited to a single package or feature. So if the question was "is the lack of namespace a bug", the answer is no. If it was more "I'd like this to be namespaced", it can be done if we think there should be a convention to always use namespaces.

MythicManiac (approved):
Looks good to me 👍

ebkr (approved):

PR #882, opened 2022-11-25
Fix Xbox Game Pass directory resolver bugs

MythicManiac:
Fix a bug in the Xbox Game Pass directory resolver where an if check unintentionally always returned true due to a missing await keyword.

Also modify the error handling block to not fail if something that doesn't have a message attribute attached to it is thrown.

### DIFF
--- a/src/providers/generic/game/directory_resolver/win/XboxGamePassDirectoryResolver.ts
+++ b/src/providers/generic/game/directory_resolver/win/XboxGamePassDirectoryResolver.ts
@@ -17,7 +17,7 @@ export default class XboxGamePassDirectoryResolver extends GameDirectoryResolver
             const installDirectoryQuery = `get-appxpackage -Name ${game.activePlatform.storeIdentifier} | select -expand InstallLocation`;
             const queryResult: string = child.execSync(`powershell.exe "${installDirectoryQuery}"`).toString().trim();
             const realInstallLocation = await FsProvider.instance.realpath(queryResult);
-            if (FsProvider.instance.exists(realInstallLocation)) {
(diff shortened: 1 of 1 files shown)


src/providers/generic/game/directory_resolver/win/XboxGamePassDirectoryResolver.ts
```diff
@@ -26,7 +26,7 @@ export default class XboxGamePassDirectoryResolver extends GameDirectoryResolver
         } catch (err) {
             return new R2Error(
                 `Unable to resolve the ${game.displayName} install directory`,
-                err.message,
+                `${err}`,
```
ebkr, line 29: General pattern is to cast to an `Error` and then pass through the message, so this wasn't inherently wrong. 

What does the output look like if you intentionally throw?

MythicManiac: The reasoning for this change is that it's not guaranteed that all throws are of the `Error` type (or some of its subclass), given that you can throw anything in JS. This is also reflected in the type checking and you can get warnings about it.

As such it's probably better to simply cast it to a string, at best you'll get output like below, at worst you'll probably see `[object Object]`.

image

ebkr (approved):

src/providers/generic/game/directory_resolver/win/XboxGamePassDirectoryResolver.ts
ebkr: Cool, just looks like it prepends "Error:", so this is fine

PR #883, opened 2022-11-25
Refactor profile code download

MythicManiac:
Create a module dedicated for profile import and export related logic
and move parts of the profile code download logic there away from the
Vue component.

Related discussion: https://github.com/ebkr/r2modmanPlus/pull/869#discussion_r1029625254

### DIFF
--- a/src/r2mm/mods/ProfileImportExport.ts
+++ b/src/r2mm/mods/ProfileImportExport.ts
@@ -1,0 +1,59 @@
+import FileUtils from '../../utils/FileUtils';
+import path from 'path';
+import PathResolver from '../../r2mm/manager/PathResolver';
+import FsProvider from '../../providers/generic/file/FsProvider';
+import { ProfileApiClient } from '../../r2mm/profiles/ProfilesClient';
+
+let fs: FsProvider
+
+const IMPORT_CACHE_DIRNAME = "_import_cache";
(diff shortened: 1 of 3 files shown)


src/r2mm/mods/ProfileImportExport.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const getImportCacheDir = async (): Promise<string> => {
+    const dir = path.join(PathResolver.ROOT, IMPORT_CACHE_DIRNAME);
+    await FileUtils.ensureDirectory(dir);
+    return dir;
+}
+
+const getDownloadDestination = async (): Promise<string> => {
+    const cacheDir = await getImportCacheDir();
+    // TODO: This really should always generate an unique filename in order to
```
ebkr, line 20: We have a blocking modal so this should(™️) never happen

Worth removing the TODO if there's no set plan to actually resolve this. If there is then I'd say link it to a clickup issue ref. If no ref, we can still keep the comment, just lose the TODO.

ebkr (approved):

src/r2mm/mods/ProfileImportExport.ts
MythicManiac: As discussed on Discord, I'll leave this TODO up since it's descriptive enough & we'll automate issue extraction from TODO comments

PR #879, opened 2022-11-22
Decouple category filter modal

anttimaki:

PR #885, opened 2022-11-25
Decouple category filter modal

MythicManiac:
This PR replaces #879 as some commits in #879 had to be merged ahead of time; this PR is the re-ordered result.

This should be merged after #886

### DIFF
--- a/src/components/modals/CategoryFilterModal.vue
+++ b/src/components/modals/CategoryFilterModal.vue
@@ -1,0 +1,154 @@
+<template>
+    <modal v-show="isOpen" :open="isOpen" :show-close="false">
+        <template v-slot:title>
+            <p class="card-header-title">Filter mod categories</p>
+        </template>
+        <template v-slot:body>
+            <div class="input-group">
+                <label>Categories</label>
+                <select class="select select--content-spacing" @change="selectCategory($event)">
+                    <option selected disabled>
+                        Select a category
+                    </option>
+                    <option v-for="(key, index) in unselectedCategories" :key="`category--${key}-${index}`">
+                        {{ key }}
+                    </option>
+                </select>
+            </div>
+            <br/>
+            <div class="input-group">
+                <label>Selected categories:</label>
+                <div class="field has-addons" v-if="selectedCategories.length > 0">
+                    <div class="control" v-for="(key, index) in selectedCategories" :key="`${key}-${index}`">
+                        <span class="block margin-right">
+                            <a href="#" @click="unselectCategory(key)">
+                                <span class="tags has-addons">
+                                    <span class="tag">{{ key }}</span>
+                                    <span class="tag is-danger">
+                                        <i class="fas fa-times"></i>
+                                    </span>
+                                </span>
+                            </a>
+                        </span>
+                    </div>
+                </div>
+                <div class="field has-addons" v-else>
+                    <span class="tags">
+                        <span class="tag">No categories selected</span>
+                    </span>
+                </div>
+            </div>
+            <hr/>
+            <div>
+                <input
+                    v-model="allowNsfw"
+                    id="nsfwCheckbox"
+                    class="is-checkradio has-background-color"
+                    type="checkbox"
+                    :class="[{'is-dark': !isDarkTheme, 'is-white': isDarkTheme}]"
+                >
+                <label for="nsfwCheckbox">Allow NSFW (potentially explicit) mods</label>
+            </div>
+            <br/>
+            <div>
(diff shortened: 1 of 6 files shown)


src/components/modals/CategoryFilterModal.vue
```diff
@@ -0,0 +1,155 @@
...
+import CategoryFilterMode from "../../model/enums/CategoryFilterMode";
+import GameManager from "../../model/game/GameManager";
+import ManagerSettings from "../../r2mm/manager/ManagerSettings";
+
+@Component({
+    components: { Modal }
+})
+export default class CategoryFilterModal extends Vue {
+    activeGame = GameManager.activeGame;
+    settings: ManagerSettings = new ManagerSettings();
```
MythicManiac, line 90: Does this work as intended, as in, do these get cleared from the cached values when appropriate (e.g. game change)?

src/components/modals/CategoryFilterModal.vue
```diff
@@ -0,0 +1,155 @@
...
+        return Object.values(CategoryFilterMode);
+    }
+
+    close() {
+        this.onClose();
+        this.$store.commit("closeCategoryFilterModal");
+    }
+
+    async created() {
+        this.settings = await ManagerSettings.getSingleton(this.activeGame);
```
MythicManiac, line 121: It looks like `this.activeGame` is always set, is there a reason we don't use it directly on line 90?

Though I'd be more comfortable taking in the active game as a prop rather than relying on global state

MythicManiac (commented):
A few questions, otherwise looks good. Can't approve this myself since I re-opened the PR on behalf of @anttimaki

src/components/modals/CategoryFilterModal.vue
anttimaki: This approach is copied from the Manager page. AFAIK this is just a dummy value so we don't have to check if `settings` is `undefined` in every place where it's called (or use `!` here). The actual value will be overwritten in `created`.

And yes, based on manual testing it works as intended. The whole component will get unmounted/remounted if user wants to change the game. At least for now when the game can only be changed in the `GameSelectionScreen`.

src/components/modals/CategoryFilterModal.vue
anttimaki: I don't think we can call `await` on line 90. But since `this.activeGame` is used only on this line, we can call `GameManager.activeGame` on this line and ditch `this.activeGame`.

I've assumed one of the main reasons the static class/singleton pattern is used widely on this code base is that we can access e.g. the active game from anywhere without having to pass it around. I'm not against passing it as a prop, but if we no longer want to access it via GameManager, I'd suggest maybe moving it into Vuex store.

anttimaki (approved):
Marking approved on behalf of Mythic.

src/components/modals/CategoryFilterModal.vue
```diff
@@ -0,0 +1,154 @@
...
+    close() {
+        this.onClose();
+        this.$store.commit("closeCategoryFilterModal");
+    }
+
+    async created() {
+        this.settings = await ManagerSettings.getSingleton(GameManager.activeGame);
+    }
+
+    // TODO: The value doesn't get updated if the theme is toggled while
```
ebkr, line 123: Might be good to link this TODO to something in clickup

src/store/modules/ModFilterModule.ts
```diff
@@ -0,0 +1,65 @@
...
+                        .flat()
+                )
+            );
+            categories.sort();
+            return categories;
+        },
+
+        unselectedCategories (state, getters) {
+            // TODO: try to find a way to fix typing for getters parameter.
+            const cats: string[] = getters.allCategories;
```
ebkr, line 36: Rename to `categories`
imo there's no reason to shorten this

ebkr (approved):
One minor comment and a suggestion to not have a permanent TODO. Happy to merge once resolved.

src/components/modals/CategoryFilterModal.vue
MythicManiac: Removed this comment entirely since I don't think there's any way to toggle the theme while this modal is open, but if you feel like this should be tracked as a TODO let me know

anttimaki: @MythicManiac While the modal isn't open, it's still mounted when user changes the color theme via settings, meaning the value won't get updated until user leaves and returns to the manager page. (And to clarify, the problem existed and was in fact worse before the changes in this PR.)

Also noticed you removed another TODO comment. Is this to say we should avoid TODO comments in this code base?

PR #895, opened 2022-11-29
Suppress error the first time TS modlist fails to refresh in background

anttimaki:
Thunderstore's sentry receives error reports where the background update process has run without problems for a while and then fails to read the mod list from API with HTTP status code 0.

The assumption is that it's caused by users closing the window while the update process is in progress. To test this assumption, suppress the error the first time the update fails. If the next update fails too, throw the error normally.


MythicManiac (approved):
Looks good to me 👍

ebkr (approved):

PR #902, opened 2022-12-05
Deprecated packages hidden by default from the online list

ebkr:
image

image

image

Formatting changes appear to be done by IntelliJ, not overly fussed which convention we use and is mixed in places anyway


MythicManiac (approved):
Looks good to me 👍

anttimaki (approved):
LGTM too.

Also I take it single quotes are preferred for import statements?

PR #908, opened 2022-12-11
Added lodash debouncing to wrap watched method to prevent number of filtering searched

ebkr:


src/pages/Manager.vue
```diff
@@ -362,7 +363,11 @@ import GameRunningModal from '../components/modals/GameRunningModal.vue';
 		}
 
 		@Watch('thunderstoreSearchFilter')
-		onThunderstoreFilterUpdate() {
+        onThunderstoreFilterUpdate = debounce(() => {
```
anttimaki, line 366: Nitpick: could just pass `this.performThunderstoreFilterUpdate` instead of wrapping it into an arrow function.

anttimaki (approved):

MythicManiac (approved):
Looks good to me 👍

If you end up making changes, I'd bump the delay on the debounce from 50ms to 150-300ms just so it has a higher chance to actually catch multiple keystrokes before running. Too high will of course make it feel unresponsive.

ebkr:
Superseded by #934

PR #911, opened 2022-12-12
Ensure settings can be loaded before mounting the app

anttimaki:
For example a corrupted IndexedDB storage may cause loading the settings to fail, which would result in the users seeing an emptyish screen. To prevent that, try to load the settings as the first thing when the app is launched. If that fails, log the error for debug purposes and display an error modal to the user.

Refs TS-1357

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,5 +1,5 @@
 <template>
-    <div id="q-app">
+    <div>
 
         <router-view @error="showError" v-if="visible"/>
 
@@ -90,23 +90,19 @@ export default class App extends mixins(UtilityMixin) {
     }
 
     async created() {
-
         // Use as default game for settings load.
-        const riskOfRain2Game = GameManager.gameList.find(value => value.displayName === "Risk of Rain 2")!;
-        GameManager.activeGame = riskOfRain2Game;
+        GameManager.activeGame = GameManager.unsetGame();
 
         this.hookThunderstoreModListRefresh();
         this.hookProfileModListRefresh();
 
-        const settings = await ManagerSettings.getSingleton(riskOfRain2Game);
+        const settings = await ManagerSettings.getSingleton(GameManager.activeGame);
         this.settings = settings;
 
         InstallationRuleApplicator.apply();
         InstallationRules.validate();
 
         ipcRenderer.once('receive-appData-directory', async (_sender: any, appData: string) => {
-
-            await settings.load();
             PathResolver.APPDATA_DIR = path.join(appData, 'r2modmanPlus-local');
             // Legacy path. Needed for migration.
             PathResolver.CONFIG_DIR = path.join(PathResolver.APPDATA_DIR, "config");
@@ -130,7 +126,6 @@ export default class App extends mixins(UtilityMixin) {
             ipcRenderer.once('receive-is-portable', async (_sender: any, isPortable: boolean) => {
                 ManagerInformation.IS_PORTABLE = isPortable;
                 LoggerProvider.instance.Log(LogSeverity.INFO, `Starting manager on version ${ManagerInformation.VERSION.toString()}`);
-                await settings.load();
                 this.visible = true;
             });
             ipcRenderer.send('get-is-portable');
--- a/src/components/SettingsLoader.vue
+++ b/src/components/SettingsLoader.vue
@@ -1,0 +1,92 @@
+<template>
+    <div id="settings-loader">
+        <slot v-if="isLoaded" />
+
+        <div v-else :class="['modal', 'z-top', {'is-active': error}]">
+            <div class="modal-content">
+                <div class="notification is-danger">
+                    <h3 class="title">Error</h3>
+                    <h5 class="title is-5">{{error && error.name}}</h5>
+                    <p>{{error && error.message}}</p>
+                    <div v-if="error && error.solution">
+                        <br/>
+                        <h5 class="title is-5">Suggestion</h5>
+                        <p>{{error && error.solution}}</p>
+                    </div>
+                </div>
+            </div>
+        </div>
+    </div>
+</template>
+
+<script lang="ts">
+import Vue from "vue";
+import { Component, Prop } from "vue-property-decorator";
+
+import R2Error from "../model/errors/R2Error";
+import GameManager from "../model/game/GameManager";
+import ManagerSettings from "../r2mm/manager/ManagerSettings";
+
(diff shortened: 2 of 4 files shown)


MythicManiac (approved):
Left a comment but I'm not considering it a blocker as this should already be an improvement to the current situation.

Interested to see how the error modal would look like in practice however.

src/components/SettingsLoader.vue
```diff
@@ -0,0 +1,84 @@
...
+        }
+        
+        try {
+            settings = await ManagerSettings.getSingleton(defaultGame);
+        } catch (e) {
+            this.handleError("Failed to read ManagerSettings", `${e}`, solution);
+            return;
+        }
+
+        const error = await settings.load();
```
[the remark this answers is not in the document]
anttimaki: Well the whole method is just a try-catch block, but I guess something might go wrong while creating the `R2Error` from the actual error that was raised. Or someone might change it later and this gets outdated.

anttimaki:
Added one more try-catch. The thing in all its beauty is something like this:
image

MythicManiac (approved):
Nice 👍

MythicManiac:
> Is there any reason the user still have to reset their settings manually? May as well do it for them no?

It's a destructive operation, generally makes sense to prompt for confirmation in such scenarios. We're not entirely certain it's the only way to fix things after all, it's just the easy way.

src/App.vue
```diff
@@ -90,23 +90,19 @@ export default class App extends mixins(UtilityMixin) {
     }
 
     async created() {
-
         // Use as default game for settings load.
-        const riskOfRain2Game = GameManager.gameList.find(value => value.displayName === "Risk of Rain 2")!;
-        GameManager.activeGame = riskOfRain2Game;
+        GameManager.activeGame = GameManager.unsetGame();
```
ebkr, line 94: Nice catch

src/components/SettingsLoader.vue
```diff
@@ -0,0 +1,91 @@
...
+    @Prop({required: true})
+    private logError!: (error: R2Error) => void;
+
+    error: R2Error|null = null;
+    isLoaded = false;
+
+    get defaultGame() {
+        const defaultGame = GameManager.unsetGame();
+
+        // Don't trust the non-null asserted typing of .unsetGame().
```
ebkr, line 44: Any particular reason for this? I feel like we can guarantee there will always be one game that the manager supports.
(It'd be a pretty terrible mod manager if it didn't 😅)

src/components/SettingsLoader.vue
```diff
@@ -0,0 +1,91 @@
...
+        return defaultGame;
+    }
+
+    handleError(name: string, message: string, solution: string|null = null) {
+        this.error = new R2Error(name, message, solution);
+        this.logError(this.error);
+    }
+
+    async created() {
+        let defaultGame;
```
ebkr, line 58: Why are we using a computed property if it's only used within this block?

src/components/SettingsLoader.vue
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        try {
+            settings = await ManagerSettings.getSingleton(defaultGame);
+        } catch (e) {
+            this.handleError("Failed to read ManagerSettings", `${e}`, solution);
+            return;
+        }
+
+        try {
```
ebkr, line 76: Could this instead be (or at least similar to):
```ts
settings.load()
  .then(x => {
    if (x instanceof R2Error) {
      throw x;
    }
  }).then(x => {this.isLoaded = true})
  .catch(e => {this.handleError("Failed to load ManagerSettings", `${e}`, solution)})
```
Not a deal breaker, just curious as I'd prefer this approach if possible. Might look weird when displaying though.

ebkr (approved):
Minor comments but happy to approve

src/components/SettingsLoader.vue
MythicManiac: I suppose it's in the spirit of this component, that being it ensures things are set up properly before mounting the child. Perhaps we'll break the `GameManager.unsetGame()` method in the future e.g. by introducing a bug which isn't caught by the linter (due to the non-null assertion), at least this component would be able to deal with that scenario in some user-friendlier way.

So probably not a necessary addition, but shouldn't be a bad addition either

src/components/SettingsLoader.vue
MythicManiac: I'd prefer a function over a computed property in this case, but that aside I'd think splitting code to standalone functions where possible like this makes sense

src/components/SettingsLoader.vue
MythicManiac: It probably could, mostly comes down to preference. Personally I try to stay away from promise chaining since it tends to be harder to follow for larger chains or complex logic, but that's also ultimately down to preference.

src/components/SettingsLoader.vue
anttimaki: Yeah, just being intentionally paranoid, since the idea is to make the SettingsLoader as bullet-proof as possible so we can always show a clear error message to the users.

src/components/SettingsLoader.vue
anttimaki: I agree, computed property doesn't make much sense here, [extracted to a separate function](https://github.com/ebkr/r2modmanPlus/pull/911/files#diff-449d18893f0dea4f8c02461069f6129fde06e2558467115d7766953e2f3b43abR81).

src/components/SettingsLoader.vue
anttimaki: Comes down to personal preference. I'll always choose `await` unless the existing code prompts otherwise. It just makes the code easier to grasp and edit IMO.

Also I think there's a minor difference between the code in the PR and the one suggested in the comment: PR reuses the `R2Error` returned by `settings.load()`, calling `this.handleError(error.name, error.message, solution);` whereas the code in the comment overrides the first parameter with a fixed text.

src/components/SettingsLoader.vue
anttimaki: Well, seeing what that change does to the next PR in the chain, I'd flip this around and as why not use computed property (or plain method) here? Apart from the theoretical easier testing?

MythicManiac: I don't mind it either way around, but in the interest of pushing the changes forward, I think we should merge this version now & make more changes in the latter PR if necessary

PR #914, opened 2022-12-13
Fix highlighting number of online mods in NavigationMenu

anttimaki:
The span showing number of available mods had the same background color as the parent element when user navigated to download moniton view.


MythicManiac (approved):
LGTM 👍

PR #917, opened 2022-12-15
Offer an option to reset local settings if loading them fails at launch

anttimaki:
If SettingsLoader fails to load the settings, a button for clearing out the IndexedDB containing local settings is shown. Also offers different suggestions for solutions based on what went wrong.

Refs TS-1384


anttimaki:
If loading the info of default game (RoR2) fails: 
image

If loading the settings fails the first time (this was achieved by putting a non-JSON string in the storage):
image

If user clicks the reset button, but the resetting itself fails:
image

I'm not sure if we want to use link here. It's easier for the user than copy-pasting a long URL, but it's opened in another window spawned by the mod manager rather than user's default browser.

If resetting IndexedDB succeeds, but the settings still can't be loaded:
image

Any suggestions for better wording etc. is welcome.

anttimaki:
Here's a video of how things should ideally work. The modal is trickered by reloading the app via dev tools (because it was easier for me to record the video this way) and having the code intentionally corrupt the storage on first load attempt.

https://user-images.githubusercontent.com/3500016/207885850-03727df5-cc96-4dac-b2e5-2c9ab0dbf73f.mp4

MythicManiac:
Not a full review yet, but could the button use the styling that's used elsewhere in the app? I'd assume some fairly easy to re-use classes exist for that.

src/r2mm/manager/ManagerSettings.ts
```diff
@@ -45,6 +45,18 @@ export default class ManagerSettings {
         await settings.save();
     }
 
+    /**
+     * Clear local settings stored in IndexedDB.
+     */
+    public static async resetLocalSettings() {
+        await ManagerSettings.DEXIE_STORE.delete();
```
MythicManiac, line 52: I don't like the idea that the function that's supposed to solve the potentially faulty state management of settings is itself depending on said potentially faulty state management. I'd go about the deletion in the most direct way possible with the least state management involved, i.e. the closest way comparable to the user manually deleting the IndexedDB since we know that's working.

In other words, why do we go through the abstraction here at all? Seems like a more fail-safe solution would be to drop the indexeddb directly (and without relying on the potentially buggy abstraction) & then restart the app or something along those lines

MythicManiac (commented):
Main question here, didn't review the smaller details yet

src/r2mm/manager/ManagerSettings.ts
anttimaki: I agree with dropping the IndexedDB directly.

I'm not sure if the restart is needed. I guess it would simplify the code, but would probably also lead into a loop where we just offer to drop the database again and again since we don't know if user has already tried that?

anttimaki:
@MythicManiac

* The button now has rounded corners and padding that's consistent with the rest of the buttons. [The buttons offered by Bulma](https://bulma.io/documentation/elements/button/) mostly looked off when set on the red background, so the button is still white.
* The settings database is now dropped using IndexedDB API directly. For now I chose to not require a restart, since it didn't seem necessary. Thoughts?

src/r2mm/manager/ManagerSettings.ts
MythicManiac: I don't think the restart is necessary if it's easy to make work without it, just figured properly resetting the state might also be bug prone. In this case it seems simple enough so marking this as resolved

src/components/SettingsLoader.vue
```diff
@@ -1,18 +1,52 @@
 <template>
     <div id="settings-loader">
-        <slot v-if="isLoaded" />
+        <slot v-if="phase === PHASES.LOADED" />
 
-        <div v-else :class="['modal', 'z-top', {'is-active': error}]">
+        <div v-if="phase > PHASES.LOADED" class="modal z-top is-active">
```
MythicManiac, line 5: Relying on enum ordering without it being explicitly mentioned on the enum declaration seems potentially fragile/error prone if changed in the future.

MythicManiac (approved):
Looks good enough to me, left a comment but I don't consider it a blocker

src/components/SettingsLoader.vue
```diff
@@ -1,18 +1,52 @@
...
+                        available, try installing it.
+                    </p>
+
+                    <div v-else-if="phase === PHASES.SETTINGS_FAILED">
+                        <p>
+                            Loading of local user settings failed. You
+                            can use the button below to reset the
+                            settings, but note that all settings for all
+                            games will be lost and this can't be undone.
+                            This won't affect the loaded mods or mod profiles.
```
ebkr, line 26: This will if a custom data directory was set. I need to work on a way to re-link easily as currently it expects an empty folder.

src/components/SettingsLoader.vue
```diff
@@ -1,18 +1,52 @@
...
+                        </p>
+                        <br />
+                        <button @click="resetSettings" class="button is-white">
+                            Reset settings
+                        </button>
                     </div>
+
+                    <p v-else-if="phase === PHASES.RESET_FAILED">
+                        Well, that's awkward. Resetting of the settings
```
ebkr, line 35: Preference to avoid things like "Well, that's awkard" as it adds no value and is just clutter

src/components/SettingsLoader.vue
```diff
@@ -1,18 +1,52 @@
...
+                        failed as well. You can still try to reset the
+                        settings manually by following these
+                        <a target="_blank" rel="noopener noreferrer"
```
ebkr, line 38: Any reason for noopener / noreferrer?

Can we bind the Link component provider earlier (if it needs binding?) as that can open in an actual browser

src/components/SettingsLoader.vue
```diff
@@ -1,18 +1,52 @@
...
+                            href="https://github.com/ebkr/r2modmanPlus/wiki/Error:-White-or-blank-game-select-screen-on-startup#corrupted-settings-on-update">
+                            instructions.
+                        </a>
+                    </p>
+
+                    <p v-else-if="phase === PHASES.RETRY_FAILED">
+                        That didn't do much, it seems. Locally stored
```
ebkr, line 45: See previous comment about wording, I think we want to avoid "friendly" wording and to keep it more professional and straight to the point

ebkr (commented):
Added comments primarily regarding wording

ebkr (dismissed):
Left it as comment by accident 👀

src/components/SettingsLoader.vue
MythicManiac: > Any reason for noopener / noreferrer?

Generally good practice, there are some security concerns with links that don't use these.

Not sure about the Link component suggestion, leaving that up to @anttimaki

anttimaki: My idea was that the SettingsLoader mounts before any providers etc. to reduce points of failure, and to ensure that any problem encountered is actually related to loading the settings. If having the link open in a browser is considered worth it to introduce the LinkingProvider earlier, and the LinkingProvider is seen as a stable and tested implementation, I see no reason why it couldn't be done. Personally I don't see it fitting into the idea (that I had) of the SettingsLoader, though.

MythicManiac: That makes sense to me at least, given that the goal was to minimize the points of failure for this component.

ebkr: I think it’d be good to do, and it’s stable as in it only depends on external libraries for opening, it’s not reliant on settings or anything user accessible

src/components/SettingsLoader.vue
anttimaki: [Explicitly mentioning](https://github.com/ebkr/r2modmanPlus/pull/917/commits/8cacfe28b284832caa85493bd2a01f12b30d59cf#diff-449d18893f0dea4f8c02461069f6129fde06e2558467115d7766953e2f3b43abR67) it now.

src/components/SettingsLoader.vue
anttimaki: Removed that line.

src/components/SettingsLoader.vue
anttimaki: Removed.

src/components/SettingsLoader.vue
anttimaki: This is now [passed to SettingsLoader as a prop](https://github.com/ebkr/r2modmanPlus/pull/917/commits/8cacfe28b284832caa85493bd2a01f12b30d59cf#diff-e7478582172dd38f0de771e700456eb24487444bd9553f2eff6e2b4ec0b4fd86) to keep things consistent with `logError`. Also skipped the provider and used the implementation directly so that all the providers are initiated in the same place (`App.vue`).

PR #920, opened 2022-12-20
Watch showDeprecatedPackages filter for state change to update mod list

anttimaki:
Changes to recently added new filter weren't watched, so the mod list wasn't updated unless some other filter was updated at the same time.


MythicManiac (approved):
LGTM 👍

ebkr (approved):

PR #921, opened 2022-12-22
Try to avoid corruption of the settings stored IndexedDB

anttimaki:
Couldn't find any obvious cause for the corruption, so these changes are just shots in the dark.


MythicManiac (approved):
Shot in the dark or not, these look like sane changes to me 👍

PR #923, opened 2022-12-23
Handle returned R2Errors

anttimaki:
I'm putting this up here to discuss whether changes like these make sense. Generally speaking the code base contains places where R2Errors are returned, and the caller:

* Intentionally ignores the error and it makes sense to me given the context
* Intentionally ignores the error and I'm not sure if it's wise
* Just ignores the error and I can't tell whether it's intentional or not

Any thoughts what should be done in these cases?

---

Another issue: there's multiple places where settings are loaded, but then it's not checked if `.load()` returned R2Error or not. Given the recent addition of SettingsLoader, is this okay, or should we always check the return value?

---

Marking as a draft for discussion and because this haven't been tested thoroughly manually yet.

### DIFF
--- a/src/pages/Splash.vue
+++ b/src/pages/Splash.vue
@@ -138,6 +138,7 @@ import Component, { mixins } from 'vue-class-component';
 
 import { Hero, Link, Progress } from '../components/all';
 import SplashMixin from '../components/mixins/SplashMixin.vue';
+import R2Error from '../model/errors/R2Error';
 import RequestItem from '../model/requests/RequestItem';
 import FsProvider from '../providers/generic/file/FsProvider';
 import GameDirectoryResolverProvider from '../providers/ror2/game/GameDirectoryResolverProvider';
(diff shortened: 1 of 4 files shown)


src/pages/Splash.vue
```diff
@@ -175,12 +176,23 @@ export default class Splash extends mixins(SplashMixin) {
 
     async moveToNextScreen() {
         if (process.platform === 'linux') {
-            if (!await (GameDirectoryResolverProvider.instance as LinuxGameDirectoryResolver).isProtonGame(this.activeGame)) {
+            const isProton = await (GameDirectoryResolverProvider.instance as LinuxGameDirectoryResolver).isProtonGame(this.activeGame);
+
+            if (isProton instanceof R2Error) {
```
ebkr, line 181: I'm thinking that maybe we'd benefit by having errors that can be logged to Sentry but having things not present in modals.

This is one example of an error we'd probably not want to show to a user as that means that relevant Steam files have been modified, and that's not really something that the user can solve themselves.

src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
```diff
@@ -440,7 +440,11 @@ export default class GenericProfileInstaller extends ProfileInstallerProvider {
             files: Array.from(existing.entries())
         }
         await FsProvider.instance.writeFile(path.join(profile.getPathOfProfile(), "_state", `${mod.getName()}-state.yml`), yaml.stringify(mft));
-        await ConflictManagementProvider.instance.overrideInstalledState(mod, profile);
+        const err = await ConflictManagementProvider.instance.overrideInstalledState(mod, profile);
```
ebkr, line 443: Could you verify that this doesn't cause any issues when installing BONEWORKS mods?
I've looked through the code and have a feeling it'll be fine, but would be good to sanity check.

ebkr (changes requested):

anttimaki:
@ebkr this is a read receipt for your comments. I will look into both issues you mentioned, but don't quite know when.

anttimaki:
Closing an ancient PR. Never got around to doing what I said I would do, and this is now outdated. I think at least some of the issues have been fixed in other PRs in the meanwhile.

PR #924, opened 2022-12-23
Fix return value typing

anttimaki:


MythicManiac (approved):
Were the type hints just wrong? I'd have expected to see changes that start throwing errors instead of returning them, but I guess that was already the paradigm here and just the type hints were wrong?

If the above is accurate, then this looks good to me

anttimaki:
> Were the type hints just wrong? I'd have expected to see changes that start throwing errors instead of returning them, but I guess that was already the paradigm here and just the type hints were wrong?

Yes, they were wrong. This purely an assumption, but maybe in the first commit saving the settings used to return an error in some scenarios in the past, but doesn't anymore? And in the other commits, maybe the idea was the methods would return errors when the abstract interface was designed, but the implementation ended up never returning them? 🤷🏻

ebkr (approved):
Yeah I think this was all legacy, happy with the change given they're all just setters

PR #925, opened 2022-12-23
Throw error in throwNotProvidedError

anttimaki:
The name implies the method throws rather than returns an Error.


MythicManiac (approved):
LGTM 👍

PR #926, opened 2022-12-23
Exclude deprecated mods from NavigationMenu's counter

anttimaki:
Users can use a filter to show deprecated mods available online. By default deprecated mods are not shown. However, they were always counted towards the number shown on the NavigationMenu. This might confuse users when the sidebar says there's 3 mods but the list only shows 2.

While the above also applies to other filters, deprecated mods is different since it hides content by default, whereas others require user action to do so.

### DIFF
--- a/src/components/navigation/NavigationMenu.vue
+++ b/src/components/navigation/NavigationMenu.vue
@@ -17,7 +17,7 @@
                         <router-link :to="{name: 'manager.installed'}" class="tagged-link">
                             <i class="fas fa-folder tagged-link__icon icon--margin-right" />
                             <span class="tagged-link__content">Installed</span>
-                            <span :class="getTagLinkClasses(['manager.installed'])">{{localModList.length}}</span>
+                            <span :class="getTagLinkClasses(['manager.installed'])">{{localModCount}}</span>
                         </router-link>
                     </li>
                     <li>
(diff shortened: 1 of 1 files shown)


anttimaki:
Alternative approach would be to always exclude the deprecated mods from NavigationMenu's count, regardless of the filters. This approach would be consistent in the sense that none of filters would affect the count.

src/components/navigation/NavigationMenu.vue
```diff
@@ -82,12 +83,16 @@ export default class NavigationMenu extends Vue {
     private contextProfile: Profile | null = null;
     private LaunchMode = LaunchMode;
 
-    get thunderstoreModList() {
-        return this.$store.state.thunderstoreModList || [];
+    get thunderstoreModCount() {
```
MythicManiac, line 86: I'm assuming this doesn't need to
```ts
@Watch("$store.state.modFilters.showDeprecatedPackages")
```
?

MythicManiac (approved):
A single question, but looks good if it works as I think it does

src/components/navigation/NavigationMenu.vue
anttimaki: Based on manual testing it's not needed, although I must admit I'm not entirely sure why watching is not needed here while it's needed in some other places.

anttimaki:
Force-pushed to remove an import that's no longer needed after the changes in this PR.

src/components/navigation/NavigationMenu.vue
MythicManiac: Perhaps because it's a getter?

anttimaki: Tried to turn it into a regular method and stuff still worked 😔

ebkr: Weird, get makes it a computed method and so it makes sense that you don't need to watch. As a regular method I'm not sure, I'd have expected a watch or something 🤔

ebkr (approved):

PR #934, opened 2023-01-07
Added debounce to improve performance of online search

ebkr:
Reimplementation with PR comments addressed from #908


MythicManiac (approved):
👍 nice

PR #945, opened 2023-01-17
Add game support: ULTRAKILL

MythicManiac:


src/r2mm/installing/default_installation_rules/InstallationRuleApplicator.ts
```diff
@@ -13,6 +13,7 @@ import InstallRules_HardBullet from '../default_installation_rules/game_rules/In
 import InstallRules_BackpackHero from '../default_installation_rules/game_rules/InstallRules_BackpackHero';
 import InstallRules_BONELAB from '../default_installation_rules/game_rules/InstallRules_BONELAB';
 import { buildBepInExRules } from '../default_installation_rules/game_rules/InstallRules_BepInex';
+import path from "path";
```
ebkr, line 16: I think the import is usually `import * as path from path`
I'm not sure if there's any functional difference but I think there may be a reason it's done that way

src/r2mm/installing/default_installation_rules/InstallationRuleApplicator.ts
```diff
@@ -65,7 +66,12 @@ export default class InstallationRuleApplicator {
             buildBepInExRules("TromboneChamp"),
             buildBepInExRules("RogueGenesia"),
             buildBepInExRules("AcrossTheObelisk"),
+            buildBepInExRules("ULTRAKILL", [{
+                route: path.join("BepInEx", "UMM Mods"),
+                defaultFileExtensions: [],
+                trackingMethod: "SUBDIR",
```
ebkr, line 72: Does UMM load mods in subdirectories?

ebkr (approved):

src/r2mm/installing/default_installation_rules/InstallationRuleApplicator.ts
MythicManiac: Yep

PR #949, opened 2023-01-20
Removed dotnet from ConfigSelectionLayout.vue

ebkr:
### TODO
- [x] Remove dotnet folder from new exports


MythicManiac (approved):
Not a blocker, but should we also exclude these from exports?

That can be done separately even if we do though, so feel free to merge this.

ebkr:
Good catch, they probably should. Will add a task.

PR #965, opened 2023-02-06
Add support for several new games

MythicManiac:
Add support for the following games:
- Brotato
- Ancient Dungeon VR
- Atrio: The Dark Wild
- Ultimate Chicken Horse

Additionally, add support for a new install rule type which simply moves the package zip to the target directory, as that's what the godot mod loader expects.

This commit does not yet include installation logic for the godot mod loader, as that doesn't have a locked in spec yet.


src/model/game/GameManager.ts
```diff
@@ -340,6 +340,30 @@ export default class GameManager {
...
+            GameSelectionDisplayMode.VISIBLE, GameInstanceType.GAME, PackageLoader.BEPINEX, ["uch"]),
+
+        new Game("Atrio: The Dark Wild", "AtrioTheDarkWild", "AtrioTheDarkWild",
+            "Atrio The Dark Wild", ["Atrio.exe"], "Atrio_Data",
+            "https://thunderstore.io/c/atrio-the-dark-wild/api/v1/package/", "https://raw.githubusercontent.com/ebkr/r2modmanPlus/master/modExclusions.md",
+            [new StorePlatformMetadata(StorePlatform.STEAM, "1125390")], "atrio-the-dark-wild.jpg",
+            GameSelectionDisplayMode.VISIBLE, GameInstanceType.GAME, PackageLoader.BEPINEX, ["adw"]),
+
+        new Game("Brotato", "Brotato", "Brotato",
+            "Brotato", ["Brotato.exe"], "Atrio_Data",
```
KANAjetzt, line 357: Looks like a typo? "Atrio_Data" -> "Brotato_Data"?

MythicManiac: Good catch! It should actually be empty since Brotato isn't an Unity game, and this is required for Unity games only

src/r2mm/installing/default_installation_rules/InstallationRuleApplicator.ts
```diff
@@ -14,6 +14,10 @@ import InstallRules_BackpackHero from '../default_installation_rules/game_rules/
 import InstallRules_BONELAB from '../default_installation_rules/game_rules/InstallRules_BONELAB';
 import { buildBepInExRules } from '../default_installation_rules/game_rules/InstallRules_BepInex';
 import * as path from 'path';
+import { buildGodotMlRules } from "../default_installation_rules/game_rules/InstallRules_GodotML";
```
ebkr, line 17: This would probably read better as `buildGodotModLoaderRules`
Alternatively further down you use `ML` instead of `Ml` so it'd be good for consistency if you prefer ML.

src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
```diff
@@ -286,6 +287,40 @@ export default class GenericProfileInstaller extends ProfileInstallerProvider {
...
 
+    private async installPackageZip(profile: Profile, rule: ManagedRule, installSources: string[], mod: ManifestV2) {
+        /*
+            This install method repackages the entire mod as-is and places it to the
+            destination route. Essentially the same as SUBDIR_NO_FLATTEN, but as a
+            zip instead of a directory. The zip name will be the mod ID.
+         */
+        const destDir = path.join(profile.getPathOfProfile(), rule.route);
+        await FileUtils.ensureDirectory(destDir);
+        const destination = path.join(destDir, `${mod.getName()}.ts.zip`);
```
ebkr, line 298: Might be getting confused but does this include the author name?

Think mod.getFullName() might do it if it doesn't?

src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
```diff
@@ -286,6 +287,40 @@ export default class GenericProfileInstaller extends ProfileInstallerProvider {
...
+        const cacheDirectory = path.join(PathResolver.MOD_ROOT, 'cache');
+        const cachedLocationOfMod: string = path.join(cacheDirectory, mod.getName(), mod.getVersionNumber().toString());
+        const builder = ZipProvider.instance.zipBuilder();
+        await builder.addFolder("", cachedLocationOfMod);
+        await builder.createZip(destination);
```
ebkr, line 303: Not much of a concern right now but it may be nice to have this stored somewhere in the cache so we don't need to do it each time. One issue is storage space, but generally I think it would help to speed up reinstalls across profiles as zipping larger mods may take a while.

ebkr (changes requested):
A few minor comments but otherwise looks good 👍

src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
MythicManiac: This is a manifest which seems to do
```ts
this.setName(mod.getFullName());
```
leading the `getName()` to return the fullName. In fact it doesn't have the `fullName` property at all.

src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
MythicManiac: I considered it but left it out for the sake of code simplicity. I figured we'll eventually rework this area and can account for it when that happens (be it just refactoring the current installer or migrating to TCLI).

ebkr (approved):

PR #967, opened 2023-02-12
Proton detection

ebkr:
Added depots to aid proton detection.
Games with Linux support have a reasonable default, and unknown games will fallback to an exe lookup to determine proton/native.

### DIFF
--- a/src/depots/depot.schema.json
+++ b/src/depots/depot.schema.json
@@ -40,7 +40,7 @@
                 ]
             }
         },
-        "additionalProperties": false,
+        "additionalProperties": false
     },
     "required": [
         "depots", "$schema"
--- a/src/depots/depot.schema.json
+++ b/src/depots/depot.schema.json
@@ -1,0 +1,48 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "depot.v1",
+    "title": "Depot",
+    "description": "Launch determination via depot version",
+    "type": "object",
+    "properties": {
+        "$schema": {
+            "description": "The schema to use",
+            "type": "string"
+        },
+        "depots": {
+            "description": "Container for all depots",
+            "type": "object",
+            "default": {
+                "description": "The fallback if the depot can't be resolved",
+                "type": "object",
+                "use_proton": {
+                    "description": "Should proton be used",
+                    "type": "boolean"
+                },
+                "required": [
+                    "use_proton"
+                ]
+            },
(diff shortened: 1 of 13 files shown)


src/depots/loader/DepotLoader.ts
```diff
@@ -0,0 +1,46 @@
...
+            this.load();
+        }
+        const depots = this.LOADED_DEPOTS.get(game.settingsIdentifier);
+        if (depots === undefined) {
+            throw new R2Error(
+                `Unable to load depot for ${game.settingsIdentifier}`,
+                "This may be an issue with the manager. Report the issue in the appropriate discord server."
+            );
+        }
+        const foundDepotKey = Object.keys(depots).find(value => value === depotIdentifier);
```
MythicManiac, line 35: Couldn't this be simplified to just access the object via key and check if the result is undefined? Given that accessing a key that doesn't exist in javascript just returns `undefined` rather than failing.

I also think the `depots` variable should be type-hinted on line 28 already in a fashion that doesn't require casting later on, assuming it's possible.

src/depots/depot.schema.json
```diff
@@ -0,0 +1,48 @@
+{
```
MythicManiac, line 1: Given that we have a schema, should there maybe be a test case that validates all depot JSONS are compatible with the schema?

src/depots/loader/DepotLoader.ts
```diff
@@ -0,0 +1,46 @@
...
+    public static DEPOT_DEFAULT_KEY = "default";
+
+    private static LOADED_DEPOTS: Map<string, Depot[]> = new Map();
+
+    private static load(): Map<string, Depot[]> {
+        const master = require('../master.json');
+        console.log("Master:", master);
+        const depotMap = new Map<string, Depot[]>();
+        for (let depotsKey in master.depots) {
+            // @ts-ignore
```
MythicManiac, line 16: Is this really required? You could probably log it with ``console.log(`Depot key: ${depotsKey}`);`` without ts-ignore. Alternatively (and I'd argue the better option) you could explicitly type-hint the `master` variable on line 12

MythicManiac (approved):
I don't see any blocker level issues, some code quality questions & comments, up to you if you want to do anything about them

src/depots/depot.schema.json
ebkr: I'll raise an issue to resolve this

PR #985, opened 2023-02-27
Include GameRunningModal in all pages supporting game launching

anttimaki:
Game can be launched from all pages containing the NavigationMenu, but downloads, config editor, and help page didn't actually contain the modal that's supposed to be shown when a game is launched.


MythicManiac (approved):
LGTM 👍

ebkr (approved):

PR #986, opened 2023-03-01
Replace bulma-tooltip with floating-vue

anttimaki:
Bulma-tooltip positions the tooltip relative to the element that has it. This works fine on r2mm, but new layouts in tsmm use "overflow: hidden", which also clips the tooltips.

Floating-vue positions the tooltip more globally. For it to work we need to add "position: relative" to body element though.

- Replace tooltips in templates
- Add tooltips to "drag to reorder" and "expand"/"collapse" card icons, seemed like an oversight that they didn't have tooltips while other icons did
- Use separate tooltips for "deprecated" and "disabled" tags. While the old implementation went to some lengths to have them share a tooltip, using separate tooltips on the tags themselves rather than the wide wrapping element makes it safer to position tooltips so they don't get placed partially outside the viewport


anttimaki:
Installed mods:
image
image
image
image
image
image
image
image
image

Online mods:
image
image
image
image

ebkr (approved):
Looks good

MythicManiac (approved):
LGTM

PR #998, opened 2023-03-26
Update modExclusions.md

codengine:


ebkr:
Thanks for the PR!
The line should be `codengine-SOTFEdit` as it's matched on the dependency string.

ebkr (approved):

PR #1001, opened 2023-04-01
Skul the hero slayer fixes

ebkr:


src/r2mm/installing/profile_installers/ModLoaderVariantRecord.ts
```diff
@@ -76,5 +76,10 @@ const VARIANTS = {
 // Exported separately from the definition in order to preserve the key names in the type definition.
 // Otherwise this would become [key: string] and we couldn't use the game names for type hinting elsewhere.
 // Casting is done here to ensure the values are ModLoaderPackageMapping[]
+
+export const VARIANT_DEPENDENCIES: {[key: string]: string[]} = {
+    SkulTheHeroSlayer: ["UnityLibs-SkulTheHeroSlayer_UnityUnstripped"]
+}
```
MythicManiac, line 82: Rather than adding a new map here, could we just extend the `ModLoaderPackageMapping` class to include a field for the dependencies? That'd be a bit easier to follow in terms of the code & would probably be more semantically correct, as it's not the _game_ that's bound to those dependencies, but rather a specific mod loader package for that game.

src/r2mm/launching/instructions/DynamicGameInstruction.ts
```diff
@@ -2,6 +2,7 @@ export enum DynamicGameInstruction {
     BEPINEX_PRELOADER_PATH = "@bepInExPreloaderPath",
     PROFILE_DIRECTORY = "@profileDirectory",
     BEPINEX_CORLIBS = "@bepInExCorlibs",
+    SKUL_THE_HERO_SLAYER_CORELIBS = "@skulTheHeroSlayerCorlibs",
```
MythicManiac, line 5: Any way to make this not be skul specific? Can the `BEPINEX_CORLIBS` perhaps be re-used instead (AFAIK there shouldn't be anything special about how skul handles them right?)

src/r2mm/launching/instructions/instructions/loader/SkulTheHeroSlayerGameInstructions.ts
```diff
@@ -0,0 +1,60 @@
+import GameInstructionGenerator from '../GameInstructionGenerator';
+import { GameInstruction } from '../../GameInstructions';
+import Game from '../../../../../model/game/Game';
+import Profile from '../../../../../model/Profile';
+import FsProvider from '../../../../../providers/generic/file/FsProvider';
+import path from 'path';
+import { DynamicGameInstruction } from '../../DynamicGameInstruction';
+import { GameInstanceType } from '../../../../../model/game/GameInstanceType';
+
+export default class SkulTheHeroSlayerGameInstructions extends GameInstructionGenerator {
```
MythicManiac, line 10: Any way we can avoid creating an entirely game-specific dedicated instruction class just for skul, and instead extend the generic instructions?

MythicManiac (commented):
I'm mostly concerned about changes that are made very skul specific, as there's at least a few places which look like they could be made generic instead. Primarily I'm considering this from the maintenance perspective, as a single system is of course easier to maintain than having game specific special scenarios sprinkled around. This naturally applies to any CLI port of the mod management logic too.

Anyway, I don't know how much of what I commented on is reasonable to make generic, and how much requires hardcoding anyway, so very much open to discussion at this point.

Issue #1008, opened 2023-04-13
[BUG] - Quill editor sets cursor to index 0 after every change


sheybey:
**Describe the bug**
When you are editing a config file, every change moves the cursor to the first character of the buffer, making it extremely difficult to type anything in.

**To Reproduce**
1. Select a mod profile for any game that has installed mods with config files
2. Click on "Config editor"
3. Find a config file that isn't parsed by r2modman (any JSON should do)
4. Click "Edit config"
5. Try to type in a word

**Expected behavior**
The edit is made and the cursor stays where it is

**Additional context**
This is caused by keeping the entire content of the buffer as a component prop and calling setText on every update. The quill documentation implies that it manufactures a minimal delta and keeps other state, but in reality it is just generating a delete(...).insert(newValue) Delta, which causes the cursor to move to index 0 every time.

This can be worked around by saving the selection and manually restoring it after every update (PR incoming.)

ebkr:
Closing as merged

PR #1009, opened 2023-04-13
save and restore editor selection on model update

sheybey:
This fixes the cursor being moved to the beginning of the buffer on every edit.

This PR saves the selection before calling setText and restores the cursor index afterwards, clamping it at the end of the new buffer. However, this is still very slow for large config files since it is replacing the entire contents of the editor on every update. Maybe a better solution would be to let Quill manage its own state (model would be initialValue, onSave(newValue)?)

I am not familiar enough with vue/electron and their performance characteristics to make a more intelligent change here. This just fixes the cursor getting reset every keypress.


sheybey:
#1008

src/components/QuillEditor.vue
```diff
@@ -35,7 +35,11 @@
         @Watch("textValue")
         private updatedModel(newValue: string) {
             if (this.editor !== null) {
+                const selection = this.editor.getSelection();
                 this.editor.setText(newValue);
+                if (selection) {
+                    this.editor.setSelection(Math.min(newValue.length, selection.index), 0);
```
MythicManiac, line 41: I'm pretty sure you can omit the second parameter entirely as it defaults to 0

MythicManiac (approved):
I can't really comment on the performance aspect either since I don't know how the editor is implemented, but this sounds like an improvement. Thanks for the contribution!

ebkr:
Thanks for the contribution! There was no mention of it being an issue before you pointed it out and it never used to behave like that, so I have no idea how long this been an issue for

PR #1018, opened 2023-04-26
resolve end header not found

### LINKED ISSUE #933: [BUG] - Import Code Fails with "Invalid or unsupported zip format. No END header found"
**Describe the bug**
This issue is occurring with Valheim, but I suspect that #916 is experiencing the same issue with RoR2. A friend of mine is exporting their mod profile as a code so that I may import the mod profile as a code. Attempting to do so produces the error, "Invalid or unsupported zip format. No END header found". 

**To Reproduce**
1. Install Thunderstore app
2. Create a profile for Valheim and install mods
3. Launch Valheim for initial setup of mod files
4. Export profile with a code
5. Install and launch r2modmanPlus
6. Select Valheim as the game
7. In Profile Selection, choose "Import/Update"
8. Choose "Import new profile" or "Update existing profile" (both provide the same result)
9. Select "From code"
10. Paste the code into the input element and press "Import" (screenshot provided)
11. See error (screenshot provided)

**Expected behavior**
A new profile should be created containing the mods associated with the exported profile represented by the code.  

**Screenshots**
2023-01-06-024103_646x245_scrot
2023-01-06-023841_650x181_scrot

**Additional context**
The code that we were using: 0185868e-7969-ef67-b878-7db08f3a0d45

The mod zip file appears to be found over HTTPS. Perhaps the zip file is not downloaded completely before the extraction process begins?

The error is caught and displayed in the following code from `src/pages/Profiles.vue` and would be a good place to start investigating. Will do this myself if there is time; otherwise this could be a good first issue for others.
2023-01-06-024724_679x158_scrot

wagyourtail:
I figured out why this is a problem...
it appears that with thunderstore app, the end of the zip gets padded with a bunch of 0's after for some reason.
I would guess this is an issue with padding on their buffer that they just write instead of trimming first.

image
the zip reader doesn't like that

this pr fixes that by ensureing the zips are trimmed

resolves #933


MythicManiac:
Good find! The current implementation seems to load the entire zip to memory before writing it back to disk, which is potentially a bad idea to do as some profile exports can be quite large. I can't immediately think of a good way to address that, but we might be able to fix how zip creation works on the other mod manager & ensure it doesn't create trailing zeroes at least in the meanwhile.

wagyourtail:
well. the FsProvider doesn't have the node functions for streaming file content...

wagyourtail:
done

src/pages/Profiles.vue
```diff
@@ -486,6 +486,40 @@ export default class Profiles extends Vue {
...
+        // ensure exitst
+        if (headerLoc === -1) {
+            throw new Error("Failed to find zip end header");
+        }
+
+        // header loc is size, so we must add 1 since it's 0 indexed
+        headerLoc += 1;
+
+        // trim file
+        // zip end header is 22 bytes long
```
ebkr, line 517: Is it guaranteed to always be 22 bytes long? Is there any case where it could be 21/23?

src/pages/Profiles.vue
```diff
@@ -486,6 +486,40 @@ export default class Profiles extends Vue {
...
+        await fs.truncate(file, headerLoc + 22);
+        
+        console.log(file)
```
ebkr, line 520: Would be good to remove this prior to a merge

src/pages/Profiles.vue
```diff
@@ -486,6 +486,40 @@ export default class Profiles extends Vue {
...
+                    break;
+                } else {
+                    headerLoc += data.length;
+                }
+            }
+        });
+
+        // ensure exitst
+        if (headerLoc === -1) {
+            throw new Error("Failed to find zip end header");
```
ebkr, line 510: We tend to throw `R2Error` instead of `Error` as we use it for display purposes. They help the user identify what went wrong and how to potentially solve it. It's also somewhat an identifier as to where it broke in the process as the user will send a screenshot.

src/pages/Profiles.vue
```diff
@@ -495,6 +529,7 @@ export default class Profiles extends Vue {
         if (files[0].endsWith('.r2x')) {
             read = await fs.readFile(files[0]).toString();
         } else if (files[0].endsWith('.r2z')) {
+            await this.fixThunderstoreFile(files[0]);
```
ebkr, line 532: Maybe `sanitizeImportEndHeader` would be better?

src/providers/generic/file/NodeFs.ts
```diff
@@ -152,4 +153,33 @@ export default class NodeFs extends FsProvider {
...
+            });
+        });
+    }
+
+    async createReadStream(path: string, chunk_size: number, callback: (arg0: Readable) => Promise<void>): Promise<void> {
+        return new Promise((resolve, reject) => {
+            NodeFs.lock.acquire(path, async () => {
+                try {
+                    const stream = fs.createReadStream(path, {
+                        highWaterMark: chunk_size
```
ebkr, line 175: What does this do? Could possibly benefit from a comment or a link to docs

ebkr (changes requested):
Added a few comments, thanks for the contribution! Sorry for the delayed response.

MythicManiac:
The real blocker with the PR is that on TMM the I/O implementation is done by a C# plugin, and we can't pass complex types through that IPC layer, including streams I believe. In other words, this has to be implemented in some fashion that avoids passing complex types to/from the FS provider.

One option would be to implement the whole zip preprocessing logic in the FS provider, which we can keep as a stub on the C# plugin since the C# code already is able to handle the zips as is.

src/pages/Profiles.vue
wagyourtail: yes https://en.wikipedia.org/wiki/ZIP_(file_format)#End_of_central_directory_record_(EOCD) (note that for some reason wikipedia's using the other endianness for each thing, the bytes for the start of the header are in the right order), 

unless it's a zip64, idk if we need to handle those tho, are they ever created?

wagyourtail:
> The real blocker with the PR is that on TMM

It's fixed here, so I don't see why fixing in on TMM would block this PR seeing as this is the fix for the issue on this end.

MythicManiac:
> It's fixed here, so I don't see why fixing in on TMM would block this PR seeing as this is the fix for the issue on this end, and doesn't rely on TMM fixing it.

It's because the same code runs on both for a large portion of the core mod install logic to keep mod installs consistent between the two. In other words, we'll need to make similar adjustments to the filesystem provider implementation on TMM as was done here or the build breaks. And as things currently are implemented, I don't think it's possible to do due to the C# interop layer not supporting complex types.

So I'd suggest moving the current `sanitizeImportEndHeader` functionality to the FsProvider, e.g. as a function called `sanitizeZip` which takes in a filepath as a parameter and overwrites it in-place. That way there's no need to pass complex types through the interop layer on TMM but the fix works here.

ebkr:
@wagyourtail For context, r2modman is a git submodule that TMM uses, so the code internally isn't actually modifiable from TMM's end. We use a "provider" that lets TMM override behaviour of certain areas. So by having the code in the provider TMM can avoid using it and use the C# layer which works as intended.

PR #1019, opened 2023-05-07
Add current game to the profile meta data

Oksamies:
refs -


src/model/exports/ExportFormat.ts
```diff
@@ -2,17 +2,29 @@ import ExportMod from './ExportMod';
 
 export default class ExportFormat {
     private profileName: string = '';
+    private profileGameDisplayName: string = '';
```
ebkr, line 5: Why is this needed?

PR #1053, opened 2023-08-02
Make modloader mappings global

MythicManiac:
Edit the hardcoded modloader mappings so that any modloader package can be used with any game.

Rationale:

There has been a few instances in the past where a community has swapped the package they use for mod loaders, and until now it has always required an update to the hardcoded definitions. While this change does not yet remove the hardcoding entirely, it brings the system more in line with the future goals of having the packages declare how they should be installed themselves. In other words, whether or not a package is handled as a mod loader should not depend on which game are we managing mods for, which is exactly what this change addresses.


src/r2mm/installing/profile_installers/ModLoaderVariantRecord.ts
```diff
@@ -1,89 +1,140 @@
 import ModLoaderPackageMapping from '../../../model/installing/ModLoaderPackageMapping';
 import { PackageLoader } from '../../../model/installing/PackageLoader';
 
+
 /**
- * Used to record which package to handle based on the current game.
+ * The hardcoded metadata here should really be a property of the packages themselves & parsed from there when
+ * installing, but as that doeos not yet exist, we have to maintain a mapping here.
```
anttimaki, line 7: "doeos"

anttimaki (approved):
Could this be a good moment to order `MODLOADER_PACKAGES` e.g. by loader and/or  package? Or does the order in which the games have been added convey some meaning?

MythicManiac:
Could be sorted yes, but I don't have a non-manual way of doing it so I'm not going to

PR #1054, opened 2023-08-02
Add a node script for syncing supported games list

MythicManiac:
Add a new script to the project which can be used to update the in-repo copy of the games JSON.

Currently the JSON is not used anywhere yet, but the plan is to convert current hardcoded game definitions to be loaded from the JSON file, and to implement a system which can update it without having to do a full app build.

### DIFF
--- a/scripts/sync.ts
+++ b/scripts/sync.ts
@@ -1,0 +1,19 @@
+import fs from "fs";
+
+/**
+ * This script synchronizes the in-repo ecosystem schema JSON to the latest
+ * version.
+ */
+
+async function updateSchema() {
+    console.log("Updating games.json...")
+    const response = await fetch("https://thunderstore.io/api/experimental/schema/dev/latest/");
+    if (response.status !== 200) {
+        throw new Error(`Received non-200 status from schema API: ${response.status}`);
+    }
+    const data = Buffer.from(await response.arrayBuffer());
+    fs.writeFileSync("./src/assets/data/games.json", data);
+    console.log("Done!");
+}
+
+updateSchema();
--- a/src/assets/data/games.json
+++ b/src/assets/data/games.json
@@ -1,0 +1,5886 @@
+{
+  "schemaVersion": "0.0.11",
+  "games": {
+    "20-minutes-till-dawn": {
+      "uuid": "57ae7cc1-b21a-4592-8415-73512a91b180",
+      "label": "20-minutes-till-dawn",
+      "meta": {
+        "displayName": "20 Minutes Till Dawn",
+        "iconUrl": "20MinutesTillDawn.jpg"
+      },
+      "distributions": [
+        {
+          "platform": "steam",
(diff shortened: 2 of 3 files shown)


scripts/sync.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * This script synchronizes the in-repo ecosystem schema JSON to the latest
+ * version.
+ */
+
+async function updateSchema() {
+    console.log("Updating games.json...")
+    const response = await fetch("https://thunderstore.io/api/experimental/schema/dev/latest/");
+    if (response.status !== 200) {
```
anttimaki, line 11: Since this is run before all builds and publish command, Is totally preventing the execution what we want if the schema file is unavailable?

src/assets/data/games.json
anttimaki, line 1: Did not review this file.

anttimaki (approved):

scripts/sync.ts
```diff
@@ -0,0 +1,19 @@
+import fs from "fs";
+
+/**
+ * This script synchronizes the in-repo ecosystem schema JSON to the latest
+ * version.
+ */
+
+async function updateSchema() {
+    console.log("Updating games.json...")
```
ebkr, line 9: Nit: we can probably drop the console.log but not overly bothered

ebkr (approved):
Approving - Cautious on image handling when synced but isn't a concern currently as not in use

src/assets/data/games.json
```diff
@@ -0,0 +1,5886 @@
...
+        "displayName": "20 Minutes Till Dawn",
+        "iconUrl": "20MinutesTillDawn.jpg"
+      },
+      "distributions": [
+        {
+          "platform": "steam",
+          "identifier": "1966900"
+        }
+      ],
+      "r2modman": {
```
ebkr, line 17: Should this be something like `mod-manager` instead of `r2modman`?

ebkr (approved):
Quick skim of the games.json file, approving with one comment but as it's generated it doesn't particularly affect this PR

scripts/sync.ts
MythicManiac: The build failing was intentional (as the idea was you should always have the latest version of the schema), but I'm going to remove it from the build script after all since we don't do any validation on it, meaning you could end up with an invalid schema in your build in some specific failure scenarios.

scripts/sync.ts
MythicManiac: This is only shown when you run a CLI command in the project so that you can keep track of what's going on, as it was hard to tell without any status indicators.

src/assets/data/games.json
MythicManiac: No since it's explicitly tailored for r2modman's needs currently, and I wanted the name to indicate that. (In fact, everything within the r2modman object is exported from the r2modman code).

PR #1055, opened 2023-08-03
Refactor modloader installation

MythicManiac:
Start splitting the massive GenericProfileInstaller class to smaller and better contained units, starting by creating a dedicated PackageInstaller class for each of the mod loaders that have special installation functionality.

### DIFF
--- a/src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
+++ b/src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
@@ -21,6 +21,84 @@ import ZipProvider from "../../../providers/generic/zip/ZipProvider";
 
 const basePackageFiles = ["manifest.json", "readme.md", "icon.png"];
 
+
+
+abstract class PackageInstaller {
+    abstract install(mlLocation: string, modLoaderMapping: ModLoaderPackageMapping, profile: Profile): Promise<void>;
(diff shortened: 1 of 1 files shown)


src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
```diff
@@ -21,6 +21,84 @@ import ZipProvider from "../../../providers/generic/zip/ZipProvider";
 
 const basePackageFiles = ["manifest.json", "readme.md", "icon.png"];
 
+
+
+abstract class PackageInstaller {
+    abstract install(mlLocation: string, modLoaderMapping: ModLoaderPackageMapping, profile: Profile): Promise<void>;
```
anttimaki, line 27: Kinda annoyed that first param uses `ml` abbreviation and the second one doesn't, but I guess changing that would cause noise in other files?

anttimaki (approved):

src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
MythicManiac: Potentially, I tried to keep changes to a minimum here so this is just the same function as before moved to a different class, didn't refactor it further than that yet.

PR #1064, opened 2023-08-19
Adds compatibility with steam flatpak.

PedroVH:


src/r2mm/launching/runners/linux/SteamGameRunner_Linux.ts
```diff
@@ -68,8 +69,10 @@ export default class SteamGameRunner_Linux extends GameRunnerProvider {
 
         LoggerProvider.instance.Log(LogSeverity.INFO, `Steam directory is: ${steamDir}`);
 
+        const steamCmd = steamDir.indexOf(path.join(homedir(), '.var', 'app', 'com.valvesoftware.Steam')) == 0 ? `flatpak run --branch=stable --arch=x86_64 --command=/app/bin/steam-wrapper --file-forwarding --filesystem=home/.config/r2modmanPlus-local com.valvesoftware.Steam` : `"${steamDir}/steam.sh"`;
```
ebkr, line 72: A few questions which could be good to cover as a comment block above the line:

1. Why is `--branch=stable` needed?
2. What is this doing? `--command=/app/bin/steam-wrapper`
3. What does `--filesystem=home/.config/r2modmanPlus-local com.valvesoftware.Steam` do?
4. File forwarding?

Is there a way we can determine if it's installed using flatpak beforehand rather than essentially doing: `if exe else flatpak`?

src/r2mm/launching/runners/linux/SteamGameRunner_Linux.ts
```diff
@@ -7,10 +7,11 @@ import R2Error from '../../../../model/errors/R2Error';
 import Profile from '../../../../model/Profile';
 import ManagerSettings from '../../../manager/ManagerSettings';
 import GameDirectoryResolverProvider from '../../../../providers/ror2/game/GameDirectoryResolverProvider';
-import LoggerProvider, { LogSeverity } from '../../../../providers/ror2/logging/LoggerProvider';
-import { exec } from 'child_process';
+import LoggerProvider, {LogSeverity} from '../../../../providers/ror2/logging/LoggerProvider';
+import {exec} from 'child_process';
 import GameInstructions from '../../instructions/GameInstructions';
 import GameInstructionParser from '../../instructions/GameInstructionParser';
+import {homedir} from "os";
```
ebkr, line 14: Looks like you can remove this

src/r2mm/launching/runners/linux/SteamGameRunner_Linux.ts
PedroVH: Yeah, these arguments were necessary when running r2modman **as a** flatpak, and i forgot to reevaluate them for just steam flatpak compatibility. I will update my PR, only `flatpak run com.valvesoftware.Steam` is necessary.

As for determinating beforehand if steam is flatpak, I'll have to have a better look at the code. Do you have any suggestions of where I could make this check?

README.md
PedroVH, line 1: for this to work, the user needs to give Steam flatpak access to ~/.config/r2modmanPlus-local, so I added an note in the README. This is necessary, or else steam won't be able to start BepInEx.

src/r2mm/launching/runners/linux/SteamGameRunner_Linux.ts
PedroVH: Oh, I found this LinuxGameDirectoryResolver, there is already checks like isProtonGame in that class. I think I'll put a "isFlatpakSteam" there

lbrunkho:
Was just looking into making these changes myself! Glad I checked the list of PR. I am running Fedora with Steam installed as a flatpak and r2modman as an PRM package. I think this should solve the issues I am seeing by correctly calling the flatpak steam executable. Let me know if you would like me to build and test this, else I will wait for it to be merged.

lbrunkho:
If this gets merged in, I can make a request on [flathub forms](https://discourse.flathub.org/c/requests/5). 


In the event the maintainers don't want to ship a flatpak themselves.

PedroVH:
The goal of this PR is to make r2modman compatible with steam flatpak. 
This [other PR](https://github.com/ebkr/r2modmanPlus/pull/798) does the same thing AND also makes necessary changes to pack r2modman as a flatpak. 

Since that fork was not being maintained, I made my fork based on my needs. I'm keeping it in sync with the official releases, and I'm packaging to pacman, so if anyone wants to use that just go to the releases page in my fork. Also if anyone wants it packaged in another way just ping me

src/r2mm/launching/runners/linux/SteamGameRunner_Linux.ts
RedTopper, line 72: I'm using Flatpak on immutable Kinoite. By default, r2modman seems to detect the steam directory under `/var/home/<username>/.var/app/com.valvesoftware.Steam/.local/share/Steam`. Since Kinoite `/home` is a symlink to `/var/home`, `homedir()` returns a path starting with `/home/<username>` and not `/var/home/<username>`. So while that path will be contained in steamDir it's index won't be 0.

 I changed the ` == 0` to ` >= 0` and was able to launch via flatpak successfully because of that.

Mervius:
If that fork were a flatpak or rpm, it'd be so much nicer for me. As it is, I can open the .pacman file and launch the application by finding it in the directory, but that is very inconvenient.

PedroVH:
> If it were an rpm that would be a bit more convenient for me personally, but I can still run the program as it is now.

Check the latest release in my fork, just added other packagings. I haven't tested besides pacman tho

fanuch:
I've been trying to get r2modman to work with Flatpak and couldn't as it wouldn't accept the steam binary in the relevant flatpak dirs for steam. Running @PedroVH fork immediately worked for me. 

Would be great to see this merged in some capacity. 

(Tested appimage from fork releases)

JustinLex:
This fix is really important to me. I didn't realize switching to Flatpak Steam would break r2modman. I would love to see this PR reviewed and merged soon.

craigcabrey:
```
#!/bin/sh

sed -i '/#!\/usr\/bin\/env bash/aif [ -z $FLATPAK_ID ]; then\n  exec flatpak run com.valvesoftware.Steam $@\nfi' $HOME/.var/app/com.valvesoftware.Steam/data/Steam/steam.sh

r2modman.AppImage
```

Here is my janky script to make it work without any changes to r2modman. Put this & the r2modman AppImage in your PATH (e.g `~/.local/bin`). This just inserts a preample in `steam.sh` to re-exec in the flatpak context.

It needs to be re-inserted everytime since Steam detects the change and overwrites it (which also means a slow Steam startup).

rehashedsalt:
Are there any blockers to this PR? What's preventing this from being merged to main?

PR #1079, opened 2023-09-23
Add Palworld + Voices of the Void support, initial Unreal engine support

ethangreen-dev:
Voices of the Void is a game about hunting aliens and the wacky / spooky events that happen as a result.

UE4SS (https://github.com/UE4SS-RE/RE-UE4SS) is the standard modloader for a variety of Unreal engine games.

Shimloader is my WIP modloader-loader. It shims the game and UE4SS (the actual modloader) to allow Unreal to load blueprint (.pak files) and UE4SS (.lua) mods from external directories, thus enabling us to use the R2MM "linker" method of mod management.

Blueprint and eu4ss mods are respectively installed into:
- `PROFILE/shimloader/bp_mods` (mapped to `VotV/Content/Paks/LogicMods`)
- `PROFILE/shimloader/ue4ss_mods` (mapped to `VotV/Binaries/Win64/Mods`)

Config files are of `.modconf` extension and are installed into `PROFILE/shimloader/config`, which is then mapped to `VotV/Content` at runtime.

These paths are then injected into the game via the `--ue4ss-mods`, `--bp-mods` and `--config-dir` CLI args, which are then captured by the shimloader binary and used to setup win32 filesystem redirection logic.

The shimloader package can be found here:
https://thunderstore.io/c/voices-of-the-void/p/0xFFF7/votv_shimloader/


MythicManiac:
Nice! I skimmed this through briefly and noticed uninstallation seems to not be implemented for the shimloader at least. Whether or not the mods themselves get uninstalled properly is another question that would be good to double-check, since IIRC the way uninstalling works isn't the most reliable at the moment (I seem to remember that the uninstaller didn't necessarily have enough metadata available about what it's uninstalling to really be intelligent about it, but can't remember anymore)

Another suggestion I'd have is to set up a templated way to configure games to use this pattern (specifically the install rules), e.g. akin to how it's done with bepinex now.

I'll do a more thorough review once you've checked the uninstaller aspect, but could also be it's already handled & I just glanced over it so let me know if that's the case.

ethangreen-dev:
@MythicManiac 

I've been working through uninstalling for the better part of this morning and haven't yet found a good solution. Top-level binaries are cleaned up a-ok, but `shimloader/ue4ss_mods` and its contents are not removed. Blueprint mods (which contain just .pak files) can be uninstalled just fine though.

Similarly, ue4ss lua mods are also not currently installable. These are mods which contain script files (usually) in the following directory structure:
```
Author-ExampleVotVMod:
- VotV/Binaries/Win64/Mods/ExampleVotVMod/
- - Scripts/
- - - main.lua
- - - secondary_script.lua
- - - ...
```

I have yet to investigate implementing a custom install rule, but the goal here would be to filter by the `.lua` extension, resolve the second parent of the script file's path, and then move that directory into `shimloader/ue4ss_mods`. This would result in the following dir structure:

```
Author-ExampleVotVMod:
- Profile/Default/shimloader/ue4ss_mods/ExampleVotVMod
- - Scripts/
- - - main.lua
- - - secondary_script.lua
- - - ...
```

Shouldn't be awful to implement.

ethangreen-dev:
Made further tweaks to get *complete* shimloader uninstallation working. This necessitated the following changes:

- Added a `mod: ManifestV2` argument to the `ProfileInstallerProvider::installModLoader` abstract function. This makes it convenient for modloader install scripts to interact directly with state files.
- Added state file tracking support to the shimloader install script. This is done by instantiating a new `GenericProfilerInstaller` and using its `addToStateFile` function with the provided mod argument.
- Reworked `GenericProfileInstaller::uninstallState` to recursively remove directories whose contents were removed as a part of the uninstall process. This will not remove directories with contents not targeted within the state file.

ethangreen-dev:
Updates -

- Made the Shimloader installer completely generic
- Added Palworld support

These changes are entirely dependent on the following:
1. The creation of a `palworld` community.
2. Me submitting `unreal_shimloader` to the newly created community.

Oksamies (approved):
Reviewed together with the author.

src/r2mm/launching/instructions/instructions/loader/ShimloaderGameInstructions.ts
```diff
@@ -0,0 +1,21 @@
...
+    public async generate(game: Game, profile: Profile): Promise<GameInstruction> {
+        const shimloader = path.join(profile.getPathOfProfile(), "shimloader");
+
+        const luaDir = path.join(shimloader, "mod");
+        const pakDir = path.join(shimloader, "pak");
+        const cfgDir = path.join(shimloader, "cfg");
+
+        return {
+            moddedParameters: `--mod-dir "${luaDir}" --pak-dir "${pakDir}" --cfg-dir "${cfgDir}"`,
+            vanillaParameters: `--disable-mods`
```
MythicManiac, line 18: This should probably be the default operation mode as opposed to requiring a parameter to be passed in?

src/r2mm/installing/ConflictManagementProviderImpl.ts
```diff
@@ -54,7 +54,10 @@ export default class ConflictManagementProviderImpl extends ConflictManagementPr
             }
             const modState: ModFileTracker = yaml.parse(stateFileContents);
             modState.files.forEach(([key, value]) => {
-                if (mod.isEnabled()) {
+                // HACK: Never reset UE4SS-settings.ini, ever. Seriously. It should always be enabled.
```
MythicManiac, line 57: Why? Better yet, why is it in the state if we shouldn't track it in the state?

src/r2mm/installing/ConflictManagementProviderImpl.ts
```diff
@@ -80,6 +83,15 @@ export default class ConflictManagementProviderImpl extends ConflictManagementPr
...
                         await FileUtils.ensureDirectory(path.dirname(path.join(profile.getPathOfProfile(), file)));
+
+                        // HACK: Check if UE4SS-settings.ini exists and, if it doesn't, add it. If it does exist, DONT TOUCH.
+                        if (file.toLocaleLowerCase() == "ue4ss-settings.ini") {
+                            if (!await FsProvider.instance.exists(path.join(profile.getPathOfProfile(), file))) {
+                                await FsProvider.instance.copyFile(key, path.join(profile.getPathOfProfile(), file));
+                            }
+
+                            break;
+                        }
```
MythicManiac, line 94: This seems like the wrong location to implement this, there should be code somewhere which ensures all pre-launch configuration is set up right before a launch is attempted & that'd seem like where this should also be placed?

MythicManiac: Seems like I might be thinking of an EGS specific pre-launch procedure rather than something universal, not that an universal one would be hard to implement. The main point is that I don't think adding more "magic" here is a good idea & finding a place more suitable for it would be better.

src/model/game/GameManager.ts
```diff
@@ -562,6 +562,19 @@ export default class GameManager {
             "https://thunderstore.io/c/sailwind/api/v1/package/", "https://raw.githubusercontent.com/ebkr/r2modmanPlus/master/modExclusions.md",
             [new StorePlatformMetadata(StorePlatform.STEAM, "1764530")], "Sailwind.png",
             GameSelectionDisplayMode.VISIBLE, GameInstanceType.GAME, PackageLoader.BEPINEX, []),
+        new Game(
+            "Voices of the Void", "VotV", "VotV",
+            "", ["VotV-Win64-Shipping.exe"], "",
```
MythicManiac, line 567: Should the data folder name for VotV be defined? the mod linker now depends on it for ue4ss.dll & dwmapi.dll placement logic, so just checking if this is actually correct? (it's the empty string at the end atm)

src/r2mm/installing/ConflictManagementProviderImpl.ts
MythicManiac: Didn't seem necessary & refactored this out

src/r2mm/installing/ConflictManagementProviderImpl.ts
MythicManiac: Managed to refactor this out

src/model/game/GameManager.ts
ethangreen-dev: Yeah, that should be set to `VotV`. Will fix.

MythicManiac (approved):

Buckminsterfullerene02:
So what about supporting the automatic installation/uninstallation of basic asset replacement pak mods? I.e. not UE4SS ones that go into `Content/Paks/` folder and are "patch paks" by appending the pak name with `_P` (though usually in order to prioritise loader order with games with pakchunks they can be installed into `Content/Paks/~mods/`).

PR #1087, opened 2023-10-09
Skip '.git' directory in ModLinker processing.

Frenvius:
Added a check in the ModLinker class to exclude '.git' directories when looping over profile files. This is important to prevent the '.git' directory from being treated as a mod, when the mod profile is hosted on a github repository.


ebkr:
So I don’t particularly think that this is something that the manager needs to support.

It’s not a concern of the manager that you’re using Git to sync profiles instead of the in built code/file export feature.

The downside to using a git repo for this is that you’re also including the DLLs, where redistribution of said DLLs is often not allowed

Frenvius:
You have a point, considering this I will close a pull request.

PR #1089, opened 2023-10-15
Issue/1088/remove single dependency

### LINKED ISSUE #1088: [FEATURE REQUEST] - Option to uninstall/disable (fake) dependencies
**What problem does the feature request solve?**
There are often mods that come with dependencies that are not actually a necessity, but rather a recommendation of the mod creator, he deems it to be a must have or has created it himself. Those are fake dependencies and the dependency feature is used to force mods upon one.

Those fake dependencies are often unwanted mods and sometimes even break the game because they are faulty or incompatible with other already installed mods.

R2modman unfortunately does not let you uninstall or even disable a mod that is a (fake) dependency of another mod. You are stuck with the unwanted/faulty/incompatible mod that isn’t even needed or are forced to uninstall/disable the dependent mod along with it.

**Example mock-up (optional)**
You could be warned when a dependency is uninstalled/disabled and with the default behaviour being to uninstall/disable the dependent mods, but offer the option to keep the dependent mod installed/enabled. This mod can then be marked with a red exclamation mark that, when hovered over or clicked on, tells you that one or more dependencies are missing with an option to reinstall them.

**Additional information (optional)**
Actually I find this behavior of the r2modman quite patronising and for me it is a missing key feature. I literally just started modding and have installed less than 50 mods, but this problem already came up multiple times. Like it is, this mod manager is oversimplified in this regard and thus makes modding actually harder, not easier.

ebkr:
closes #1088 

image

image


MythicManiac (commented):
Two considerations:

1. Any chance you could reduce the amount of copypaste/boilerplate in the new code? It looks like the code is mostly the same between the old & new methods, only has minor variations.
2. I wonder how much does `getDisabledDependencies` affect rendering performance when large amounts of mods are installed? I know the installed mods view has rendering performance issues as is when lots of mods are installed, and this seems like it might potentially make that worse.

In terms of functionality I don't see anything wrong, but reviewing this is much more complex due to the code duplication than it would be if there was code re-use.

Issue #1115, opened 2023-12-04
[BUG] - Version links are broken


malleoz:
**Describe the bug**
Clicking on the button containing the version number for a mod sends you to a broken link.

**To Reproduce**
1. Expand mod card.
2. Click on the button with the version number.
3. Observe broken link page.

**Expected behavior**
Your browser should open to the mod page.

**Additional context**
In `LocalModList.vue` on line 182, you likely just need to add `v/` to the URL. i.e. instead of `https://thunderstore.io/c/lethal-company/p/BepInEx/BepInExPack/5.4.2100` it should be `https://thunderstore.io/c/lethal-company/p/BepInEx/BepInExPack/v/5.4.2100`
I don't have time to attempt to fix this on my own (especially because I have no clue how to build this project lol)

nikolateslax:
This only affects the Lethal Company part of Thunderstore, from what I can tell. r2modman was originally designed for Risk of Rain, and if you go over to the Risk of Rain side of Thunderstore, the links as generated by r2modman are accurate. This issue also affects the Thunderstore app.

MythicManiac:
This should be fixed in the next release

PR #1116, opened 2023-12-04
Fix broken links for mod version button

malleoz:
This fixes https://github.com/ebkr/r2modmanPlus/issues/1115

### DIFF
--- a/src/components/views/LocalModList.vue
+++ b/src/components/views/LocalModList.vue
@@ -179,7 +179,7 @@
                     <a class='card-footer-item' @click="enableMod(key)" v-else>Enable</a>
                 </template>
                 <a class='card-footer-item' @click="viewDependencyList(key)">Associated</a>
-                <Link :url="`${key.getWebsiteUrl()}${key.getVersionNumber().toString()}`"
+                <Link :url="`${key.getWebsiteUrl()}v/${key.getVersionNumber().toString()}`"
                       :target="'external'"
                       class="card-footer-item">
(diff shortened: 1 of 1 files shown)


MythicManiac:
Yeah this is a bit problematic since the URL schema is different for domainless communities. I also a bit awkward in general for the client to arbitrarily append content to URLs returned by the API, as the API has no visibility of that and leads to situations like this bug.

The "correct" fix would be one of the following:
1. Just use the versionless link to the package as that's guaranteed to be correct. This would mean direct version links aren't available in the manager.
2. Have the backend return version-specific links via the API and adjust the mod manager to use those in this scenario. This is not the best idea as it'd increase the payload size quite a lot (and it's already quite big)
3. Document a deterministic & future-proof-guaranteed way for clients to build direct links to packages themselves. This does not currently exist as a concept as far as website URLs go, so we'd also need to reserve an URL pattern for it.

I'd probably suggest doing option 1 for now as it's immediately available without being blocked by anything else and it doesn't really degrade the UX much.

Option 3 would be the best long-term, but that's blocked by requiring backend changes. It would be possible to contribute those changes, but you'd have to be pretty proactive about it since it's not immediately obvious on what the correct way to implement it would be and what URL pattern could should be reserved for it.

Let me know if you (or someone else) wants to go for option 3 and I'll help facilitate that, otherwise I'd suggest option 1 is done now.

MythicManiac:
Closing this as #1140 was merged

Issue #1039, opened 2023-06-16
[FEATURE REQUEST] - Add new Sort option, Last Updated, for Installed Mods


pjmichor:
**What problem does the feature request solve?**
Allow users to sort by mods that have been recently updated. This will help narrow down figuring out which mods may need updates after the game updates. It's a bit odd that the sort options weren't the same as the Online sort options.

**Example mock-up (optional)**
Current options for Installed Sort are:

- Custom (don't even know what this does)
- Mod Name
- Author Name

Would like to suggest adding

- Last updated

ebkr:
I'm closing this as the option to achieve the same effect is sorting by "Install date"

PR #1119, opened 2023-12-06
Add Last updated sorting option for locally installed mods

grahameron:
I thought #1039 was a good feature for the manager to have, so I decided to take a stab at implementing it. In order to get this change to work, I added a new sort type to the local mods sorting options enum, added a new sort case for that enum, and modified ManifestV2 to store dateUpdated and receive dateUpdated information from inputs that provide it.


src/r2mm/mods/ModListSort.ts
```diff
@@ -62,6 +62,8 @@ export default class ModListSort {
                 return a.getAuthorName().localeCompare(b.getAuthorName());
             case SortNaming.CUSTOM:
                 return 0;
+            case SortNaming.LAST_UPDATED:
+                return b.getDateUpdated().toISOString().localeCompare(a.getDateUpdated().toISOString());
```
MythicManiac, line 66: Any specific reason we're doing a string comparison instead of comparing dates directly?

MythicManiac (commented):
Thanks for the contribution! On the surface this looks good, but I'd have to test it before merging and that'll probably take at least a week or two before I get to it.

Bonus points for adding test cases, nice 👍 

Not sure what's going on with the commit history though but that's not a blocker

src/r2mm/mods/ModListSort.ts
grahameron: I don't remember off the top of my head, but I don't think it was working when I tried to compare dates directly. Not sure why that wouldn't work though, so maybe it wasn't working because I was doing something else wrong. I'll look at it later today and see if I can figure out why I did that.

And as for the commit history, I just had a series of consecutive brain farts lol

bubin7766: Oh man this is an update i've been waiting on forever! Crazy to see someone forked it out and tested it :)

PR #1123, opened 2023-12-07
Adding "A Robot Named Fight!" to the r2modman

BetaSprite:
I'm working on an Archipelago mod for "A Robot Named Fight!", but there is no mod distribution platform for that game yet.  The mod I'm working on is reliant on BepInEx, HarmonyLib, and Archipelago's C# dll, but I need the game to be in r2modman before I can do anything else, I believe.


MythicManiac:
Hey! As far as I could see the changes seemed correct, but I take it this is no longer relevant?

BetaSprite:
> Hey! As far as I could see the changes seemed correct, but I take it this is no longer relevant?

That's correct.  I had started this process before I got confirmation from the ARNF mod community about where they wanted to congregate for mod releases, and they were talking writing their own repository/client for distribution.  As a result, I think that adding it to r2modman would be more confusing for end users (unfortunately).

Thank you for confirming that I did it right, though!  This is all new to me, so I was just imitating what appeared to be correct as best as I could.

Issue #1048, opened 2023-07-21
[FEATURE REQUEST] - Back to Profile Selection Clickable on the Profile Screen


AgentX2006:
**What problem does the feature request solve?**
To change profiles you have to click Settings, then Profile, and finally Change profile. This is not only time consuming (It's faster to close and relaunch the app) but also not at all intuitive to have this buried in the settings.

**Example mock-up (optional)**
Just have it the same as the Back to game selection clickable that is found on the Profile selection screen.
![image](https://github.com/ebkr/r2modmanPlus/assets/51245403/0db4c0f2-7fcd-43c5-b384-8939e3802a33)

**Additional information (optional)**
Changing profiles really does not belong in the Settings. Imagine having to go into your browsers settings every time you wanted to go back a page.

GregariousJB:
Same thing with changing the game.

I was closing and reopening the program to switch games, not realizing you could change the game in the settings. What are these extremely important UI elements doing in a settings menu?

KlutzyBubbles:
Does someone know how to get the PR i made over a year ago recognized for merge? Seems like such an essential feature.

ebkr:
It’s not how I’d want to implement behaviour to go back to that screen.

The functionality is something I’d want to do however it’s definitely not essential. Most people load the manager, select a game, start modded, play the game, and they’re done. It’s really only annoying for people who set a game as the default option.

The left-hand side navigation doesn’t really have a section that would work for changing the profile, and adding a new section increases the content on the left and just makes it harder to read. It’s an easy solution but honestly it kind of sucks.

The ideal scenario is likely reworking the sidebar to be more inline with Thunderstore Mod Manager’s profile selection behaviour.

KlutzyBubbles:
Thank you for the swift reply with the explanation. TL;DR I still think the button is needed but wont push further if you don't want to.

> It’s not how I’d want to implement behaviour to go back to that screen.
>
> The functionality is something I’d want to do however it’s definitely not essential. Most people load the manager, select a game, start modded, play the game, and they’re done. It’s really only annoying for people who set a game as the default option.

I wonder if an extra button in the sidebar just to go back to profiles can be implemented until a more ideal solution can be made?

From what I’ve seen, a fair number of users seem to switch profiles more often, whether it be because someone accidentally loads the wrong profile or switching profiles for group switching or any other reason, and the ability to switch back in the same amount of clicks as getting into a profile should be default behavior.


[...]

PR #1124, opened 2023-12-07
Add change game / profile buttons

KlutzyBubbles:
Although the change game option is available in the settings, it is (for a lot of people) non existent buried in there.

This also solves #1048. The change game button uses the same code as the change game button in the settings, but i was unsure if the change profile button is the correct way of doing it, and can probably be removed if it isn't right.


MythicManiac:
Hey, thanks for the contribution! At a glance this seems alright to me, but we'd still need to test it before merging. In my case it's unlikely I can divert my attention here before January, but I'm not sure what's the situation of @ebkr

KlutzyBubbles:
Well its been a while, hoping it hasn't been forgotten?

KlutzyBubbles:
Any news on this? It seems like such a small thing that should already be implemented.

ebkr:
@KlutzyBubbles Apologies for the lack of communication on this PR.

There have been some updates since you made this PR that have tackled this in a different way.

image

I think this PR can be closed, but I'm happy to discuss if you have any other thoughts.

PR #1130, opened 2023-12-14
Support alternative Thunderstore CDN

anttimaki:


anttimaki:
Known issues:

1. I tested this by manipulating the hosts file of Windows. Despite my best efforts it seems the health test requests get cached sometimes. IDK but I think it happens on OS/hardware level since it seems to sometimes happen on regular browser too.

2. Seems the error logging in UtilityMixin breaks on TSMM because it claims the logging provider is not initiated. The logging is done on UtilityMixin in the hopes of avoiding code duplication between the projects. Maybe I need to move where things are done around a bit, but the logic itself shouldn't change, so this is good to review I think.

Related TS PR: https://github.com/thunderstore-io/Thunderstore/pull/960

anttimaki:
> Rather than a boolean toggle state, are you able to structure this so it has N potential options, and the exact one which should be used is passed to the code which needs to consume it?

Made it so.

src/utils/CdnUtils.ts
```diff
@@ -0,0 +1,55 @@
...
+export const getReachableCdn = async () => {
+    for await (const cdn of CDNS) {
+        if (await canReachCdn(cdn)) {
+            return cdn;
+        }
+    };
+
+    throw new R2Error(
+        "Can't reach content delivery networks",
+        `Both Thunderstore CDNs seem to be currently unreachable from
```
MythicManiac, line 38: Would replace `Both` with `All`, not a blocker

src/utils/CdnUtils.ts
```diff
@@ -0,0 +1,55 @@
...
+    const headers = {
+        "Cache-Control": "no-cache",
+        "Pragma": "no-cache",
+        "Expires": "0",
+    };
+    const params = {"disableCache": new Date().getTime()}
+    let res;
+
+    try {
+        res = await axios.get(url, {headers, params});
```
MythicManiac, line 21: We probably want to add a timeout of 5s or so here, otherwise the initial app startup could freeze for who knows how long as it's pending for CDN checks (given that CDN checks are blocking the app startup currently).

Axios doesn't have a timeout at all by default if I remember correctly.

MythicManiac (approved):
The startup time blocking is a real concern, but in general this looks good to me.

Pre-approving

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -288,7 +288,13 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
             callback(100, StatusEnum.SUCCESS, null);
             return;
         }
-        axios.get(combo.getVersion().getDownloadUrl(), {
+
+        let downloadUrl = combo.getVersion().getDownloadUrl()
+        if (this.getPreferredCdn()) {
+            downloadUrl += `?cdn=${this.getPreferredCdn()}`;
```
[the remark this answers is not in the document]
anttimaki: Now implemented with [helper utils](https://github.com/ebkr/r2modmanPlus/pull/1130/commits/bd64b188cc3b62eb992d847cb0f5aed03d122d7b)

src/utils/CdnUtils.ts
anttimaki: Fixed, was leftover from the original implementation that supported just one alternative mirror.

src/utils/CdnUtils.ts
anttimaki: Now implemented with [helper utils](https://github.com/ebkr/r2modmanPlus/pull/1130/commits/6f9bb959392a53500fbda3a16e92dcb8255569bc)

src/utils/HttpUtils.ts
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export const isNetworkError = (responseOrError: unknown) =>
+    responseOrError instanceof Error && responseOrError.message === "Network Error";
+
+/**
+ * Is the Error thrown by Axios request caused by a response timeout?
+ */
+export const isResponseTimeout = (error: unknown) =>
+    error instanceof Error && /timeout of (\d+)ms exceeded/i.test(error.message);
```
MythicManiac, line 38: These seem unused?

MythicManiac (approved):
Looks good to me

src/utils/HttpUtils.ts
anttimaki: They are. The whole file was "moved" from tsmm to r2mm. Didn't want to split the file and leave some of the stuff in tsmm only.

PR #1140, opened 2023-12-28
Remove version number from links leading to Thunderstore website

anttimaki:
Newer communities have a different URL structure for showing a specific version of a package than the older communities that use subdomains. Installed mods tried to link to a specific version, but the links were broken for the new communities, which confused users.

Always link to the non version specific page as a quick fix. A proper fix would require backend changes and/or an actual spec. One could sniff the base URL to determine how to link to a specific version, but this would be prone to break if the URL structure change again (e.g. once the new Thunderstore website is launched).

Since the button doesn't lead to the actual installed version anymore, use "website" as the label instead of the version number. This is in line with how the non-installed mods currently link to the website.

Since the version number needs to be shown somewhere, show it on the package title row instead.

### DIFF
--- a/src/components/views/LocalModList.vue
+++ b/src/components/views/LocalModList.vue
@@ -138,12 +138,15 @@
                             Disabled
                         </span>
                         <span class="card-title selectable">
-                            <template v-if="key.isEnabled()">
-                                {{key.getDisplayName()}} <span class="card-byline selectable">by {{key.getAuthorName()}}</span>
-                            </template>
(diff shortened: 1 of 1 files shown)


anttimaki:
* Fixes https://github.com/ebkr/r2modmanPlus/issues/1115
* Alternative to https://github.com/ebkr/r2modmanPlus/pull/1116, based on review comments

image

MythicManiac (approved):

PR #1145, opened 2024-01-01
Fix version number links leading to GitHub website

itsschwer:
The default `{url}{version}` generates broken links with GitHub urls, which may be used when a mod is imported locally (and has a GitHub website set in its manifest).

This is a quick fix to adjust any GitHub urls to point to `{url}/releases/tag/{version}` while preserving the default behaviour for non-GitHub urls.

URL | Before | After
--- | --- | ---
https://github.com/risk-of-thunder/R2API | https://github.com/risk-of-thunder/R2API4.4.0 | https://github.com/risk-of-thunder/R2API/releases/tag/4.4.0
https://thunderstore.io/package/tristanmcpherson/R2API/ | https://thunderstore.io/package/tristanmcpherson/R2API/5.0.5 | https://thunderstore.io/package/tristanmcpherson/R2API/5.0.5

### DIFF
--- a/src/components/views/LocalModList.vue
+++ b/src/components/views/LocalModList.vue
@@ -179,7 +179,7 @@
                     <a class='card-footer-item' @click="enableMod(key)" v-else>Enable</a>
                 </template>
                 <a class='card-footer-item' @click="viewDependencyList(key)">Associated</a>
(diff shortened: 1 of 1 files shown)


MythicManiac:
Closing this as this should be addressed by #1140 although slightly differently.

PR #1149, opened 2024-01-05
Improve pagination support for large page counts

MythicManiac:
Change the online mod list pagination to only render page buttons for pages in the near vicinity of the current page as well as the first and last page. This significantly improves the experience for games with a lot of mods available.

| Before | After |
|---|---|
| image | image |


anttimaki (approved):
Nice improvement.

PR #1151, opened 2024-01-10
Use caching to improve performance of large profiles

anttimaki:
When the local mod list is browsed vue seems to render the whole list again on some user actions. This causes three methods of the component to get called repeatedly:

- getMissingDependencies, which uses modifiableModList property, which is based on the mod list stored in VueX store. modifiableModList is used by other parts of the component as well, so I thought better not touch it at this time since figuring out unintended side effects would be a lot of work
- getThunderstoreModFromMod, which used the Thunderstore mod list stored in VueX, but now uses the list stored in the ThunderstorePackages. As far as I can tell both are updated in the Splash view and by the scheduled background process in UtilityMixin. Ergo this shouldn't break things, but this is the most significant functional change in this commit, and therefore most likely culprit should problems arise
- isLatestVersion, which did and still does use ThunderstorePackages for its shenanigans

So while this commit doesn't reduce the incessant function calls, it caches the results to a simple object to reduce required calculations. Effects were tested with a profile containing a mod pack with 109 mods. Completing the following tasks were timed (roughly and manually), with the accompanying results (original vs. cached):

- Initial rendering of the local mod list when moving from profile selection view: 7.0s vs. 5.4s
- Opening modal to disable a mod with two dependants: 4.4s vs. 1,2s
- Closing the modal without disabling the mod: 4.4s vs. 1.2s
- Opening modal to uninstall the same mod: 4.5s vs. 1.0s
- Uninstalling the mod: 15.8s vs 6.0s

(There might be further changes for optimizing the uninstall process, since it seems some stuff is done after each dependant is uninstalled, while it MIGHT be enough to do it just once in the end.)

For a small profile with 3 mods there's no noticeable difference between the performance of the old and new implementation.

### DIFF
--- a/src/r2mm/mods/ModBridge.ts
+++ b/src/r2mm/mods/ModBridge.ts
@@ -3,20 +3,25 @@ import ThunderstoreVersion from '../../model/ThunderstoreVersion';
 import ManifestV2 from '../../model/ManifestV2';
 import ThunderstorePackages from '../data/ThunderstorePackages';
 
-export default class ModBridge {
+interface CachedMod {
+    tsMod: ThunderstoreMod | undefined;
+    isLatest: boolean;
+}
 
-    public static getLatestVersion(mod: ManifestV2, modList: ThunderstoreMod[]): ThunderstoreVersion | void {
+interface ModCache {
+    [key: string]: CachedMod;
+}
+
+export default class ModBridge {
+    private static CACHE: ModCache = {}
+
+    public static getLatestVersion(mod: ManifestV2, modList: ThunderstoreMod[]): ThunderstoreVersion | undefined {
         const matchingMod: ThunderstoreMod | undefined = modList.find((tsMod: ThunderstoreMod) => tsMod.getFullName() === mod.getName());
         if (matchingMod === undefined) {
             return;
         }
         // Compare version numbers and reduce.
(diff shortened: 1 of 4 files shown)


anttimaki:
Forgot to implement the cache clearing when the package list is updated, so changed this to draft.

anttimaki:
Now clears cache when new package list is downloaded from the API.

src/r2mm/mods/ModBridge.ts
```diff
@@ -27,10 +32,45 @@ export default class ModBridge {
         const mod: ManifestV2 = new ManifestV2().fromReactive(vueMod);
         const latestVersion: ThunderstoreVersion | void = ModBridge.getLatestVersion(mod, ThunderstorePackages.PACKAGES);
         if (latestVersion instanceof ThunderstoreVersion) {
-            return mod.getVersionNumber()
-                .isEqualTo(latestVersion.getVersionNumber()) || mod.getVersionNumber().isNewerThan(latestVersion.getVersionNumber());
+            return mod.getVersionNumber().isEqualOrNewerThan(latestVersion.getVersionNumber());
         }
         return true;
     }
```
MythicManiac, line 38: Could this entire function be removed? Doesn't seem to be used outside of a single place now

MythicManiac (approved):
Seems to work as expected, local mod imports included (which I'm guessing you didn't explicitly account for)

Issue #995, opened 2023-03-21
[BUG] - Linux Appimage - Spaces in Profile Names


Aerodin:
**Describe the bug**
Use a profile with space(s) in it.  IE Valheim Mod Pack

**To Reproduce**
Create modded instance using BepinEx and BepinEx mods.  Name the profile something with spaces.

Ensure the parameters are set in Steam Valheim with the %command% at the end as instructed.

Run modded, and observe it start, then stop.  No error messages or other indicators as to why it's failing.

**Expected behavior**
Do the above but leave out the spaces for example ValheimModPack

Observe R2modmanplus execute Valheim with all mods loaded.

**Screenshots**

**Additional context**
Spaces can be a problem on Linux.  Profile names for app image or linux installs should enforce this particularly if passing arguments to Valheim.

PR #1154, opened 2024-01-11
Support spaces in arguments (eg spaces in file path)

Amzd:
This fixes spaces in file paths for me from r2modmanPlus side (bepinex also doesn't support this but those scripts dont update every time you restart r2modman so manual patches stay there). 

I think this might also fixe #995 as I have spaces in my profile name and it works but this needs someone more experienced with the project to test.

@NyaomiDEV tag because the script says to blame you if something does not work

### DIFF
--- a/public/linux_wrapper.sh
+++ b/public/linux_wrapper.sh
@@ -7,7 +7,7 @@ a="/$0"; a=${a%/*}; a=${a#/}; a=${a:-.}; BASEDIR=$(cd "$a"; pwd -P)
 
 R2PROFILE=""
 R2STARTSERVER=""
-args=""
+args=()
 
 while :; do
     case $1 in
@@ -26,11 +26,7 @@ while :; do
             if [ -z "$1" ]; then
                 break
             fi
-            if [ -z "$args" ]; then
-                args="$1"
-            else
-                args="$args $1"
-            fi
+            args+=( "$1" )
             ;;
     esac
     shift
@@ -38,13 +34,13 @@ done
 
 if [ -z "$R2PROFILE" ]; then
     echo "[R2MODMAN LINUX WRAPPER] Launching vanilla!"
-    exec $args
+    exec "${args[@]}"
 fi
 
 [ -n "$R2STARTSERVER" ] && exec "$BASEDIR/profiles/$R2PROFILE/start_server_bepinex.sh" $args || true
 
 if test -f "$BASEDIR/profiles/$R2PROFILE/run_bepinex.sh"; then
-    exec "$BASEDIR/profiles/$R2PROFILE/run_bepinex.sh" $args
+    exec "$BASEDIR/profiles/$R2PROFILE/run_bepinex.sh" "${args[@]}"
 else
-   exec "$BASEDIR/profiles/$R2PROFILE/start_game_bepinex.sh" $args
+   exec "$BASEDIR/profiles/$R2PROFILE/start_game_bepinex.sh" "${args[@]}"
 fi


NyaomiDEV:
It was so very much ago when I made this script, haha! Can't believe someone actually tagged me.

I should contribute more lately... sorry guys.

well, that social experiment is complete! (please don't tag me anymore lol)

MythicManiac (commented):
The change seems good on paper, but I'm not familiar enough with all the different quirks & considerations one would have to think of to properly review shell script changes, especially for the syntax being used here.

How confident are you that this should run on any system the script has previously been working on?

Happy to merge this once I have more confidence it actually does what it appears to be doing, I'm assuming you've tested this at least on your system?

Amzd:
`gamemoderun` uses the `"$@"` syntax which I think is the same as what I've changed it to but I use it with a named variable (`"$args[@]"`) and gamemoderun is probably the most used steam launch argument, at least on Linux.

https://github.com/FeralInteractive/gamemode/blob/8cea4c2418d8893e0352ac0718adca48478c33ea/data/gamemoderun#L9

Amzd:
For completions sake I tested on linux native Valheim installed at the windows default path (because of shared drive) which includes `Program Files (x86)` with spaces. 

<details>

<summary>The `start_server_bepinex.sh` has the same issue of not working with spaces so I updated it with the same bash array stuff I did in this PR:</summary>

---

Note that this script does not update every time you restart r2modman unlike the script this PR fixes, so it's not that bothersome to patch.

```sh
#!/bin/sh
# BepInEx running script
#
# HOW TO USE:
# 1. Make this script executable with `chmod u+x ./start_game_bepinex.sh`
# 2. In Steam, go to game's preferences and change game's launch args to:
#    ./start_game_bepinex.sh %command%
# 3. Start the game via Steam
#
# Note that you won't get a console this way
#
# NOTE: Edit the script only if you know what you're doing!

# Resolve base directory relative to this script
# Hopefully this resolves relative paths and links
a="/$0"; a=${a%/*}; a=${a#/}; a=${a:-.}; BASEDIR=$(cd "$a"; pwd -P)

# Special case: program is launched via Steam
# In that case rerun the script via their bootstrapper to ensure Steam overlay works
# if [ "$2" = "SteamLaunch" ]; then
#     echo "start_game_bepinex.sh: SteamLaunch"
#     cmd=( $1 $2 $3 $4 $0 )
#     shift 4
#     exec "${cmd[@]}" $@
#     exit
# fi

exec="$BASEDIR/valheim.x86_64"
rest=()
export DOORSTOP_ENABLE=TRUE
export DOORSTOP_INVOKE_DLL_PATH="$BASEDIR/BepInEx/core/BepInEx.Preloader.dll"
export DOORSTOP_CORLIB_OVERRIDE_PATH="$BASEDIR/unstripped_corlib"

# Allow to specify --doorstop-enable true|false
# Everything else is passed as-is to `exec`
while :; do
    case $1 in
        --doorstop-enable)
            if [ -n "$2" ]; then
                export DOORSTOP_ENABLE=$(echo "$2" | tr a-z A-Z)
                shift
            else
                echo "No --doorstop-enable value specified, using default!"
            fi
            ;;
        --doorstop-target)
            if [ -n "$2" ]; then
                export DOORSTOP_INVOKE_DLL_PATH="$2"
                shift
            else
                echo "No --doorstop-target value specified, using default!"
            fi
            ;;
        --doorstop-dll-search-override)
            if [ -n "$2" ]; then
                export DOORSTOP_CORLIB_OVERRIDE_PATH="$2"
                shift
            else
                echo "No --doorstop-dll-search-override value specified, using default!"
            fi
            ;;
        *)
            if [ -z "$1" ]; then
                break
            fi
            if [ -z "$launch" ]; then
                launch="$1"
            else
                rest+=( "$1" )
            fi
            ;;
    esac
    shift
done


export LD_LIBRARY_PATH="$BASEDIR/doorstop_libs:$LD_LIBRARY_PATH"
export LD_PRELOAD="libdoorstop_x64.so:$LD_PRELOAD"


# Run the main executable
# Don't quote here since $exec may contain args passed by Steam
if [ -n "$launch" ]; then
    echo "start_game_bepinex.sh: SteamLaunch"
    exec "$launch" "${rest[@]}"
else
    echo "start_game_bepinex.sh: "
    exec "$exec"
fi
```

---

</details>

NyaomiDEV:
tbf it looks good to me, but since they're using arrays instead of strings this bumps up the `bash` requirement (if the login shell is bash).

BTW: I'd recommend using https://github.com/koalaman/shellcheck to check for patterns that can lead to potential errors.

Also, this problem can be resolved in I think another way and it by using `${1@Q}`. Though I myself only recently discovered it, and it actually re-quotes the argument instead of keeping the default one. Again, it would probably bump the `bash` requirement though.

PR #1155, opened 2024-01-11
Improve performance when uninstalling multiple mods at once

anttimaki:
This is probably best reviewed commit by commit.

### DIFF
--- a/src/components/views/LocalModList.vue
+++ b/src/components/views/LocalModList.vue
@@ -100,6 +100,9 @@
                         @click="uninstallMod(selectedManifestMod)">
                     Uninstall
                 </button>
+                <span v-if="modBeingUninstalled" class="tag is-warning margin-top--1rem margin-left">
+                    Uninstalling {{ modBeingUninstalled }}
+                </span>
                 <button v-if="dependencyListDisplayType === 'view'" class="button is-info"
                         @click="selectedManifestMod = null">
                     Done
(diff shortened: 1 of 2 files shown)


anttimaki:
Screenshot showing the disabled button and the indicator that shows which mod is currently being uninstalled.
image

src/components/views/LocalModList.vue
```diff
@@ -438,47 +443,43 @@ import SearchUtils from '../../utils/SearchUtils';
...
                 this.$emit('error', result);
                 return;
             }
-            await this.$store.dispatch("updateModList", result);
-            const err = await ConflictManagementProvider.instance.resolveConflicts(result, this.contextProfile!);
-            if (err instanceof R2Error) {
-                this.$emit('error', err);
-            }
-            this.filterModList();
+            await this.updateModListAfterChange(result);
```
MythicManiac, line 482: Looks like there are a handful of code paths here which could in theory apply changes without refreshing the mod list, would it make sense to always refresh it before returning?

I'll push that change on top of this before merging since it seems a bit less prone to state management bugs that way.

MythicManiac: Done in b64c89cf306488d50b3e846aae32bbd9a5ae32a7

MythicManiac (approved):
Tested this locally and everything seems to work as expected

PR #1162, opened 2024-01-14
Fixing search to elevate exact matches

bnormoyle:
This always super annoys me when I'm searching for a specific mod with a name I already know. This reorders the exact matches to the top of the list.


MythicManiac (commented):
Did you test if this introduces noticeable lag on larger communities (e.g. lethal company)? There have been some search performance issues before, the main problem is how iteration based everything to do with the mod list is atm (overall, not just this PR change).

This looks like it shouldn't cause too significant issues, but perhaps it'd be worth restricting it to only run when the result set is already somewhat narrowed down? It should also be fairly trivial to do this in a single iteration rather than two like it does now.

PR #1170, opened 2024-01-18
Use more lenient timeouts when downloading package lists

anttimaki:


src/utils/HttpUtils.ts
```diff
@@ -10,24 +12,98 @@ const newAbortSignal = (timeoutMs: number) => {
...
+
+    const onDownloadProgress = (progress: ProgressEvent) => {
+        clearTimeout(rollingTimeout);
+        rollingTimeout = setTimeout(abort, transmissionTimeout);
+
+        if (typeof downloadProgressed === "function") {
+            // TODO: Progress percentage can't be calculated since the
+            // content length is unknown. Looks like this hasn't worked
+            // in a while.
+            downloadProgressed(0);
```
MythicManiac, line 90: Cloudflare seems to return the `Content-Length` header only if the response is resolved to an encoding format which is served by the origin. In this case, the encoding format the origin serves is `gzip` but cloudflare likely serves something like brotli, in which case it doesn't know the file size ahead of time. Easiest fix is to add `Accept-Encoding: gzip` to the request headers

MythicManiac (approved):

PR #1181, opened 2024-01-24
Store list of local mods with updates on VueX

anttimaki:
For a profile with 109 mods and 40 updateable mods, calling DownloadProvider's getLatestOfAllToUpdate takes a second or two.

- When entering local mod list, it gets called twice: once to show the banner about available updates, and once by DownloadModModal
- While in settings view once per second due to setInterval in SettingsItem (that might also be something to look into)
- Ridiculous number of times when the mods were updated via DownloadModModal. The update time for the example profile went from 315 seconds to 65 seconds (all downloads were cached both times)

Calling the method in VueX getter caches the value for subsequent calls, reducing the duration to milliseconds.


MythicManiac (approved):
LGTM

PR #1183, opened 2024-01-25
Improve profile code API error handling

MythicManiac:
This commit ensures the profile code API client is properly checking for and handling errors for both profile creation and import API calls.

Previously error handling was only done for profile exports and it wasn't handling some edge cases properly (apparently Axios does not always throw an error but instead returns it, this can occur at least if CORS checks fail).

Now the same error handling is used for both API calls and that error handling has been adjusted to account for the edge case mentioned earlier. In practice this means users should see somewhat more legible error messages for common failures such as 404 HTTP status codes as opposed to an arbitrary 'Cannot read properties of undefined'.

### DIFF
--- a/src/r2mm/profiles/ProfilesClient.ts
+++ b/src/r2mm/profiles/ProfilesClient.ts
@@ -6,45 +6,60 @@ const getProfileUrl = (profileImportCode: string): string => {
     return `https://thunderstore.io/api/experimental/legacyprofile/get/${profileImportCode}/`;
 }
 
-async function createProfile(payload: string): Promise<AxiosResponse<{ key: string }>> {
-    try {
-        return await Axios.post(
-            "https://thunderstore.io/api/experimental/legacyprofile/create/",
-            payload,
-            {
-                headers: { 'Content-Type': 'application/octet-stream' }
+function formatApiError<T>(e: T, genericTitle: string): R2Error | T {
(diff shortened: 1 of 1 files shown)


src/r2mm/profiles/ProfilesClient.ts
```diff
@@ -6,45 +6,60 @@ const getProfileUrl = (profileImportCode: string): string => {
...
         }
-        throw e;
+        return response;
+    } catch (e: any) {
+        throw formatApiError(e, errorTitle);
     }
 }
 
-async function getProfile(profileImportCode: string): Promise<AxiosResponse<string>> {
+function createProfile(payload: string): Promise<AxiosResponse<{ key: string }>> {
```
anttimaki, line 48: A bit confusing how you drop the async/await from this middle function and I don't quite know if it has any benefits given how it's currently used?

anttimaki (approved):

src/r2mm/profiles/ProfilesClient.ts
MythicManiac: async + await is slightly less efficient than simply returning a promise from a sync function is (due to the overhead caused by context swaps, especially if there are _a lot_ of other async code pending in the queue), in some niche cases significantly so. I'd not say it's necessarily a bad practice to use async + await here, but personally I'm in the habit of not using it unless necessary

PR #1188, opened 2024-01-30
Refactor LocalModList.getDisabledDependencies

anttimaki:
Improve the readability of the code quite a bit and performance a little bit.


MythicManiac (approved):
Nice, seems equivalent in behavior to me

PR #1189, opened 2024-01-30
Add LocalModCard

anttimaki:
Create a separate component for displaying a single locally installed mod. This might make caching the component easier, speeding up rendering larger profiles. If not, at least it cleans up a bit the rather extensive LocalModList component.


MythicManiac (approved):
👍 nice, changes look good on the surface at least, although I didn't do a super detailed review.

PR #1190, opened 2024-01-30
Clean up code related to ordering local mod list

anttimaki:
As far as I can tell none of the code removed in this commit is actually needed, or even used. Both the UI and mods.yml file seem to update just fine without them.


MythicManiac (approved):
Couldn't find any uses for these either on my double-check, sounds good enough to me

PR #1191, opened 2024-01-30
Improve performance of Manager.setAllModsEnabled

anttimaki:
- Skip processing the mods that already are in the desired state
- Update VueX store only once at the end, not after each mod, since this will trigger a costly rerendering (which will be addressed separately

### DIFF
--- a/src/pages/Manager.vue
+++ b/src/pages/Manager.vue
@@ -499,30 +499,46 @@ import CategoryFilterModal from '../components/modals/CategoryFilterModal.vue';
 		}
 
         async setAllModsEnabled(enabled: boolean) {
-            for (const mod of this.localModList) {
-                let profileErr: R2Error | void;
-                if (enabled) {
-                    profileErr = await ProfileInstallerProvider.instance.enableMod(mod, this.contextProfile!);
-                } else {
-                    profileErr = await ProfileInstallerProvider.instance.disableMod(mod, this.contextProfile!);
-                }
-                if (profileErr instanceof R2Error) {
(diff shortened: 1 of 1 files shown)


MythicManiac (commented):
Should probably wrap the logic to a try/catch to avoid state drift on a thrown error? Even though a lot of the code uses the paradigm where errors are returned instead of thrown, nothing actually guarantees this to be the case throughout all the code paths (this is also the main reason why I don't like this paradigm in JS; you need to be prepared to handle exceptions anyway and I'm not aware of any linters that could guarantee this is handled properly)

src/pages/Manager.vue
```diff
@@ -499,30 +499,47 @@ import CategoryFilterModal from '../components/modals/CategoryFilterModal.vue';
...
+                    if (update instanceof R2Error) {
+                        this.showError(update);
+                    } else {
+                        lastSuccessfulUpdate = update;
+                    }
+                }
+            } catch (e) {
+                const name = `Error ${enabled ? "enabling" : "disabling"} mods`;
+                const message = (e as Error).message;
+                this.showError(new R2Error(name, message));
```
MythicManiac, line 536: There's a `R2Error.fromThrownValue` class method (or something like that) available which you can use for this kind of stuff btw

MythicManiac (approved):
LGTM

src/pages/Manager.vue
anttimaki: Changed to use `fromThrownValue`.

PR #1192, opened 2024-01-30
Mount dependency string modal only when it's opened

anttimaki:
The modal rerenders when ever the local mod list updates. This is unnecessary since the list can be changed from e.g. LocalModList, while the modal can be opened only from the settings view.

Normally this wouldn't be much of a problem, but currently the method that's used to get the list of local mods does some extra stuff and is slow.


MythicManiac (approved):
LGTM

PR #1194, opened 2024-01-31
Trim the code when importing a profile

imatrisciano:
When sharing a profile code with friends sometimes extra spaces are copied alongside the id. If that's the case a 404 error will be generated. This commit adds a trim() to the code to remove any whitespaces.


NicholasHallman:
I was just about to fork a branch for this hehe

MythicManiac (approved):
Thanks for the contribution, this is something I keep forgetting to fix but definitely has been a problem

PR #1198, opened 2024-02-01
Defer mod list filtering when user types on search input

anttimaki:


src/components/views/LocalModList.vue
```diff
@@ -7,7 +7,13 @@
 
                     <div class="input-group input-group--flex margin-right">
                         <label for="local-search" class="non-selectable">Search</label>
-                        <input id="local-search" v-model='searchQuery' class="input margin-right" type="text" placeholder="Search for an installed mod"/>
+                        <DeferredInput
+                            @changed="(value) => searchQuery = value"
+                            id="local-search"
+                            class="input margin-right"
+                            type="text"
+                            placeholder="Search for an installed mod"
```
MythicManiac, line 15: Where do these actually end up in the DOM? I'm not familiar enough with vue to know, wondering if it creates a HTML wrapper element (e.g. div) or if it just applies them to the first component in the `<template>` block?

MythicManiac (approved):
LGTM

PR #1202, opened 2024-02-05
Refactor LocalModList search/sort

anttimaki:


src/components/DeferredInput.vue
```diff
@@ -1,19 +1,22 @@
...
-    internalValue = '';
+    internalValue = this.value;
 
     @Watch('internalValue')
     onInternalValueChange = debounce(this.emitChange, this.delay);
 
     emitChange() {
-        this.$emit('changed', this.internalValue);
+        this.$emit('change', this.internalValue);
     }
```
MythicManiac, line 20: Is this actually needed anymore now that the model binding is used?

src/components/views/LocalModList/SearchAndSort.vue
```diff
@@ -0,0 +1,132 @@
...
+export default class SearchAndSort extends Vue {
+    settings = new ManagerSettings();
+
+    get order() {
+        return this.$store.state.profile.order;
+    }
+
+    set order(value: SortNaming) {
+        this.$store.commit('profile/setOrder', value);
+        this.settings.setInstalledSortBy(value);
```
MythicManiac, line 24: I feel like the setting value change should be handled by the store rather than this component?

src/components/views/LocalModList/SearchAndSort.vue
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    get direction() {
+        return this.$store.state.profile.direction;
+    }
+
+    set direction(value: SortDirection) {
+        this.$store.commit('profile/setDirection', value);
+        this.settings.setInstalledSortDirection(value);
```
MythicManiac, line 33: Likewise here, shouldn't the store be the primary "API" for making these changes, and as such take care of syncing the state to where it should get synced?

Alternatively, the settings could subscribe itself to track the profile store I suppose.

src/components/views/LocalModList/SearchAndSort.vue
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    get disabledPosition() {
+        return this.$store.state.profile.disabledPosition;
+    }
+
+    set disabledPosition(value: SortLocalDisabledMods) {
+        this.$store.commit('profile/setDisabledPosition', value);
+        this.settings.setInstalledDisablePosition(value);
```
MythicManiac, line 42: Same as above

MythicManiac (approved):
Calling the new VueX store module the `ProfileModule` is a bit misleading IMO given that it primarily is concerned with the local mod list search & sort functionality.

I'm also assuming there's a reason you didn't move the settings updating code to the vuex store yet (perhaps due to the current game/settings not being as easily accessible there) so I don't mind if this version of the code gets merged as is; those concerns can be addressed later / aren't blockers.

So overall this looks good to me 👍

src/components/DeferredInput.vue
anttimaki: Yes, it's still needed to convey the changes to the `internalValue` to the external `value`.

src/components/views/LocalModList/SearchAndSort.vue
anttimaki: As per Discord discussions, to make this change properly we'd need to also have the active game and profile available in Vuex. That and these changes are now on my todo list but meanwhile I won't block these changes.

src/components/views/LocalModList/SearchAndSort.vue
```diff
@@ -0,0 +1,136 @@
...
+
+        this.$store.commit('profile/initialize', [
+            this.settings.getInstalledSortBy(),
+            this.settings.getInstalledSortDirection(),
+            this.settings.getInstalledDisablePosition()
+        ]);
+    }
+
+    async destroyed() {
+        this.$store.commit('profile/setSearchQuery', '');
```
anttimaki, line 76: Added this so the search query gets cleared when the component is unmounted, to match the existing functionality.

PR #1203, opened 2024-02-06
Update all - query performance

ebkr:
There was some code iterating dependencies in the previous version however it doesn't appear to actually be needed. The dependencies are still downloaded as expected and the visual indication of the mods that are required to be updated are identical.

Overall this saves around 5 seconds of hanging when uninstalling, disabling, and dragging items in the installed mod view.

The reason for the performance impact is due to the localModList state update also triggering a check to fetch available updates for the update all banner. We could probably further improve performance by not doing the check when the banner is no longer visible however the performance gain from that is neglible in comparison.

**For TMM**
As pointed out by @anttimaki this will need updating on the TMM implementation, however should be able to be copy/pasted without any changes for now.

### DIFF
--- a/src/r2mm/data/ThunderstorePackages.ts
+++ b/src/r2mm/data/ThunderstorePackages.ts
@@ -3,12 +3,14 @@ import Game from '../../model/game/Game';
 import ApiResponse from '../../model/api/ApiResponse';
 import ConnectionProvider from '../../providers/generic/connection/ConnectionProvider';
 import ModBridge from '../mods/ModBridge';
+import ThunderstoreVersion from 'src/model/ThunderstoreVersion';
 
 export default class ThunderstorePackages {
 
     public static PACKAGES: ThunderstoreMod[] = [];
     public static PACKAGES_MAP: Map<String, ThunderstoreMod> = new Map();
     public static EXCLUSIONS: string[] = [];
+    public static LATEST_VERSIONS: Map<String, ThunderstoreVersion> = new Map();
 
     /**
      * Fetch latest V1 API data and apply to {PACKAGES}
@@ -36,6 +38,14 @@ export default class ThunderstorePackages {
             map.set(pkg.getFullName(), pkg);
             return map;
         }, new Map<String, ThunderstoreMod>());
+
+        ThunderstorePackages.LATEST_VERSIONS = ThunderstorePackages.PACKAGES.reduce((map, pkg) => {
+            var latestVersion = pkg.getVersions().sort((a, b) => a.getVersionNumber().compareToDescending(b.getVersionNumber()))[0];
+            if (latestVersion != undefined) {
+                map.set(pkg.getFullName(), latestVersion);
+            }
+            return map;
+        }, new Map<String, ThunderstoreVersion>());
     }
 
     public static getDeprecatedPackageMap(): Map<string, boolean> {
--- a/src/r2mm/data/ThunderstorePackages.ts
+++ b/src/r2mm/data/ThunderstorePackages.ts
@@ -3,14 +3,12 @@ import Game from '../../model/game/Game';
 import ApiResponse from '../../model/api/ApiResponse';
 import ConnectionProvider from '../../providers/generic/connection/ConnectionProvider';
 import ModBridge from '../mods/ModBridge';
-import ThunderstoreVersion from 'src/model/ThunderstoreVersion';
 
 export default class ThunderstorePackages {
 
     public static PACKAGES: ThunderstoreMod[] = [];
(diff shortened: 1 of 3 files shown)


anttimaki (approved):
Tested and reviewed, LGTM.

anttimaki:
Here's some possible future iterations I came up with, any thougts?

**Version comparison**

Depending on the predicted future use cases, `compareToDescending` could be refined to:

1. A static method `(a, b) => a.getVersionNumber().compareToDescending(b.getVersionNumber())`, which could then be used like `pkg.getVersions().sort(ThunderstoreMod.sortVersionsToDescending)`
2. A method `pkg.getLatestVersion()`

Or either of those could be implemented separately, utilising `compareToDescending`. These are just for ease-of-use, and since there's currently no other use cases, this is not very important now.

**Caching**

To avoid having two "caches" (`PACKAGES_MAP` and `LATEST_VERSIONS`), we could do any of the following:

1. Add `ThunderstoreMod.latestVersion` field which is populated in `handlePackageApiResponse` in the same manner as `LATEST_VERSIONS` is done now
2. Like above, but use the existing `versionNumber` field, which `ThunderstoreMod` inherits from `Mod` via `ThunderstoreVersion`. This would be of type `VersionNumber` rather than `ThunderstoreVersion`, so there would be extra computation when `getLatestOfAllToUpdate` requires the latter.
3. Theoretically we could use `ThunderstoreCombo` in `PACKAGES_MAP` but this would require more refactoring elsewhere

anttimaki:
Oh I was supposed to mention this too, but forgot when I got distracted by the thoughts in the earlier comment.

* [This](https://github.com/ebkr/r2modmanPlus/pull/1203/files#diff-47656f57a1bf6f833edf87218c4093ffd33d6581f0d99c77896616b7cb86cc1cR108) could probably use the [`getCachedThunderstoreModFromMod`](https://github.com/ebkr/r2modmanPlus/blob/develop/src/r2mm/mods/ModBridge.ts#L58) I added recently?
* I feel some of the changes here might be redundant to [`isCachedLatestVersion`](https://github.com/ebkr/r2modmanPlus/blob/develop/src/r2mm/mods/ModBridge.ts#L62). IDK which place would make most sense for these, but having two somewhat competing implementations seems bad design? Something to think about in conjunction with caching-stuff mentioned above.

ebkr:
> [This](https://github.com/ebkr/r2modmanPlus/pull/1203/files#diff-47656f57a1bf6f833edf87218c4093ffd33d6581f0d99c77896616b7cb86cc1cR108) could probably use the [getCachedThunderstoreModFromMod](https://github.com/ebkr/r2modmanPlus/blob/develop/src/r2mm/mods/ModBridge.ts#L58) I added recently?

Yeah I agree, I'll change that

> I feel some of the changes here might be redundant to [isCachedLatestVersion](https://github.com/ebkr/r2modmanPlus/blob/develop/src/r2mm/mods/ModBridge.ts#L62). IDK which place would make most sense for these, but having two somewhat competing implementations seems bad design? Something to think about in conjunction with caching-stuff mentioned above.

What happens if the mod doesn't exist as part of the TS API (eg: locally imported)

anttimaki:
> What happens if the mod doesn't exist as part of the TS API (eg: locally imported)? I guess it'd return false, and then we try to ModBridge it and it wouldn't resolve to anything

I'm not sure I get what you're asking about, i.e. what happens *at what part of code*? The cached ModBridge implementation returns `undefined` instead of `ThunderstoreMod` when asking for an unknown mod, and `true` when asking if the unknown mod is at the latest version. Returning `true` for the latter would make sense to me, since we probably don't want to show "update this mod" when we have no idea if there's a newer version for it.

ebkr:
@anttimaki Should be good for a re-review

src/r2mm/data/ThunderstorePackages.ts
```diff
@@ -36,6 +38,14 @@ export default class ThunderstorePackages {
             map.set(pkg.getFullName(), pkg);
             return map;
         }, new Map<String, ThunderstoreMod>());
+
+        ThunderstorePackages.LATEST_VERSIONS = ThunderstorePackages.PACKAGES.reduce((map, pkg) => {
```
anttimaki, line 42: Calculating this now seems unnecessary use of resources?

anttimaki (commented):
The changes in the latest commit look good to me.

src/r2mm/data/ThunderstorePackages.ts
ebkr: Oops, good catch

PR #1206, opened 2024-02-08
Improve LocalModCard performance

anttimaki:


anttimaki:
Some manual performance testing with a profile that contains 282 mods. Numbers are in seconds and are averages from three rounds, except for the the built column, which was checked only once.

|                                                 | develop   | mod-typing | PR    | PR built |
| ------------------------------- | ---------- | ------------ | ---- | -------- |
| Initial list render                       | 13,5         | 17,3            | 17,4 | 14,3      |
| Open disable modal                 | 5,7           | 5,6              | 0 ‡    | 0 ‡        |
| Disable only BepInEx                | 25,4         | 22,5            | 14,1 | 12,3      |
| Enable BepInEx                         | 9,8           | 9,9              | 9,8   | 8,5       |
| Open uninstall modal               | 4,5           | 4,6              | 0 ‡    | 0 ‡        |
| Uninstall only BepInEx              | 22,0         | 17,5            | 10,2 | 8,5       |
| Reinstall BepInEx                      | 11,6         | 11,1            | 11,2 | 10,0     |
| Disable BepInEx & dependants | N/A †       | N/A †           | 46,9  | 32,3     |

‡ = modal opens up practically instantly
† = Takes ridiculous amount of time. Had this run for 16 minutes while on lunch and it didn't finish in that time. Also the background update of package list gets triggered every 5 minutes, which slows both processes down even more

Summary: the changes depend on earlier branches in this stack, which slow down the initial rendering a bit. As a tradeoff it significantly improves performance of some operations. Staring just the numbers can be a bit misleading though, since all use cases include the initial render, while someone disabling BepInEx and all the mods depending on it seems an unlikely scenario. On the other hand initial rendering is done only once, while user might e.g. open the modals multiple times in a session.

src/components/views/LocalModList/LocalModCard.vue
```diff
@@ -67,7 +96,13 @@ export default class LocalModCard extends Vue {
     viewDependencyList() {
         this.$emit('viewDependencyList', this.mod);
     }
+
+    created() {
+        this.updateDependencies();
+    }
 }
+
+const dependencyStringToModName = (x: string) => x.substring(0, x.lastIndexOf('-'));
```
ebkr, line 105: This should be declared as an explicit function rather than an anonymous one. Aware there's not a difference in terms of execution however it makes it clear initially that it's a function.

Just a personal preference.

ebkr (approved):

src/components/views/LocalModList/LocalModCard.vue
anttimaki: Fair enough, I can change it when/if #1202 is merged.

MythicManiac (approved):
LGTM 👍

PR #1208, opened 2024-02-09
Replace mod exclusions URL with a constant

MythicManiac:
Replace the mod exclusions list URL with a constant as the same value is used for all the games.

This could perhaps be moved to the game definition directly, but that would need a larger change than this commit is intended as.

**Merge #1079 first!**

### DIFF
--- a/src/model/game/GameManager.ts
+++ b/src/model/game/GameManager.ts
@@ -8,6 +8,9 @@ import PathResolver from '../../r2mm/manager/PathResolver';
 import FileUtils from '../../utils/FileUtils';
 import * as path from 'path';
 
+
+const EXCLUSIONS = 'https://raw.githubusercontent.com/ebkr/r2modmanPlus/master/modExclusions.md';
+
 export default class GameManager {
 
     private static _activeGame: Game;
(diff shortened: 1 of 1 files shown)


src/r2mm/manager/ModLinker.ts
```diff
@@ -85,28 +104,36 @@ export default class ModLinker {
...
+                            throw new FileWriteError(
+                                `Couldn't copy file ${file} to ${game.displayName} directory`,
+                                err.message,
+                                `Try running ${ManagerInformation.APP_NAME} as an administrator`
+                            )
                         }
                     } else {
                         if ((await fs.lstat(path.join(profile.getPathOfProfile(), file))).isDirectory()) {
-                            if (!["bepinex", "bepinex_server", "mods", "melonloader", "plugins", "userdata", "_state", "userlibs", "qmods"].includes(file.toLowerCase())) {
+                            const exclusionsList = [
```
ebkr, line 130: You could declare this at the top of the file outside of the class if that's easier. It wouldn't need to be recreated every time this is hit.

ebkr (commented):
I think the diff is slightly off here

src/r2mm/manager/ModLinker.ts
MythicManiac: This was a part of #1079 but yeah agreed, didn't end up making it a blocker however.

ebkr (approved):

PR #1210, opened 2024-02-10
Update game install rule validation

MythicManiac:
Update the game install rule validation check to account for games which are bound to specific installer IDs instead of relying on install rules


Oksamies (approved):

Oksamies (approved):
Nice

PR #1216, opened 2024-02-12
Make DisableModModal more self-sustained

anttimaki:


anttimaki:
Changes in this PR allows running "disable BepInEx and all dependants" on a profile with 282 mods in under 20 seconds, whereas before the changes it took 40+ seconds.

While the changes in this particular PR doesn't seem to affect the speed of the first local mod list rendering, it's worth noting the effect this four PR stack has on that performance (numbers with same 282 mod profile and dev build):

* `develop` branch: ~16 seconds
* `update-all-query-performance` on top of `develop` branch: ~9 seconds
* This branch on top of `develop` branch: ~20 seconds
* `update-all-query-performance` on top of this branch on top of `develop` branch: ~13 seconds

In other words the changes in this stack seems to eat roughly half of the gains from `update-all-query-performance`. On the other hand the changes do speed different use cases after the list have been rendered significantly, and at least IMO cleans up the code quite nicely.

Anyway the question remains, are these changes something we're willing to merge? I don't think makes much sense for me to build on top of this stack before I have the answer to that. I'll try to see if I can improve the performance without messing with the same files, and if not, then do something unrelated to performance instead.

src/utils/Profile.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    await disableModsFromProfile(mods, profile, onProgress);
+};
+
+export const disableModsFromProfile = async (
+    mods: ManifestV2[],
+    profile: Profile,
+    onProgress?: (mod: ManifestV2) => void,
+) => {
+    // Disable mods on disk.
```
anttimaki, line 32: This should probably be wrapped in a try-catch as was done on #1191.

anttimaki:
Based on Discord discussion, changed the PR so that mod disabling is done via Vuex. Initial rendering of a profile with 282 mods now takes ~3 seconds, thanks to other changes that have been merged since the PR was first opened.

src/components/views/LocalModList/DisableModModal.vue
```diff
@@ -1,41 +1,75 @@
...
+    }
 
-    @Prop({required: true})
-    readonly dependantsList!: Set<ManifestV2>;
+    get mod(): ManifestV2 {
+        if (this.$store.state.modals.disableModModalMod === null) {
+            this.$emit('error', new R2Error(
+                'Error while opening DisableModModal',
+                'Mod not provided'
+            ));
```
MythicManiac, line 34: This should probably throw instead of just emit an error because this is also relied on functions such as `disableModIncludingDependants` which should fail if this fails. I'm not sure if whether or not that scenario can actually occur right now but it seems like something that could happen due to a race condition given the async context we're dealing with.

src/components/views/LocalModList/DisableModModal.vue
```diff
@@ -1,41 +1,75 @@
...
+            await this.$store.dispatch(
+                'profile/disableModsFromActiveProfile',
+                { mods, onProgress }
+            );
+        } catch (e) {
+            const err: Error = e as Error;
+            LoggerProvider.instance.Log(LogSeverity.ACTION_STOPPED, `${err.name}\n-> ${err.message}`);
+            this.$emit('error', e);
+        } finally {
+            this.$store.commit('closeDisableModModal');
```
MythicManiac, line 60: Would it make sense for this to call `this.onClose()` instead?

src/store/modules/ProfileModule.ts
```diff
@@ -87,4 +92,73 @@ export default {
...
+
+                // Update mod list status to mods.yml and Vuex.
+                const updatedList = await ProfileModList.updateMods(mods, profile, (mod) => mod.disable());
+                if (updatedList instanceof R2Error) {
+                    throw updatedList;
+                } else {
+                    lastSuccessfulUpdate = updatedList;
+                }
+            } finally {
+                if (lastSuccessfulUpdate.length) {
```
MythicManiac, line 152: I guess this could technically be a bug: if for some reason the length becomes 0 during our mutations, we should still update the local mod list. I don't think it matters in the current codebase but commenting on it because I'd see it safer to check against null/undefined than 0 length in scenarios like this. A lack of a value isn't the same as an empty value after all.

MythicManiac (approved):
Spotted a couple of potential improvement areas within the context of this PR directly, but nothing I'd consider a blocker myself.

src/components/views/LocalModList/DisableModModal.vue
anttimaki: It's intended as a "this should never happen" safety, since the whole component is mounted only if `isOpen` is true, which checks that the mod is actually provided. That being said, I guess throwing might be marginally better approach.

Changed.

src/components/views/LocalModList/DisableModModal.vue
anttimaki: I don't know. Hopefully it does, because I changed it.

src/store/modules/ProfileModule.ts
anttimaki: IIRC this was more or less copied from the Manager component. But I changed it `undefined` instead. This means that `ConflictManagementProvider.instance.resolveConflicts` never gets called if successful updates haven't been made. Hopefully that's the right thing to do.

PR #1217, opened 2024-02-14
Use cached mod methods in remaining files

Lordmau5:
Additionally adds a `@deprecated` JSDoc to the old methods and links to the new, cached ones


anttimaki (approved):
LGTM 👍🏻

PR #1219, opened 2024-02-15
Use Vuex when disabling all mods from settings page

anttimaki:


src/pages/Manager.vue
```diff
@@ -625,10 +613,14 @@ import CategoryFilterModal from '../components/modals/CategoryFilterModal.vue';
                     this.settings = (() => this.settings)();
                     break;
                 case "EnableAll":
-                    this.setAllModsEnabled(true);
+                    this.setAllModsEnabled();
```
MythicManiac, line 616: Should this also await?

MythicManiac (approved):
Looks good to me

PR #1224, opened 2024-02-16
Make Uninstall Buttons Red

Decryptu:
## Enhancements to Uninstall Buttons

In line with the suggestions discussed in [this thread](https://github.com/ebkr/r2modmanPlus/discussions/1223):

### Changes:
- **LocalModCard.vue**: Applied `is-danger` class to the "Uninstall" button for visual emphasis on action's risk.
- **UninstallModModal.vue**: Used `is-warning` for the "Uninstall All" button to distinguish between single and bulk uninstall actions.
- **custom.scss**: Updated to ensure `is-danger` class applies correct border and background color changes on hover, focus, and active states.

### Visuals:
- Uninstall button now red: Uninstall Button
- Uninstall All button in modal: Uninstall All Modal

### DIFF
--- a/src/components/views/LocalModList/LocalModCard.vue
+++ b/src/components/views/LocalModList/LocalModCard.vue
@@ -175,7 +175,7 @@ function dependencyStringToModName(x: string) {
         </template>
 
         <!-- Show bottom button row -->
-        <a @click="uninstallMod()" class='card-footer-item'>
+        <a @click="uninstallMod()" class='card-footer-item is-danger'>
             Uninstall
         </a>
 
--- a/src/components/views/LocalModList/UninstallModModal.vue
+++ b/src/components/views/LocalModList/UninstallModModal.vue
@@ -64,7 +64,7 @@ export default class UninstallModModal extends Vue {
             </div>
         </template>
         <template v-slot:footer>
-            <button class="button is-info"
+            <button class="button is-warning"
                     :disabled="isLocked"
                     @click="onUninstallIncludeDependents(mod)">
                 Uninstall all (recommended)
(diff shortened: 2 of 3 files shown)


ebkr:
To further the reply from the discussion on https://github.com/ebkr/r2modmanPlus/discussions/1223

I like the idea of having some distinguishing factor for uninstall however I do have some reservations:

1. The colour is too bright. It draws away focus from the other elements which I'd argue are far more important. It's kind of like having the manager say "hey, uninstall this mod!". I don't think muting the colours would even necessarily solve this, so it's potentially worth avoiding entirely on the normal mod list screen.
2. I'm not a huge fan of using `is-danger` to mark the uninstall button. `is-danger` is more for an irreversible action. The act of clicking `Uninstall` on the mod list view does not uninstall the mod unless it's a single item, but in that case I definitely see it as less of a problem as it's extremely easy to get it back from the online list. Even if the action is red in that scenario, it's not like there's a secondary prompt so colour alone isn't really a good visual indicator. It's also worth noting that the `Uninstall X only` is not marked as `is-danger` which would be identical.
3. The `Uninstall all` action isn't really a warning in this case as it's the finalization of the action. This is probably more severe than `Uninstall X only` as it will include dependent mods too.

Decryptu:
Feel free to share any color preferences you have in mind, and I can make the updates accordingly. Alternatively, I can experiment with some options and share screenshots for you to see how they might look. I've noticed that Bulma's color options seem limited [Bulma Colors](https://bulma.io/documentation/overview/colors/), but it might not be a significant issue, and we could stick with the original color scheme before the changes if that works better.

PR #1225, opened 2024-02-16
Use Vuex to manage error modal

anttimaki:
- Main benefit is that emitted errors no longer need to bubble all the way to the App.vue to get shown. This should prevent situations where refactoring components distrupts the emit chain and leaves the error unshown
- Refactors the modal to a single standalone component
- Fixes the odd naming where R2Error's "name" was stored in variable called "message", and "message" was stored in "stack" variable
- Unifies some logged messages from "name\n-> message" format to "[name]: message" format

### DIFF
--- a/src/store/modules/ErrorModule.ts
+++ b/src/store/modules/ErrorModule.ts
@@ -1,0 +1,40 @@
+import R2Error from "../../model/errors/R2Error";
+import LoggerProvider, { LogSeverity } from '../../providers/ror2/logging/LoggerProvider';
+
+interface State {
+    error: R2Error | null;
+}
+
+interface ErrorWithLogging {
+    error: R2Error;
+    severity: LogSeverity;
+    logMessage?: string;
+}
+
+export default {
+    namespaced: true,
+
+    state: (): State => ({
+        error: null,
+    }),
+
+    mutations: {
+        discardError: function(state: State): void {
+            state.error = null;
+        },
+
+        handleError: function(
+            state: State,
+            error: R2Error | ErrorWithLogging
+        ): void {
+            state.error = error instanceof R2Error ? error : error.error;
+
+            if (error instanceof R2Error) {
+                LoggerProvider.instance.Log(LogSeverity.ERROR, `[${error.name}]: ${error.message}`);
+            } else {
+                const msg = error.logMessage || `[${error.error.name}]: ${error.error.message}`;
+                LoggerProvider.instance.Log(error.severity, msg);
+            }
+        },
+    }
+}
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,25 +1,7 @@
 <template>
     <div>
-
-        <router-view @error="showError" v-if="visible"/>
-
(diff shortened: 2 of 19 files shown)


anttimaki:
- Will probably conflict with all other changes, so let me know when it's okay to merge this and I can take care of conflicts
- Requires changes to TSMM, current avaialble as a draft PR

MythicManiac (commented):
Haven't finished reviewing yet but immediate thought is (and has been for a few other PRs already): we should stop expanding the `ModalsModule.ts` to become a mega-monolith of 329 different modal use cases, most of the modals have nothing to do with each other so it's a bit silly to store their state in the same module.

I don't think it should be considered a blocker for this PR necessarily, but I don't like the trend I've seen with how the code has evolved in that module over time. It's not even namespaced so it could easily lead to unintentional conflicts in the future at this rate.

src/store/modules/ModalsModule.ts
```diff
@@ -1,17 +1,29 @@
...
+
 export default {
     state: (): State => ({
         downloadModModalMod: null,
+        errorModalError: null,
         isCategoryFilterModalOpen: false,
         isDownloadModModalOpen: false,
+        isErrorModalOpen: false,
         isGameRunningModalOpen: false,
     }),
```
MythicManiac, line 28: what's the benefit in tracking the error state & modal open state separately? as far as I can tell those two fields are supposed to always be in sync anyway, could we not just check if the error exists? would eliminate any possibility of the two desyncing & potentially leading to weird UI bugs down the line.

src/store/modules/ModalsModule.ts
```diff
@@ -38,6 +55,21 @@ export default {
             state.isDownloadModModalOpen = true;
         },
 
+        openErrorModal: function(
```
MythicManiac, line 58: I feel like this does more than just opens the modal so the name could be a bit misleading, would it make sense to just call this something like `handleError`?

MythicManiac (approved):
I'd say the improvement suggestions I left are more than just nitpicks, but I also don't consider them a blocker for merging as they could easily be addressed as separate PRs.

So overall LGTM

src/store/modules/ModalsModule.ts
anttimaki: Just to stay consistent with `downloadModModalMod` and `disableModModalMod` (not pictured). But I agree, it's unnecessary.

src/store/modules/ErrorModule.ts
```diff
@@ -0,0 +1,40 @@
...
+            state: State,
+            error: R2Error | ErrorWithLogging
+        ): void {
+            state.error = error instanceof R2Error ? error : error.error;
+
+            if (error instanceof R2Error) {
+                LoggerProvider.instance.Log(LogSeverity.ERROR, `[${error.name}]: ${error.message}`);
+            } else {
+                const msg = error.logMessage || `[${error.error.name}]: ${error.error.message}`;
+                LoggerProvider.instance.Log(error.severity, msg);
```
ebkr, line 36: This is brilliant 👍

ebkr (approved):

PR #1228, opened 2024-02-19
Move localModList from main Vuex store to ProfileModule

anttimaki:
"local" in the variable name refers to locally installed mods, i.e. the mods currently included in the profile. It therefore makes sense for the mod list to be part of the ProfileModule.


anttimaki:
Requires related changes to TSMM, currently in draft PR state.

MythicManiac (approved):
LGTM 👍

PR #1230, opened 2024-02-20
Duplicate active game and settings instance in Vuex

anttimaki:


src/model/game/GameManager.ts
```diff
@@ -586,26 +586,26 @@ export default class GameManager {
...
-    }
-
     static set activeGame(game: Game) {
         this._activeGame = game;
     }
 
+    // Used for loading game specific settings before game is selected.
+    static get defaultGame(): Game {
+        return this._gameList.find(value => value.internalFolderName === "RiskOfRain2")!;
+    }
```
MythicManiac, line 596: Why not just hardcode the pointer to this game instance? Why do we need to enumerate any lists? We know the exact value this should point to at build time, there's no need to leave room for uncertainty

src/components/SettingsLoader.vue
```diff
@@ -173,14 +173,12 @@ export default class SettingsLoader extends Vue {
...
-    if (defaultGame === undefined) {
-        throw new Error("GameManager.unsetGame() returned undefined");
+    // Don't trust the non-null asserted typing of GameManager.defaultGame.
+    if (GameManager.defaultGame === undefined) {
+        throw new Error("GameManager.defaultGame returned undefined");
     }
 
-    return defaultGame;
+    return GameManager.defaultGame;
 };
```
MythicManiac, line 182: And as there's no need for uncertainty, this entire snippet can be removed

src/pages/Manager.vue
```diff
@@ -644,8 +644,20 @@ import CategoryFilterModal from '../components/modals/CategoryFilterModal.vue';
...
+            // ensure the settings are loaded before LocalModList
+            // accesses visibleModList from Vuex store.
+            await this.$store.dispatch('profile/loadOrderingSettings');
+
+            // Reset the mod list to prevent the previous profile's list
+            // flashing on the screen while a new profile's list is loaded.
+            await this.$store.dispatch('profile/updateModList', []);
+
+            // Used by OnlineModView, called here for consistency.
+            this.$store.commit('modFilters/reset');
```
MythicManiac, line 660: Should this `await`?

MythicManiac (approved):
Looks good to me, but again going to question the need for having all of this functionality in the same VueX modules. In my opinion there are distinctly different functionalities being mixed together right now that have no reason to be as highly coupled as they are.

Anyway, nothing I'd consider a blocker for merging

src/model/game/GameManager.ts
anttimaki: I'm not entirely sure why this has been implemented like this earlier. Took a quick peek now at the git history but that didn't provide much information. I can change the implementation if we think that's safe to do, but that'll have to wait until after my vacation.

src/pages/Manager.vue
anttimaki: `.dispatch` (actions) need to await, `commit` (mutations) are synchronous methods.

anttimaki:
> going to question the need for having all of this functionality in the same VueX modules. In my opinion there are distinctly different functionalities being mixed together right now that have no reason to be as highly coupled as they are.

To clarify, are you referring to having the active game and settings in the root store? Or the new dispatchers in profile module? In either case, what would you suggest as a sensible structure?

src/model/game/GameManager.ts
ebkr: I'm fairly sure it was done this way because of uncertainty towards moving to a cloud based game list. If the game were to be removed then we'd return undefined rather than erroring on the `.internalFolderName` because `_gameList.RiskOfRain2` wouldn't exist in the result

src/pages/Manager.vue
```diff
@@ -644,8 +644,20 @@ import CategoryFilterModal from '../components/modals/CategoryFilterModal.vue';
...
+        async beforeCreate() {
             this.activeGame = GameManager.activeGame;
+
+            // Used by SearchAndSort, but need to be called here to
+            // ensure the settings are loaded before LocalModList
+            // accesses visibleModList from Vuex store.
+            await this.$store.dispatch('profile/loadOrderingSettings');
+
+            // Reset the mod list to prevent the previous profile's list
+            // flashing on the screen while a new profile's list is loaded.
```
ebkr, line 656: Does this actually solve it? The beforeCreate is async and I'm not sure if Vue even awaits for this to complete before rendering

ebkr (approved):

MythicManiac (approved):

src/pages/Manager.vue
MythicManiac: I'm under the assumption that everything that relies on this would get re-rendered if it changes regardless if there's a race condition or not, so it wouldn't be a problem?

anttimaki: I tested this by making the Vuex dispatch function `sleep` a bit if an empty array of mods was passed to it. As a result, when called in `beforeCreate`, it rendered the proper mod list first, and then once the sleep finished, rendered the view with empty mod list. So a totally artificial test and quite unlikely that earlier call with empty array would complete later than a later call with non-empty array. But theoretically possible since the first call isn't awaited, methinks?

PR #1243, opened 2024-03-04
Duplicate active profile in Vuex

anttimaki:
Continues the work started on #1230 

Intentionally left the following to separate PRs:
- Refactoring other components to access the profile via Vuex
- Cleaning up the global-ish settings variable in Profiles.Vue and using the settings store in Vuex instead
- Refactoring the modals in Profiles.Vue to separate components


MythicManiac (approved):
LGTM 👍

PR #1246, opened 2024-03-05
Handle errors thrown by profile export zip creation

anttimaki:
One identified source for errors when exporting a profile as a code is when the zip file is already open in another app. This doesn't affect export as a file operations, since the zip used there is timestamped. I assume the zip in code export is not to avoid bloating the data directory with multiple archives of the same profile over time.


MythicManiac (approved):
> I assume the zip in code export is not to avoid bloating the data directory with multiple archives of the same profile over time.

This assumption would be correct

PR #1266, opened 2024-03-13
Add TsModsModule for Vuex

anttimaki:
The module will hold the list of all mods available on Thunderstore API for the current active game, and provide helpers related to that data.


MythicManiac (approved):
Seems good on the surface, didn't verify the implementation to match 1:1 to the previous behavior when reviewing.

PR #1269, opened 2024-03-15
Fix deprecations

anttimaki:


anttimaki:
In addition to reviewing the code the actual idea behind the changes should be verified. Do we want to mark any mods with deprecated dependencies as deprecated? The old implementation implied that's what's supposed to happen, but then didn't quite work that way. I.e. after the changes people will see more mods deprecated than before (or likely won't see them, since they will be properly filtered out in the OnlineModList.

MythicManiac:
> In addition to reviewing the code the actual idea behind the changes should be verified. Do we want to mark any mods with deprecated dependencies as deprecated?

I suspect hiding them from listings when a dependency is deprecated but the main package isn't is overkill at least for now (as it'd be potentially disruptive & breaking backwards-compat), but we should ideally display that somehow on the listing (e.g. a yellow warning sign for issues with dependencies with a mouse-over tooltip that notes some of the dependencies are deprecated)

How big of an effort would implementing something like that be?

MythicManiac (commented):
Highly relevant to this change, there are some other considerations the current logic fails to account for. 

Lets start with noting that at least the following different "problems" can exist for any given package that's returned by the package list API:

- The package is deprecated
- The package has a dependency that's deprecated
- The package has a dependency that's unavailable (rejected, removed, or in another community)

The logic that deals with deprecations & indicating their status on the UI should ideally account for all of these problem scenarios in some fashion. Additionally, we might want to have other similar status attributes for packages in the future (for example, "verified" packages) which might also benefit from bubbling upwards (in the case of verified mods, perhaps it would be worth making a distinction between modpacks that only use verified dependencies and ones that don't)

All of the above taken into account, it's pretty clear that the way the system currently works is less than ideal. As for how much we should address in this specific PR already I'm not entirely sure.

I suspect it could be best to keep the behavior the same as it has been before if it's easy enough to do & do an improvement pass later on that adds support for multiple different statuses instead of a boolean deprecated status.

I think my thoughts condensed amount to: Marking mods deprecated that shouldn't be deprecated is worse than not marking mods deprecated that should due to dependencies.

Also just to illustrate an interesting potential issue, consider a package that gets flagged as deprecated due to a dependency marked deprecated, and then later on suddenly becomes undeprecated because the deprecated dependency is removed entirely. Right now there would be no warning sign to the users that the package is not in fact fine.

anttimaki:
> I suspect hiding them from listings when a dependency is deprecated but the main package isn't is overkill at least for now (as it'd be potentially disruptive & breaking backwards-compat), but we should ideally display that somehow on the listing (e.g. a yellow warning sign for issues with dependencies with a mouse-over tooltip that notes some of the dependencies are deprecated)
> 
> How big of an effort would implementing something like that be?

IDK but I'd suspect rather big since then we'd have to keep the old broken implementation (and refactor it to fit the other changes in this PR stack) and the new implementation and use one in some places and the new in others.

> I suspect it could be best to keep the behavior the same as it has been before if it's easy enough to do & do an improvement pass later on that adds support for multiple different statuses instead of a boolean deprecated status.

As above, IDK how easy doing that would be since other changes have been built on top of these in the past month.

> Marking mods deprecated that shouldn't be deprecated is worse than not marking mods deprecated that should due to dependencies.

I'm not sure what this refers to. I agree with the statement, but as far as I can remember, the changes in this PR doesn't cause false positives, but removes false negatives. So more relevant statement would be "Not marking mods deprecated that should be due to dependencies is worse than the code to actually do what it was intended to do".

PR #1270, opened 2024-03-15
Add subdirectory executable support to direct game runner

ethangreen-dev:
Simple change, this introduces nested subdir support to the direct game runner. This means that games defined within `GameManager.ts` can refer to executable path(s) as relative paths, rooted by the game directory itself.

This change is primarily intended to support the Nickel modloader, whose executable is located at `GAME/Nickel/Nickel.exe`.


MythicManiac (commented):
Any reason we couldn't change the game definition config to include an option for configuring the exe dirpath? That would seem like a more reasonable approach to me here

ethangreen-dev:
Heavily out of date, closing.

PR #1271, opened 2024-03-18
Fix Thunderstore subdomain change

MSchmoecker:
Thunderstore recently changed all old subdomains to the new scope community style, breaking the mod manager API requests.

The replacement was done with a regex. For the possibility a community has a different community name, it will still be broken. I only tested a handful of games to verify that the new url works.


anttimaki:
I think this was fixed on infra level so the code changes are not needed. Thanks for the PR though!

PR #1279, opened 2024-03-25
Prewarm mod cache

anttimaki:


MythicManiac (approved):
I'm not 100% certain about what's going on just from the context of this PRs changes (as they refer getters/setters defined in prior changes), but at least the goal of prewarming the cache seems like a good idea & at a high level I didn't spot any issues.

PR #1300, opened 2024-04-18
add: bopl-battle (BepInEx)

ethangreen-dev:

PR #1301, opened 2024-04-18
add support for five games

ethangreen-dev:
Blocked by #1300


anttimaki:
For anyone interested in these changes, please note that as of this writing, the communities for the new games aren't yet available on Thunderstore, so the changes are likely to do you little good. Hopefully this will be remedied later today.

anttimaki:
Update to the above: communities are now available on Thunderstore.

mathlover3:
> Update to the above: communities are now available on Thunderstore.

no they arent?

Mangochicken13:
> > Update to the above: communities are now available on Thunderstore.
> 
> no they arent? at least for the app

That's not what the message means, up on the site and in the latest release of r2mm are two seperate things. Again, patience is key; good things come to those who wait

PR #1304, opened 2024-04-19
Ignore error about double navigation to splash screen

anttimaki:
The error is silenced to reduce noise in Thunderstore Sentry logs.

I'm not entirely sure how the user managed to trigger the error, since it seems they pressed the button multiple times over several seconds. Their manager must have been really slowed down since I had to add some sleep time to actually be able to duplicate the issue.


anttimaki:
Well it seems the same issue occurs in other places too. Might make sense to fix this globally somehow rather than separately on each places where navigation happens.

anttimaki:
I think I'll want to rethink this.

PR #1321, opened 2024-05-06
Add ids to buttons (and anchors) in Profiles.vue

VilppeRiskidev:
- IDs were added to buttons (and anchors) in Profiles.vue page to make debugging/logging easier
  - `modal-` prefix was used for modals' buttons


anttimaki (approved):
LGTM

PR #1325, opened 2024-05-10
Refactor Delete Profile Modal by separating it from Profiles.vue

VilppeRiskidev:
- `DeleteProfileModal` has been separated from `Profiles.vue`
- `profileList` has been added to Vuex storage

### DIFF
--- a/src/components/profiles-modals/DeleteProfileModal.vue
+++ b/src/components/profiles-modals/DeleteProfileModal.vue
@@ -1,0 +1,82 @@
+<script lang="ts">
+import { Vue, Component } from 'vue-property-decorator';
+import { ModalCard } from "../all";
+import FileUtils from "../../utils/FileUtils";
+import R2Error from "../../model/errors/R2Error";
+import Profile from "../../model/Profile";
+import FsProvider from "../../providers/generic/file/FsProvider";
+
+
+let fs: FsProvider;
+
+@Component({
+    components: {ModalCard}
+})
+export default class DeleteProfileModal extends Vue {
+
+    async created() {
+        fs = FsProvider.instance;
+    }
+
+    get activeProfile(): Profile {
+        return this.$store.getters['profile/activeProfile'];
+    }
+
+    get isOpen(): boolean {
+        return this.$store.state.modals.isDeleteProfileModalOpen;
+    }
+
+    get profileList(): string[] {
+        return this.$store.state.profiles.profileList;
+    }
+
+    closeDeleteProfileModal() {
+        this.$store.commit('closeDeleteProfileModal');
+    }
+
+    async removeProfileAfterConfirmation() {
+        try {
+            await FileUtils.emptyDirectory(this.activeProfile.getPathOfProfile());
+            await fs.rmdir(this.activeProfile.getPathOfProfile());
+        } catch (e) {
+            const err = R2Error.fromThrownValue(e, 'Error whilst deleting profile');
+            this.$store.commit('error/handleError', err);
+        }
+        if (
+            this.activeProfile
+                .getProfileName()
+                .toLowerCase() !== 'default'
+        ) {
+            for (let profileIteration = 0; profileIteration < this.profileList.length; profileIteration++) {
+                if (this.profileList[profileIteration] === this.activeProfile.getProfileName()) {
+                    this.$store.commit('profiles/spliceProfileList', profileIteration);
+                    break;
+                }
+            }
+        }
+        await this.$store.dispatch('profile/updateActiveProfile', 'Default');
+        this.closeDeleteProfileModal();
+    }
+}
+
+</script>
+<template>
+    <ModalCard v-if="isOpen" :is-active="isOpen" @close-modal="closeDeleteProfileModal">
+
+        <template v-slot:header>
+            <p class="modal-card-title">Delete profile</p>
+        </template>
+        <template v-slot:body>
+            <p>This will remove all mods, and their config files, installed within this profile.</p>
+            <p>If this was an accident, click either the darkened area, or the cross inside located in the top right.</p>
+            <p>Are you sure you'd like to delete this profile?</p>
+        </template>
+        <template v-slot:footer>
+            <button
+                class="button is-danger"
+                @click="removeProfileAfterConfirmation()"
+            >Delete profile</button>
+        </template>
+
+    </ModalCard>
+</template>
--- a/src/components/profiles-modals/DeleteProfileModal.vue
+++ b/src/components/profiles-modals/DeleteProfileModal.vue
@@ -23,7 +23,7 @@ export default class DeleteProfileModal extends Vue {
         try {
             await this.$store.dispatch('profiles/removeSelectedProfile');
         } catch (e) {
-            const err = R2Error.fromThrownValue(e, 'Error whilst deleting profile with DeleteProfileModal');
+            const err = R2Error.fromThrownValue(e, 'Error whilst deleting profile');
             this.$store.commit('error/handleError', err);
         }
         this.closeDeleteProfileModal();
--- a/src/components/profiles-modals/DeleteProfileModal.vue
+++ b/src/components/profiles-modals/DeleteProfileModal.vue
@@ -1,11 +1,8 @@
 <script lang="ts">
 import { Vue, Component } from 'vue-property-decorator';
 import { ModalCard } from "../all";
-import FileUtils from "../../utils/FileUtils";
 import R2Error from "../../model/errors/R2Error";
-import Profile from "../../model/Profile";
 import FsProvider from "../../providers/generic/file/FsProvider";
-
 
 let fs: FsProvider;
 
@@ -13,13 +10,8 @@ let fs: FsProvider;
     components: {ModalCard}
 })
 export default class DeleteProfileModal extends Vue {
-
     async created() {
         fs = FsProvider.instance;
-    }
-
-    get activeProfile(): Profile {
-        return this.$store.getters['profile/activeProfile'];
     }
 
     get isOpen(): boolean {
@@ -34,27 +26,13 @@ export default class DeleteProfileModal extends Vue {
         this.$store.commit('closeDeleteProfileModal');
     }
 
-    async removeProfileAfterConfirmation() {
+    async removeProfile() {
         try {
-            await FileUtils.emptyDirectory(this.activeProfile.getPathOfProfile());
-            await fs.rmdir(this.activeProfile.getPathOfProfile());
+            await this.$store.dispatch('profiles/removeSelectedProfile');
         } catch (e) {
-            const err = R2Error.fromThrownValue(e, 'Error whilst deleting profile');
+            const err = R2Error.fromThrownValue(e, 'Error whilst deleting profile with DeleteProfileModal');
             this.$store.commit('error/handleError', err);
         }
-        if (
-            this.activeProfile
-                .getProfileName()
-                .toLowerCase() !== 'default'
-        ) {
-            for (let profileIteration = 0; profileIteration < this.profileList.length; profileIteration++) {
-                if (this.profileList[profileIteration] === this.activeProfile.getProfileName()) {
-                    this.$store.commit('profiles/spliceProfileList', profileIteration);
-                    break;
-                }
-            }
-        }
-        await this.$store.dispatch('profile/updateActiveProfile', 'Default');
         this.closeDeleteProfileModal();
     }
 }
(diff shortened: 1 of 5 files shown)


src/components/profiles-modals/DeleteProfileModal.vue
```diff
@@ -0,0 +1,93 @@
...
+
+
+let fs: FsProvider;
+
+@Component({
+    components: {ModalCard}
+})
+export default class DeleteProfileModal extends Vue {
+
+    //TODO: add to vuex store or something to update Profiles list correctly.
```
anttimaki, line 18: Sounds good to me. There's two ways that *might* work.

1. Have the vuex state contain the profile list, and an action that updates the list in state (i.e. doing what `updateProfileList` does)
2. Just have a getter method that returns the list (i.e. does what `updateProfileList` does, but just returns the list instead of storing it internally)

I think 2) might be the nicer way to do this IF:
- The reactivity works as expected, i.e. the UI updates always when the list on disk updates
- This doesn't become a performance hit (it might, if we end up reading from the disk every time something happens on the UI).

So I would recommend to spent a moment to check if 2) seems feasible, and falling back to 1) if it gives us trouble.

anttimaki, line 18: Also, this is one of those "getting sidetracked" things we discussed, so I would maybe recommend putting this branch on ice and starting by creating a PR that only contains the "move profile list to Vuex" change using the existing modals/implementation. Once that's done continuing on this work will be easier and a smaller PR.

src/pages/Profiles.vue
```diff
@@ -401,6 +405,10 @@ export default class Profiles extends Vue {
         this.removingProfile = true;
     }
 
+    openDeleteProfileModal() {
+        this.$store.commit('openDeleteProfileModal', this.selectedProfile);
```
anttimaki, line 409: This is an example of the poor type support of the version of Vuex we use: the second parameter is not actually supported but no warning is shown about this.

Anyway, since it looks like we end up creating a separate Vuex module for this view, I think it would make sense to store the `selectedProfile` there. One would expect to find such closely related things in the same place.

anttimaki (commented):
This on right track.

src/store/modules/ProfilesModule.ts
```diff
@@ -0,0 +1,25 @@
+interface State {
+    profileList: string[];
+}
+
+/**
+ * State for Profiles, i.e. list for profiles in a single game/community.
+ */
+export default {
```
anttimaki, line 8: Nitpick: another thing we've been trying to move away from is the default exports. So `export const ProfilesModule = {` would be preferred here. (IIRC Vue components require default export though.)

src/pages/Profiles.vue
```diff
@@ -650,15 +603,16 @@ export default class Profiles extends Vue {
...
         await fs.readdir(profilesDirectory).then(dirContents => {
-            dirContents.forEach(async (file: string) => {
-                if ((await fs.stat(path.join(profilesDirectory, file))).isDirectory() && file.toLowerCase() !== 'default' && file.toLowerCase() !== "_profile_update") {
-                    this.profileList = [...this.profileList, file].sort();
-                }
+            let promises = dirContents.map(async function(file) {
+                return ((await fs.stat(path.join(profilesDirectory, file))).isDirectory() && file.toLowerCase() !== 'default' && file.toLowerCase() !== "_profile_update")
+                    ? file : undefined;
             });
-        }).catch(() => { /* Do nothing */ });
```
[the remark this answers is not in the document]
VilppeRiskidev: first commit adds the whole updateProfileList()`, `with just `return;`, later one converts to this commented style, soooo... better not touch it, because it _may_ be critical?

src/components/profiles-modals/DeleteProfileModal.vue
```diff
@@ -0,0 +1,60 @@
...
+        return this.$store.state.profiles.profileList;
+    }
+
+    closeDeleteProfileModal() {
+        this.$store.commit('closeDeleteProfileModal');
+    }
+
+    async removeProfile() {
+        try {
+            await this.$store.dispatch('profiles/removeSelectedProfile', {fs: fs, store: this.$store});
```
anttimaki, line 31: - Is there any specific reason for passing the `fs` as a parameter? AFAIK the ProfilesModule can access the FsProvider directly internally. I think it would be clearer separation of concerns if the modal component didn't know about the existance of the file system operations at all. We want this component to be all about the UI
- Also it seems unnecessary to pass a reference to the store to the method of the store. I'll comment more about this on the ProfilesModule file

src/pages/Profiles.vue
```diff
@@ -650,15 +603,22 @@ export default class Profiles extends Vue {
...
 
     async updateProfileList() {
-        this.profileList = ["Default"];
         const profilesDirectory: string = this.activeProfile.getDirectory();
-        await fs.readdir(profilesDirectory).then(dirContents => {
-            dirContents.forEach(async (file: string) => {
-                if ((await fs.stat(path.join(profilesDirectory, file))).isDirectory() && file.toLowerCase() !== 'default' && file.toLowerCase() !== "_profile_update") {
-                    this.profileList = [...this.profileList, file].sort();
-                }
+        let profilesDirectoryContents = ['Default'];
```
anttimaki, line 607: This seems unnecessary since the value is straight out overwritten in the try-block and not used outside of the try-block at all.

src/pages/Profiles.vue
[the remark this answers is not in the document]
anttimaki: Hard to say since we don't know if the error it's intended to silence is something we actually want to silence. I assume you haven't been able to get this part to cause any erros? If so, we can probably go with this approach and show the error modal. We can always yeet it if we get reports of users getting unnecessary modals when the error could just be ignored.

src/store/modules/ProfilesModule.ts
```diff
@@ -0,0 +1,51 @@
...
+            state.profileList = profileList;
+        },
+    },
+    actions: <ActionTree<State, RootState>>{
+        async removeSelectedProfile({commit, rootGetters, state, dispatch}, params: {fs: FsProvider, store: Store<any>}) {
+            const activeProfile: Profile = rootGetters['profile/activeProfile'];
+            const path = activeProfile.getPathOfProfile();
+            const profileName = activeProfile.getProfileName();
+
+            try {
```
anttimaki, line 32: So I probably didn't mention this earlier, but I've been using an approach where the Vuex-methods just throw R2Errors (or just plain JS errors if we don't know what errors to expect and don't know what to use as name/suggestion parameters. The UI components then try-catch the operations they call and dispatch the `error/handleError`.

The reason is more philosophical than technical. I see showing the error modal as an UI concern, so it should be handled by the UI component. This (and other) Vuex actions are more business logic, and don't care that much about how the stuff is shown on UI.

So for consistency, use that approach here and in the actions you create in the future.

src/store/modules/ProfilesModule.ts
```diff
@@ -0,0 +1,51 @@
...
+                await FileUtils.emptyDirectory(path);
+                await params.fs.rmdir(path);
+            } catch (e) {
+                const err = R2Error.fromThrownValue(e, 'Error whilst deleting profile from disk');
```
anttimaki, line 36: In this particular case, I would throw this R2Error.

src/store/modules/ProfilesModule.ts
```diff
@@ -0,0 +1,51 @@
...
+                await params.store.dispatch('error/handleError', err);
+            }
+
+            if (profileName.toLowerCase() !== 'default') {
+                try {
```
anttimaki, line 41: In this particular case, I would just remove the try-catch block. On gut feeling, the code inside `try` is not that likely to throw errors, and if it does, they will be caught by the try-catch in the modal component.

src/store/modules/ProfilesModule.ts
```diff
@@ -0,0 +1,51 @@
...
+                    state.profileList = state.profileList.filter((p: string) => p !== profileName)
+                    await params.store.dispatch('profile/updateActiveProfile', 'Default');
```
anttimaki, line 43: In case the store parameter was passed to make this line work, this approach could be used: https://vuex.vuejs.org/guide/modules.html#accessing-global-assets-in-namespaced-modules

anttimaki (changes requested):
Good work on the requested changes, but some new things emerged as a result.

src/pages/Profiles.vue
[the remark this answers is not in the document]
VilppeRiskidev: Yep, let's go with that. 👍

src/pages/Profiles.vue
```diff
@@ -399,36 +379,8 @@ export default class Profiles extends Vue {
...
-                .toLowerCase() !== 'default'
-        ) {
-            for (let profileIteration = 0; profileIteration < this.profileList.length; profileIteration++) {
-                if (this.profileList[profileIteration] === this.activeProfile.getProfileName()) {
-                    this.profileList.splice(profileIteration, 1);
-                    break;
-                }
-            }
-        }
-        await this.setSelectedProfile('Default');
```
anttimaki, line 426: The new implementation doesn't call this line anymore. (Well, Vuex can't call component methods, and while child components can emit events to parents, we don't want to do that in this case.) This creates a bug that can reproduced like this:

1. Install some mods in default profile
2. Create a second profile and install different mods
3. Return to profile selection
4. Delete the second profile, "selected profile" automatically changes to default
5. Clicking "Select profile" takes you to profile mod listing, which shows the content of the second, now deleted profile instead of the default profile

I *think* best way to fix this would be something like:

1. Copy the implementation of `Profiles.setSelectedProfile` to `ProfilesModule` as a separate action method (but as discussed earlier, leave out the try-catch and the error modal stuff)
  * Do this as a separate action since other places will also be calling it later on
2. Instead of `removeSelectedProfile` action calling `profile/updateActiveProfile`, call the new action instead (with prewarm cache)
  * We could alternatively call the new action from `DeleteProfileModal.removeProfile`, but I'm like 60% sure the suggestion above is a bit nicer

anttimaki (changes requested):
The changes themselves passed my review, but I did some manual testing and detected a bug. It's actually a regression since I broke the same thing when implementing the "cache prewarming" a while back. Anyway, it should be easy to fix (but do test it yourself since I didn't.)

src/store/modules/ProfilesModule.ts
```diff
@@ -0,0 +1,51 @@
...
+            const profileName = activeProfile.getProfileName();
+
+            try {
+                await FileUtils.emptyDirectory(path);
+                await FsProvider.instance.rmdir(path);
+            } catch (e) {
+                throw R2Error.fromThrownValue(e, 'Error whilst deleting profile from disk');
+            }
+
+            if (profileName.toLowerCase() !== 'default') {
```
anttimaki, line 39: When `Default` profile is "deleted", new empty `Default` profile should be created instantly - and it does. However, when that profile is selected, it still shows the mods of the `Default` profile that was deleted. I think this if-clause might be the culprit. Either we should always call `setSelectedProfile` at the end of `removeSelectedProfile`; or we could add an else-clause here that calls just `profile/updateModListFromFile` and `tsMods/prewarmCache`.

src/store/modules/ProfilesModule.ts
```diff
@@ -0,0 +1,51 @@
...
+                state.profileList = state.profileList.filter((p: string) => p !== profileName)
+                await dispatch('setSelectedProfile', 'Default');
+            }
+        },
+
+        async setSelectedProfile({rootGetters, state, dispatch}, profileName: string) {
```
anttimaki, line 45: This is fine now, but for RenameProfileModal this is going to need to support the additional `prewarmCache` parameter which determines whether the last two dispatches are executed (cf. `Profiles.setSelectedProfile`).

This is because the operation is quite heavy for profiles with hundreds of mods, and we don't want to do it unnecessarily. You can easily test it by installing a few mod packs for Lethal Company and switching between the profiles in the list.

Anyway, it's not necessary for this PR so it's up to you whether you want to do it here or in the RenameProfileModal PR.

anttimaki (changes requested):
The bug still remains in one special case.

src/store/modules/ProfilesModule.ts
anttimaki: You went a little overboard with this. Now, in addition to always calling `setSelectedProfile`, you also always filter the removed profile name from the list of profiles. We *don't* want to do that if it was the default profile - the `Default` should always be visible on the list.

(IDK if there's a good rationale to treat the default profile differently, but at least the code base assumes that there's always a profile called Default, so it's nice for UI to reflect that. And that's how it has worked so far, so maybe we shouldn't change that in this PR.)

anttimaki (changes requested):
Another bug crept up.

src/store/modules/ProfilesModule.ts
VilppeRiskidev: should be ok now

anttimaki (approved):
LGTM 👍🏻

PR #1328, opened 2024-05-20
Refactor Rename Profile Modal by separating it from Profiles.vue

VilppeRiskidev:
- `RenameProfileModal` has been separated from `Profiles.vue`
- Actions `renameProfile` and `updateProfileList` have been added to `ProfilesModule`
- Old profile renaming functionality has been removed from `Profiles` page

### DIFF
--- a/src/components/profiles-modals/RenameProfileModal.vue
+++ b/src/components/profiles-modals/RenameProfileModal.vue
@@ -13,7 +13,14 @@ export default class RenameProfileModal extends Vue {
 
     @Watch('$store.state.profile.activeProfile')
     activeProfileChanged(newProfile: Profile, oldProfile: Profile|null) {
-        if (oldProfile === null || oldProfile.getProfileName() === this.newProfileName) {
+        if (
+            // Modal was just created and has no value yet, use any available.
+            this.newProfileName === "" ||
+            // Profile was not previously selected, use the new active profile.
+            oldProfile === null ||
+            // Avoid losing user's changes when the profile unexpectedly changes.
+            oldProfile.getProfileName() === this.newProfileName
+        ) {
             this.newProfileName = newProfile.getProfileName();
         }
     }
--- a/src/components/profiles-modals/RenameProfileModal.vue
+++ b/src/components/profiles-modals/RenameProfileModal.vue
@@ -1,5 +1,5 @@
 <script lang="ts">
-import { Vue, Component, Watch } from 'vue-property-decorator';
+import { Vue, Component, Ref, Watch } from 'vue-property-decorator';
 import { ModalCard } from "../all";
 import sanitize from 'sanitize-filename';
 import R2Error from "../../model/errors/R2Error";
@@ -9,6 +9,7 @@ import Profile from "../../model/Profile";
     components: {ModalCard}
 })
 export default class RenameProfileModal extends Vue {
+    @Ref() readonly nameInput: HTMLInputElement | undefined;
     private newProfileName: string = '';
 
     @Watch('$store.state.profile.activeProfile')
@@ -26,7 +27,17 @@ export default class RenameProfileModal extends Vue {
     }
 
     get isOpen(): boolean {
-        return this.$store.state.modals.isRenameProfileModalOpen;
+        const isOpen_ = this.$store.state.modals.isRenameProfileModalOpen;
+
+        if (isOpen_) {
+            this.$nextTick(() => {
+                if (this.nameInput) {
+                    this.nameInput.focus();
+                }
+            });
+        }
+
+        return isOpen_;
     }
 
     get profileList(): string[] {
@@ -73,7 +84,7 @@ export default class RenameProfileModal extends Vue {
         <template v-slot:body>
             <p>This profile will store its own mods independently from other profiles.</p>
 
-            <input class="input" autofocus v-model="newProfileName" />
+            <input class="input" v-model="newProfileName" ref="nameInput" />
 
             <span class="tag is-dark" v-if="newProfileName === '' || makeProfileNameSafe(newProfileName) === ''">
                 Profile name required
--- a/src/components/profiles-modals/RenameProfileModal.vue
+++ b/src/components/profiles-modals/RenameProfileModal.vue
@@ -1,0 +1,87 @@
+<script lang="ts">
+import { Vue, Component, Watch } from 'vue-property-decorator';
+import { ModalCard } from "../all";
+import sanitize from 'sanitize-filename';
+import R2Error from "../../model/errors/R2Error";
+import Profile from "../../model/Profile";
+
+@Component({
+    components: {ModalCard}
+})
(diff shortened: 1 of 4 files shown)


src/components/profiles-modals/RenameProfileModal.vue
```diff
@@ -0,0 +1,82 @@
...
+import R2Error from "../../model/errors/R2Error";
+
+@Component({
+    components: {ModalCard}
+})
+export default class RenameProfileModal extends Vue {
+    private newProfileName: string = '';
+
+    get isOpen(): boolean {
+        this.newProfileName = this.$store.getters['profile/activeProfileName'];
```
anttimaki, line 14: I'm a bit anxious reading this value from Vuex in here.

- isOpen can be called multiple times as things get rerendered. Especially if the modal is closed, this is wasted effort
- at least in theory this can lead to situations where `this.newProfileName` is overwritten with a value from Vuex e.g. while user is editing the name in the input

However I'm not entirely sure how to fix this:

- We can't update `newProfileName` in Vue lifecycle hook like `created()` since that won't update the value when user selects a different profile
- We can't use `get`/`set` computed values to access active profile name since we don't want to update the value as the user types in the input
- Some modals (e.g. DownloadModModal) stores extra parameters that are updated when the modal is opened in the ModalsProfile, but that's for modals that don't have a better place for storing the values, while here we have the ProfilesModule

Best I can come up with is this:

```typescript
    @Watch('$store.state.profile.activeProfile')
    activeProfileChanged(newProfile: Profile, oldProfile: Profile|null) {
        if (oldProfile === null || oldProfile.getProfileName() === this.newProfileName) {
            this.newProfileName = newProfile.getProfileName();
        }
    }
```

- We're watching for changes to `activeProfile` (we can't watch `activeProfileName` since it's a getter and not in the state)
- If `oldProfile` was null (profile was first selected) we update `this.newProfileName`
- Or if `oldProfile` matched `this.newProfileName` we update the value (we don't want to overwrite changes the changes the user has made)
- The part I don't like about this solution is that it somewhat exposes too much of the Vuex side to the Vue component

Thoughts?

src/components/profiles-modals/RenameProfileModal.vue
```diff
@@ -0,0 +1,82 @@
...
+
+    makeProfileNameSafe(nameToSanitize: string): string {
+        return sanitize(nameToSanitize);
+    }
+
+    doesProfileExist(nameToCheck: string): boolean {
+        if ((nameToCheck.match(new RegExp('^([a-zA-Z0-9])(\\s|[a-zA-Z0-9]|_|-|[.])*$'))) === null) {
+            return true;
+        }
+        const safe: string | undefined = sanitize(nameToCheck);
```
anttimaki, line 34: I know you only moved this code from elsewhere but: a) can this actually return undefined? b) should we call `this.makeProfileNameSafe` instead?

I guess it's a judgement call whether we want to make such changes while refactoring, so I'll leave it up to you.

src/components/profiles-modals/RenameProfileModal.vue
```diff
@@ -0,0 +1,82 @@
...
+        if (safe === undefined) {
+            return true;
+        }
+        return (this.profileList.find(
```
anttimaki, line 38: Judgement call: Using `.some()` instead of `find()` would clean this up a bit. Currently the indentation of the last closing parenthesis makes this a bit wonky.

anttimaki (changes requested):
Quite many comments but mostly minor ones. Also they're all new ones, so good job picking up the lessons from the previous PR!

Bonus cookie points if you can make the the input submit the changes by pressing enter instead of having to click the submit button with mouse.

src/components/profiles-modals/RenameProfileModal.vue
VilppeRiskidev: I think it's ok, can't come up with anything better and I think this is exaclty the kind of thing that `Watch` exists for.

src/components/profiles-modals/RenameProfileModal.vue
```diff
@@ -0,0 +1,82 @@
...
+<template>
+    <ModalCard v-if="isOpen" :is-active="isOpen" @close-modal="closeModal">
+
+        <template v-slot:header>
+            <p class="modal-card-title">Rename a profile</p>
+        </template>
+        <template v-slot:body>
+            <p>This profile will store its own mods independently from other profiles.</p>
+
+            <input class="input" v-model="newProfileName" ref="profileNameInput" />
```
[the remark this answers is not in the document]
VilppeRiskidev: doesn't work for some reason, I guess it'd have to be a `<dialog>`

src/components/profiles-modals/RenameProfileModal.vue
```diff
@@ -0,0 +1,87 @@
...
+import { ModalCard } from "../all";
+import sanitize from 'sanitize-filename';
+import R2Error from "../../model/errors/R2Error";
+import Profile from "../../model/Profile";
+
+@Component({
+    components: {ModalCard}
+})
+export default class RenameProfileModal extends Vue {
+    private newProfileName: string = '';
```
VilppeRiskidev, line 12: This still has a bug, where the newprofilename is empty when the modal is first opened (for example coming from the game selection). Doing `private newProfileName: string = this.$store.state.profile.activeProfile.getProfileName();` doesn't work either. I'll look into this, but if you can immediately think of a solution feel free to comment.

src/store/modules/ProfilesModule.ts
```diff
@@ -39,7 +40,39 @@ export const ProfilesModule = {
...
+                throw R2Error.fromThrownValue(e, 'Error whilst renaming a profile on disk');
+            }
+            await dispatch('profile/updateActiveProfile', params.newName, { root: true });
+            await dispatch('updateProfileList');
+        },
+
+        async updateProfileList({commit, rootGetters}) {
+            const activeProfile: Profile = rootGetters['profile/activeProfile'];
+            const profilesDirectory: string = activeProfile.getDirectory();
+            let profilesDirectoryContents = ['Default'];
```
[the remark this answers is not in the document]
VilppeRiskidev: cleaned up a bit, removed the `activeProfile` too.

anttimaki (approved):
@VilppeRiskidev: Uno reverse card! I'll mark the PR approved but I expect you to review it after the changes I've done:

- The PR had merge conflicts with `delete-profile-modal-refactor`. It almost seemed like you'd manually reproduced few of the commits from the parent branch (but missed one tiny bit). Anyways, don't do that since it makes reviewing difficult. Use git's rebase instead.
- Addressed the few comments that I left unresolved by adding new commits. Check them out (commit messages too) to see how I solved them
- Added a few other fixes, check them out too
- Please test this thoroughly, I was rushing the last changes a bit since this took longer than I anticipated 😅

VilppeRiskidev:
LGTM 👍

PR #1330, opened 2024-05-20
Mod loader recommended versions

ebkr:
BONEWORKS has the problem where the latest release of MelonLoader is incompatible with the game. This is warned against on the site, however there is no indicator in the manager.

Without blocking downloads of MelonLoader, we can specify versions to use inside the MOD_LOADER_VARIANT_RECORD.


anttimaki (approved):
LGTM

PR #1336, opened 2024-05-26
Sort by install date

ebkr:
### Overview
A fairly frequently requested feature - useful for people to work out things what has changed after performing an `Update all` action.

#### When installed before feature (no install date shown)
image

#### When installed after feature (install date shown)
image

#### Sorting with a mixture of before/after feature mod install dates.
image

If no date is found, default ordering will be applied.


anttimaki (approved):
LGTM

PR #1337, opened 2024-05-26
Improved online category filtering

ebkr:
All three filtering options are now present alongside each other in the filter modal. The user no longer needs to select only one filter behaviour type.

It's a fairly common criticism that the app doesn't match the filter behaviour on the site.

This is still somewhat the case however expands to support the actual behaviour that people are after, which is to filter on both include and exclude.

The existing behaviour was for:
- Contains at least one category
- Contains all categories
- Exclude category

These three are now present in the modal under tag selections as opposed to being a radio option.

### Screenshots
image

### DIFF
--- a/src/components/modals/CategoryFilterModal.vue
+++ b/src/components/modals/CategoryFilterModal.vue
@@ -4,39 +4,30 @@
             <p class="card-header-title">Filter mod categories</p>
         </template>
         <template v-slot:body>
-            <div class="input-group">
-                <label>Categories</label>
-                <select class="select select--content-spacing" @change="selectCategory($event)">
-                    <option selected disabled>
-                        Select a category
-                    </option>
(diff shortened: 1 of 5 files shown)


anttimaki (changes requested):
Code-wise LGTM, but on the UI when multiple categories are selected, first the x-buttons drop onto their own lines, and then categories extend outside the modal.

image

ebkr:
Good spot, this is now fixed:
image

The `is-delete` class comes with Bulma and replaces the content with the `X` so is being used instead of the font-awesome one. The Bulma one does however require that there's content beforehand, this is being done with a nbsp character. We could replace is with the font-awesome one however the alignment seems ever-so-slightly nicer this way.

anttimaki (approved):
LGTM

PR #1345, opened 2024-05-30
Use new Thunderstore API to fetch community's package list

anttimaki:


src/store/modules/TsModsModule.ts
```diff
@@ -148,14 +161,61 @@ export const TsModsModule = {
     },
 
     actions: <ActionTree<State, RootState>>{
+        async _fetchPackageListIndex({}): Promise<string[]> {
+            // TODO: update GameManager to contain new API endpoint (or read from elsewhere?)
```
anttimaki, line 165: Is "hardcore endpoint URL for each game separately" still a valid strategy?

anttimaki:
@VilppeRiskidev you could also review this, even if only to know what's going on.

src/store/modules/TsModsModule.ts
anttimaki: Added [a new commit](https://github.com/ebkr/r2modmanPlus/pull/1345/commits/0d5faa48d317b8156a199d2a2a88b82c10b179a3) that just replaces the URLs in `GameManager`. If we want to do something fancier, like only pass a community-slug to the `Game` constructor and build the API URL internally from that, the changes can easily be thrown away with the commit.

PR #1347, opened 2024-05-31
Add setting for toggling preferred Thunderstore CDN

anttimaki:


anttimaki:
@VilppeRiskidev could you review and test this locally?

MythicManiac (approved):
Looks like an improvement, nice work 👍

PR #1355, opened 2024-06-06
Separate Create Profile modal from Profiles.vue

VilppeRiskidev:
- TODO: clean up old create profile stuff from Profiles.vue, which is easier to do with the refactoring of the Import module as the two were somewhat convoluted.

### DIFF
--- a/src/components/profiles-modals/CreateProfileModal.vue
+++ b/src/components/profiles-modals/CreateProfileModal.vue
@@ -1,0 +1,89 @@
+<script lang="ts">
+import { Vue, Component } from 'vue-property-decorator';
+import { ModalCard } from "../all";
+import R2Error from "../../model/errors/R2Error";
+import sanitize from "sanitize-filename";
+
+@Component({
+    components: {ModalCard}
+})
+export default class CreateProfileModal extends Vue {
+    private newProfileName = '';
+    get isOpen(): boolean {
+        return this.$store.state.modals.isCreateProfileModalOpen;
+    }
+    closeModal() {
+        this.newProfileName = '';
+        this.$store.commit('closeCreateProfileModal');
+    }
+
+    get profileList(): string[] {
+        return this.$store.state.profiles.profileList;
+    }
+
+    makeProfileNameSafe(nameToSanitize: string): string {
+        return sanitize(nameToSanitize);
+    }
+
+    doesProfileExist(nameToCheck: string): boolean {
+        if ((nameToCheck.match(new RegExp('^([a-zA-Z0-9])(\\s|[a-zA-Z0-9]|_|-|[.])*$'))) === null) {
+            return true;
+        }
+        const safe: string = this.makeProfileNameSafe(nameToCheck);
+        return (this.profileList.some(function (profile: string) {
+            return profile.toLowerCase() === safe.toLowerCase()
+        }));
+    }
+
+    // User confirmed creation of a new profile with a name that didn't exist before.
+    async createProfile() {
+        const safeName = this.makeProfileNameSafe(this.newProfileName);
+        if (safeName !== '') {
+            try {
+                await this.$store.dispatch('profiles/addProfile', safeName);
+                this.closeModal();
+            } catch (e) {
+                const err = R2Error.fromThrownValue(e, 'Error whilst creating a profile');
+                this.$store.commit('error/handleError', err);
+            }
+        }
+    }
+}
+
+</script>
+<template>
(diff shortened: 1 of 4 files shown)


src/components/profiles-modals/CreateProfileModal.vue
```diff
@@ -0,0 +1,89 @@
...
+            return true;
+        }
+        const safe: string = this.makeProfileNameSafe(nameToCheck);
+        return (this.profileList.some(function (profile: string) {
+            return profile.toLowerCase() === safe.toLowerCase()
+        }));
+    }
+
+    // User confirmed creation of a new profile with a name that didn't exist before.
+    // The profile can be either empty or populated via importing.
```
VilppeRiskidev, line 39: oops, this comment should be removed

anttimaki (commented):
Didn't do a proper review yet, but the start seems as it should 👍🏻 

One thing you might want to consider, is to centralize the methods shared between multiple profile modals, e.g. `get profileList`, `makeProfileNameSafe`, `doesProfileExist`. See e.g. how `Profiles.vue` uses `ProfileMixin.vue`. That being said, the mixin adds a bit of complexity and requires jumping between files when reading the code, so it might not be worth it if only few methods are duplicated between two files. (On the other hand it's quite nice from maintenance perspective if r2mm and tsmm have different implementations of the component that uses the mixin, like Profiles does.)

src/components/profiles-modals/CreateProfileModal.vue
```diff
@@ -0,0 +1,89 @@
...
+    async createProfile() {
+        const safeName = this.makeProfileNameSafe(this.newProfileName);
+        if (safeName !== '') {
+            try {
+            document.dispatchEvent(new CustomEvent("created-profile", {detail: safeName}));
```
anttimaki, line 44: Firing an event caught elsewhere is a bit spaghetty and not ideal from readability and maintenance perspectives. Ideally everything would be handled by the Vuex method.

Based on a quick peek, it looks the event listener might not be related to creating a blank profile at all, but for importing one. As the idea was to separate the modals from having shared code to make things clearer and more maintainable, having this modal to fire the event might not make sense.

So if the event is not required for this modal, remove the dispatch. If this modal actually needs something done in the event listener, see if it can be moved to be a part of the Vuex method instead. (Unless there's a good reason to do otherwise.)

src/components/profiles-modals/CreateProfileModal.vue
```diff
@@ -0,0 +1,89 @@
...
+            await this.$store.dispatch('profiles/addProfile', safeName);
```
anttimaki, line 45: Nitpick: this block is misaligned.

src/components/profiles-modals/CreateProfileModal.vue
```diff
@@ -0,0 +1,89 @@
...
+</script>
+<template>
+    <ModalCard v-if="isOpen" :is-active="isOpen" @close-modal="closeModal">
+
+        <template v-slot:header>
+            <p class="modal-card-title">Create a profile</p>
+        </template>
+
+        <template v-slot:body>
+            <p>This profile will store its own mods independently from other profiles.</p>
```
anttimaki, line 64: Seems something is off style-wise. The input no longer has vertical white space, causing this help text and the validation tags below cling to it.

src/components/profiles-modals/CreateProfileModal.vue
```diff
@@ -0,0 +1,89 @@
...
+            <input
```
anttimaki, line 65: It would be nice for text input to be automatically selected so user can just start typing. I think we discussed this before the vacations but I don't remember if there was some problem with making it so. Not a blocker but possibly a low hanging UX improvement.

anttimaki (changes requested):
Few observations, hopefully quick to fix.

src/components/profiles-modals/CreateProfileModal.vue
VilppeRiskidev: autofocus is not working very well with vue, we _could_ probably do it with `refs`, but it gets a bit messy. probably would be best to figure out the best way to do it and apply it everywhere.

src/components/profiles-modals/CreateProfileModal.vue
anttimaki: Ideally some scoped CSS  would've been used, but I guess this works too.

anttimaki (approved):

PR #1359, opened 2024-06-10
Introduce ProfilesMixin and clean up some of Profiles.vue's code

VilppeRiskidev:


VilppeRiskidev:
Draft, no build errors, but modal functionality isn't working.

VilppeRiskidev:
More work done, importing works, updating doesn't, I'll try to fix that next. Didn't move anything to vuex yet.

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -0,0 +1,480 @@
...
+
+    // Handles the importing of the profile (doesn't matter if it is from file or code)
+    // When creating:
+    //      1. Reads and parses the mods to be installed
+    //      2. Creates an EventListener which gets triggered after the user has created a new profile. The callback then:
+    //          1. Downloads (and extracts) the mods to the profile
+    // When updating:
+    //      1. Reads and parses the mods to be installed
+    //      2. Creates an EventListener which gets triggered after the user has selected the old profile to be updated. The callback then:
+    //          1. Creates a temporary profile (called _profile_update)
```
anttimaki, line 130: Actually to me it seems more like:

1. Deletes the temporary profile (called _profile_update) if it exists (leftover from previous failed run or something like that?)
2. Downloads (and extracts) the mods to temporary profile
3. Deletes the old profile (including the folder on disk)
4. Renames temporary profile to the chosen profile's name

Which makes more sense, since I assume the point of having the temp folder is to not lose the original profile if something goes wrong when downloading/extracting the mods.

anttimaki (changes requested):
- Please split this into two separate PRs
- The first one should introduce the mixin and move the methods from the modal components into the mixin
  - I was thinking that the mixin would be for the modals, not for the Profiles.vue. It *might* make sense to use the mixin for Profiles.vue too, but it's hard for me to judge it currently due to the scope of this PR
- The second PR should be about separating the import modal from Profiles
  - This can add additional methods to the mixin if it makes sense

src/components/mixins/ProfilesMixin.vue
```diff
@@ -0,0 +1,141 @@
...
+        return (this.profileList.some(function (profile: string) {
+            return profile.toLowerCase() === safe.toLowerCase()
+        }));
+    }
+
+    makeProfileNameSafe(nameToSanitize: string): string {
+        return sanitize(nameToSanitize);
+    }
+
+    async setSelectedProfile(profileName: string, prewarmCache = true) {
```
anttimaki, line 51: As far as I can tell, even when checking the next PR in the stack, this method and all below it in this file, are only used by Profiles.vue. If this is the case, what is the rationale of moving them into the mixin? Seems we would get the cons (having the code split into multiple places) without the pros (code reusability).

This also applies to `appName` getter.

anttimaki (approved):
LGTM

PR #1365, opened 2024-06-14
Support custom uninstallation methods and implement it for ReturnOfModding

anttimaki:


anttimaki:
Marked as draft because:
- Need to discuss with Mythic if the PackageInstallerV2 approach is a desired one
- Need to verify with the community that I understood correctly the desired changes to (un)installation logic

MythicManiac:
Somewhat related to the comment here: https://github.com/ebkr/r2modmanPlus/pull/1388#issuecomment-2298650162, I think the best approach would be to add optional methods to the current interface (so type hint it as potentially being `undefined`) rather than creating a second interface. Reasoning being that it's less code changes required elsewhere in the codebase and easier to upgrade the optional methods to required methods once we feel everything is ready for it (just changing the type definition suffices, type checks then verify everything works as intended).

anttimaki:
> Somewhat related to the comment here: [#1388 (comment)](https://github.com/ebkr/r2modmanPlus/pull/1388#issuecomment-2298650162), I think the best approach would be to add optional methods to the current interface (so type hint it as potentially being `undefined`) rather than creating a second interface. Reasoning being that it's less code changes required elsewhere in the codebase and easier to upgrade the optional methods to required methods once we feel everything is ready for it (just changing the type definition suffices, type checks then verify everything works as intended).

@MythicManiac changes added in  a separate commit https://github.com/ebkr/r2modmanPlus/pull/1365/commits/280852c1b1c039953d48bbb0da244ce0e30e7fef. Note that this drops the abstract class approach since TypeScript seems to support optional methods in abstract classes poorly.

PR #1373, opened 2024-06-19
Remove dexie version bump requirement for new game additions

ethangreen-dev:
This PR primarily implements a way to track settings without requiring each game to exist as an independent table. This is done by, simply, moving everything to a new `games` table, by which each game is a row, indexable by setting identifier.

This also removes code that migrated the legacy settings store to v2, but don't fret, I've replaced it with migration code that converts v2 game tables to the new v3 row entries.


src/r2mm/manager/SettingsDexieStore.ts
```diff
@@ -24,66 +24,47 @@ export default class SettingsDexieStore extends Dexie {
...
-                const obj = this.createNewSettingsInstance();
-                await this.global.put({ settings: JSON.stringify(obj.global) });
-                await this.gameSpecific.put({ settings: JSON.stringify(obj.gameSpecific) });
-                return obj.global;
-            }
-        });
+        const global = await this.global.get(1);
+
+        // Create the global settings row if it does not already exist.
+        if (global == undefined) {
```
anttimaki, line 46: It's customary to use `===` even if the strict equality isn't needed.

src/r2mm/manager/SettingsDexieStore.ts
```diff
@@ -24,66 +24,47 @@ export default class SettingsDexieStore extends Dexie {
...
+
+        // If we don't have a settings row for this game, either put and return a new one or migrate.
+        if (game == undefined && !(await this.applyV2ToV3MigrationPlease())) {
+            let obj = this.createNewSettingsInstance();
+            await this.games.put({ identifier: this.activeGame.settingsIdentifier, settings: JSON.stringify(obj.gameSpecific) });
+
+            return obj.gameSpecific;
+        }
+
+        return JSON.parse(game!.settings);
```
anttimaki, line 67: Doesn't using the non-null assertion here hide the situation where `game` is undefined and the V2ToV3 migration was successfully done?

src/r2mm/manager/SettingsDexieStore.ts
```diff
@@ -130,19 +111,51 @@ export default class SettingsDexieStore extends Dexie {
...
+     * @returns True if the legacy game field was migrated, false if it was not.
+     */
+    private async applyV2ToV3MigrationPlease(): Promise<Boolean> {
+        const gameTable = this.tables.find(table => table.name == this.activeGame.settingsIdentifier);
+        if (gameTable == undefined) {
+            return false;
+        }
+
+        const transaction = async () => {
+            const legacySettings = await gameTable.toCollection().first();
```
anttimaki, line 145: Isn't this a read operation and therefore not really need to be inside the transaction? I.e. is the transaction needed at all if it's just one `put` operation?

src/r2mm/manager/SettingsDexieStore.ts
ethangreen-dev: Resolved, thanks for catching that.

anttimaki (commented):
I know this is a bit late on the second review round, but did you consider using [Version.upgrade](https://dexie.org/docs/Version/Version.upgrade()) to do all this in one fell swoop, thus not requiring run-time checkups?

src/r2mm/manager/SettingsDexieStore.ts
```diff
@@ -24,66 +24,54 @@ export default class SettingsDexieStore extends Dexie {
...
-        });
+        const identifier = this.activeGame.settingsIdentifier;
+        const game = await this.games.get(identifier);
+
+        if (game !== undefined) {
+            return JSON.parse(game.settings);
+        }
+
+        // Attempt to migrate.
+        if (await this.applyV2ToV3Migration()) {
```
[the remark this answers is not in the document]
ethangreen-dev: Resolved by replacing `applyV2ToV3Migration`  with `Version.upgrade` transaction.

src/r2mm/manager/SettingsDexieStore.ts
```diff
@@ -130,57 +118,62 @@ export default class SettingsDexieStore extends Dexie {
...
-                installedDisablePosition: itf.installedDisablePosition,
-                installedSortBy: itf.installedDisablePosition,
-                installedSortDirection: itf.installedSortDirection,
-                lastSelectedProfile: itf.lastSelectedProfile,
-                launchParameters: itf.launchParameters,
-                linkedFiles: itf.linkedFiles
-            }
+    /**
+     * Migrate the current game's legacy (v2) table to the v3 schema, if applicable.
+     *
```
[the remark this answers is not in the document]
ethangreen-dev: Good motivation to setup VSCode for realsies, thanks.

src/r2mm/manager/SettingsDexieStore.ts
```diff
@@ -24,66 +22,70 @@ export default class SettingsDexieStore extends Dexie {
...
+        // - `value`: (++id, settings)              <-- Global settings.
+        this.version(75).stores(store).upgrade(async (trans) => {
+            // Migrate the current game's legacy (v2) table to the v3 schema, if applicable - versions < 75.
+            //
+            // v2 (whose types we still utilize) stored game-specific settings in individual tables.
+            // This worked but required us to bump the Dexie store version every time we added a new game. Yuck!
+            // To get around this we have moved all game settings into the `games` table, where the key
+            // is the settings identifier of the game, and the value is a JSON string.
+            const ignore = ["value", "games"];
+            const legacyTables = trans.storeNames.filter(x => !ignore.includes(x)).map(x => trans.table(x));
```
anttimaki, line 41: IDK if I'm missing something here, but on my local env `trans.storeNames` contains only "value" and "games", so none of the existing game settings are migrated. Based on a very quick check I couldn't find any way to access the old tables via `trans`, although logging the object shows that it's at least on some level aware of the old tables. So should we loop over the `GameManager.gameList` here too?

src/r2mm/manager/SettingsDexieStore.ts
```diff
@@ -24,66 +22,70 @@ export default class SettingsDexieStore extends Dexie {
...
+                if (legacySettings === undefined) {
+                    continue;
+                }
 
-        // Add all games to store. Borked v2-3 locally
-        // Increment per game or change to settings.
-        this.version(74).stores(store);
+                // If the legacy game table exists AND it contains a valid settings field, write it to the games table.
+                // Then clear the contents of the legacy game table.
+                await this.games.put({ identifier: this.activeGame.settingsIdentifier, settings: JSON.stringify(legacySettings.settings) });
```
anttimaki, line 53: I think at this point `this.activeGame` is always RoR2 so the `identifier` should use some other value, either the old table name or the settingsIdentifier from `GameManager`?

anttimaki (changes requested):
For me this doesn't work as expected, see the comments. I guess valid approaches would be:
1. Fix this one way or another to actually persist existing settings upon migration
2. Clean out the ineffective data migration code, effectively just resetting the game specific settings
3. Fall back using the older approach which didn't use `Version.upgrade()`

src/r2mm/manager/SettingsDexieStore.ts
ethangreen-dev: Applied your suggestion, worked.

anttimaki (approved):
LGTM :shipit:

Tested locally and works for me:
- Migrating to the new schema and persisting the existing settings
- Creating a new settings row when a game is accessed for the first time
- Adding new games does not require a schema version bump

PR #1384, opened 2024-06-24
Separate Import Profile modal from Profiles.vue

VilppeRiskidev:


src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -0,0 +1,378 @@
...
+    updateProfile() {
+        document.dispatchEvent(new CustomEvent("created-profile", {detail: this.activeProfileName}));
+    }
+}
+
+</script>
+<template>
+    <ModalCard :is-active="isOpen" @close-modal="closeModal">
+
+        <template v-slot:header v-if="activeStep === 'IMPORT_UPDATE_SELECTION'">
```
anttimaki, line 267: This wasn't quite what I meant with my earlier comment. This if-else-else makes it difficult to find out what the actual rendered output for each step is. I was thinking more in line having something like:

```vue
<ModalCard :is-active="isOpen" @close-modal="closeModal">
  <template v-slot:header v-if="activeStep === 'IMPORT_UPDATE_SELECTION'">
    <!-- all IMPORT_UPDATE_SELECTION markup here -->
  </template>
  
  <template v-slot:header v-if="activeStep === 'FILE_CODE_SELECTION'">
    <!-- all FILE_CODE_SELECTIONmarkup here -->
  </template>
  
  <!-- And so on -->
</ModalCard>
```

anttimaki (changes requested):
Didn't review the code yet since there's some issues that need fixing:

- Import from code button seems to do nothing
- If I cancel the file selection dialog while importing from a file, the modal just stays closed (IDK how this worked before). Clicking the import profile button then does nothing
  - Leaving and returning to the profile view then instantly opens the modal again
- When updating a profile and selecting the profile X from the dropdown, the button still says "Update profile: Y"
- At least sometimes when selecting the profile from the dropdown, error is logged in console: `[vuex] unknown mutation type: profiles/setSelectedProfile`. I assume the import doesn't then do what it's expected

anttimaki:
As discussed in the daily:

- The comments above (2024-06-27) are fixed
- New usability issue: when updating a profile, the dropdown shows whatever profile happens to be first on the list, while the button text shows the actually selected profile. This ambiguity is confusing
- On a glance it seems there's more clean up that can be done. E.g. we might not need the event listeners anymore now that this is a single component, since we can probably just trigger the action when user creates/selects the profile. However, for clarity's sake such cleanup should be done in a separate PR

VilppeRiskidev:
> As discussed in the daily:
> 
> * The comments above (2024-06-27) are fixed
> * New usability issue: when updating a profile, the dropdown shows whatever profile happens to be first on the list, while the button text shows the actually selected profile. This ambiguity is confusing
> * On a glance it seems there's more clean up that can be done. E.g. we might not need the event listeners anymore now that this is a single component, since we can probably just trigger the action when user creates/selects the profile. However, for clarity's sake such cleanup should be done in a separate PR

Abovementioned usability issue fixed ✅

anttimaki:
@VilppeRiskidev an awkwardly handled niche case:

- Create a profile share code for game X
- Change to game Y and import the profile using the code
- Error modal pops up saying that the profile contained no mods that could be imported for this game
- When the error modal is closed, the profile modal remains open, except it doesn't work anymore

I don't remember if this worked before or if it's caused by the refactoring. Maybe check if it can be fixed quickly either way?

VilppeRiskidev:
> @VilppeRiskidev an awkwardly handled niche case:
> 
> * Create a profile share code for game X
> * Change to game Y and import the profile using the code
> * Error modal pops up saying that the profile contained no mods that could be imported for this game
> * When the error modal is closed, the profile modal remains open, except it doesn't work anymore
> 
> I don't remember if this worked before or if it's caused by the refactoring. Maybe check if it can be fixed quickly either way?

This now works similarly to how it used to work (= I fixed the breaking modal). May not be perfect UX, as it still creates the profile even if the import wasn't successful.

anttimaki (approved):
Aforementioned issue is fixed. There's a minor niche issue which I think has existed before the changes here so fixing it is outside of the current scope.

anttimaki:
Force pushed to fix merge conflicts.

PR #1385, opened 2024-06-25
Reset Vuex state when returning to game selection to free up memory

VilppeRiskidev:


anttimaki (commented):
I'm not sure what "start resetting" in the PR title refers to, but yes, this looks like a good beginning of what I had in mind. However, if you check the ticket, my original idea was to call a reset method on the main/index store, and let it handle resetting the substores. Is there a reason why this now directly calls reset on profile substore? Are you planning adding the resets of the other stores in this PR or a separate one?

src/store/modules/ProfileModule.ts
```diff
@@ -292,6 +302,11 @@ export default {
             commit('setDisabledPosition', settings.getInstalledDisablePosition());
         },
 
+        async reset({commit, rootGetters}) {
+            await rootGetters['settings'].setProfile('Default');
```
anttimaki, line 306: We can't access settings here if this is called every time in the create hook of GameSelectionScreen, as this will cause an error when the app is first opened, causing the settings to not be loaded (e.g. game list always rendered as list instead of cards).

There are multiple ways how to approach fixing this but I'd rather not dive into figuring out the best (or even a working) way to solve the issue, so give it a try and let's check it together if you come up empty.

VilppeRiskidev: Now the behaviour is "fixed", but IDK what'd be the best for the UX; resetting the profile everytime or not.

anttimaki: From the top of my head I'd say we don't need to reset the profile on profile selection screen unless the selected profile changes (which I guess means we override all the relevant values anyway, so no reset is needed in that case either). On the game selection screen always resetting the profile is ok methinks.

src/pages/Profiles.vue
```diff
@@ -115,7 +115,7 @@ export default class Profiles extends ProfilesMixin {
         const settings = await this.$store.getters.settings;
         await settings.load();
 
-        const lastProfileName = await this.$store.dispatch('profile/loadLastSelectedProfile');
+        const lastProfileName = "Default";
```
anttimaki, line 118: What's the rationale behind this change? It just plain breaks the existing behaviour, and I don't think we want to remove it (and if we did then there would be more cleanup to remove the dead code).

anttimaki (changes requested):
The changes shouldn't break remembering the last selected profile.

src/store/modules/ProfileModule.ts
anttimaki: And for the record, the current solution no longer causes the error so in that sense this issue is resolved.

anttimaki:
I'll approve this PR when #1410 is approved.

anttimaki (approved):
Though approved, I'll delay merging this for now since we want to keep the "release pipeline" clear. This should be squash-merged when the time comes.

PR #1400, opened 2024-07-21
Tale spire demo support

HolloFox:
TaleSpire has a "TaleSpire - Guest Edition" on Steam which is a separate version of TaleSpire under a different Steam ID. By adding Demo support, users of "Guest Edition" can use the mod manager for their version.


HolloFox:
out of date, will make new pull request via `ecosystem-schema`

PR #1409, opened 2024-08-14
Fix OSX CI

mindstormjak:
Github recently removed macos 11 from their actions runners.
This bumps the macos version to 12 which is the oldest supported by actions.

Also renamed the test workflow from build to test.


anttimaki:
@mindstormjak is the electron-builder update required for the CI pipeline fixes? I'd have more confidence merging this without those dependency updates.

mindstormjak:
> @mindstormjak is the electron-builder update required for the CI pipeline fixes? I'd have more confidence merging this without those dependency updates.

Yes, without the update it rans into this issue https://github.com/electron-userland/electron-builder/issues/6726

PR #1410, opened 2024-08-19
Add data resetting to vuex storage modules and use the functionality in page components

VilppeRiskidev:


src/store/index.ts
```diff
@@ -70,9 +70,17 @@ export const store = {
             return settings;
         },
 
-        // TODO: Add reset actions for other modules
         async resetProfileModule({dispatch}: Context) {
             return await dispatch('profile/reset');
+        },
+        async resetProfilesModule({dispatch}: Context) {
```
anttimaki, line 76: I don't see much value in having these wrappers that just dispatch another action here. Why I first suggested having the reset function in the index file was that we'd have a *single* reset method that calls all the reset methods of submodules (i.e. caller doesn't have to care what the reset does and call each one of these separately).

src/store/modules/ProfilesModule.ts
```diff
@@ -41,6 +44,10 @@ export const ProfilesModule = {
             await dispatch('setSelectedProfile', { profileName: 'Default', prewarmCache: true });
         },
 
+        async reset({commit}) {
```
anttimaki, line 47: I don't see much value in having these async actions that call the synchronous mutations when we could just call the mutations directly.

anttimaki (commented):
Note that the earlier PR has a bug that breaks settings loading. Depending on how you fix it might affect what makes sense in this PRs context.

anttimaki:
@VilppeRiskidev could you rebase `reset-profile-data-on-game-selection-screen` to the latest `develop` and this branch on top of it to bring things up-to-date and resolve merge conflicts. I'd rather review/test these at the same time.

src/store/modules/ProfileModule.ts
```diff
@@ -14,7 +14,6 @@ import ManagerSettings from '../../r2mm/manager/ManagerSettings';
 import ModListSort from '../../r2mm/mods/ModListSort';
 import ProfileModList from '../../r2mm/mods/ProfileModList';
 import SearchUtils from '../../utils/SearchUtils';
-import GameManager from "../../model/game/GameManager";
```
anttimaki, line 17: I think this whole first commit is leftover from the rebasing and shouldn't exist on this PR anymore (i.e. if this change is actually neede, it should be either on the earlier PR or the main commit of this PR).

src/store/index.ts
anttimaki: To reword my earlier comment:

- If it's possible to have *one* action on the index store, that calls reset on the substores, then that's the way I'd like this implemented
- If we need to call different resets on different components (as in GameSelectionScreen.vue and Profiles.vue in this PR), then it makes no sense route the calls through the main store. Instead, call the reset methods directly from the substores
- If none of the resets need to be async (i.e. they just call the commit), then we should yeet the wrapping dispatch methods and call the commits directly
  - However, for consistency's sake it's nice for each substore to implement an identical method, so if one actually needs the async call, then use actions/dispatch for all
  - Also if you find it reasonable to assume that one of the resets will need to do something async soon, then it's okay to future proof this. But I personally wouldn't do it if it's not needed.

anttimaki (changes requested):
Seems to work as expected but I'd prefer some cleanup still.

src/pages/GameSelectionScreen.vue
```diff
@@ -326,7 +326,8 @@ export default class GameSelectionScreen extends Vue {
         await this.$store.dispatch('checkMigrations');
         this.runningMigration = false;
 
-        await this.$store.dispatch('resetProfileModule');
+        await this.$store.dispatch('resetLocalState');
```
anttimaki, line 329: In my earlier review I recommended making either all the resets either async/dispatch/action or sync/commit/mutation. That is not a hard requirement, just a preference. However, now the PR mixes the two approaches, and IDK if that's intentional or an oversight. So from communication perspective it would be nice if you'd left a comment stating why you decided to do it this way.

src/store/index.ts
```diff
@@ -70,10 +70,14 @@ export const store = {
             return settings;
         },
 
-        // TODO: Add reset actions for other modules
-        async resetProfileModule({dispatch}: Context) {
-            return await dispatch('profile/reset');
-        }
+        async resetLocalState({commit}: Context) {
+            return await Promise.all([
```
anttimaki, line 74: Commits are synchronous methods so using Promise.all does nothing here. Also there's no need to return anything from this function.

src/store/modules/ModFilterModule.ts
```diff
@@ -1,4 +1,4 @@
-import { GetterTree } from 'vuex';
+import { ActionTree, GetterTree } from 'vuex';
```
anttimaki, line 1: ActionTree seems unused?

src/store/modules/ProfileModule.ts
```diff
@@ -120,7 +119,7 @@ export default {
     },
 
     mutations: {
-        resetLocalState(state: State) {
+        reset(state: State) {
             state.modList = [];
```
anttimaki, line 123: I just now noticed that this doesn't reset the `activeProfile`. Why's that?

anttimaki (changes requested):
- I'm now reviewing the two PRs as one so use your judgment on where to apply possible fixes
- **MAJOR BUG**: shows empty online mod list for communities while the loading of mod list works as expected
- The bug may or may not be related to the fact that the two branches are based on different HEADs (at least it seems that way to me). Seems at least one the branches is still using SettingsDexieStrore schema version 74? Since the PRs are chained, it would probably be best that the branches are also chained. It might be easiest just to combine the commits in the PRs into one commit/PR based on the latest HEAD?

src/pages/GameSelectionScreen.vue
VilppeRiskidev: In my understanding commits are for "directly" modifying the state, dispatches are for launching other commits (or dispatches). So, in this case we're launching a few data resets from index.ts, so a commit won't do it, but in other places we're just resetting one module at a time and thus using the commit directly. I might be wrong though,

src/store/modules/ProfileModule.ts
VilppeRiskidev: this was the thing that got broken (trying to load the settings on first load and so on). I think we want to keep the selected profile selected if the user returns to game selection and selects the same game that was selected previously.
IDK... It seems to break more stuff to try and reset it, than just not resetting it. I think the're no cases where persisting the selected profile would cause any issues. If you feel this is something to try and dig deeper and try to fix LMK, maybe even creating a separate task for that would be smart.

src/pages/GameSelectionScreen.vue
anttimaki: You're not wrong, although I would say that the main point of actions is to be able to execute "external" async code before/after doing the commits. (And yes, due to technical limitations mutation can't mutate the state of sub/sibling stores).

But I guess my point is somewhere along these lines: now we have one line of code on GameSelectionScreen calling this action, and then six lines defining the action in the index store. We could instead have just four lines of code calling the commits directly on the GameSelectionScreen, with the added benefit of not having to jump to the wrapper action to see what it does. And yes, this somewhat contradicts my earlier comment about doing it exactly like it's done here, but that was done under the assumption that it would avoid some repetition when we could call the wrapper from multiple places (but now we don't since we do a bit different thing on Profiles.vue).

All that being said, **I don't consider this a blocker** as long as the choice to do it this was is deliberate.

anttimaki (changes requested):
> MAJOR BUG: shows empty online mod list for communities while the loading of mod list works as expected

This issue still persist when I try to run this locally. Did you test this? Does it work for you? As I mentioned on my earlier comment on another issue, it's hard for me to know what's going on if you just push a new version and don't acknowledge my comments in any way.

> I think we can safely reset the profile, because there is the pre-existing functionality which reselects last selected profile when entering the profile selection screen.

I might have been wrong about this, since I now get errors when returning from manager view to profile selection view. Did you test this? Does it work for you?

anttimaki:
@VilppeRiskidev I added two new commits to the PR. Check the commit messages for details and review and test them please. I'm not sure if GitHub allows you to actually post a review since it's your PR so just leave a comment in that case.

VilppeRiskidev:
> @VilppeRiskidev I added two new commits to the PR. Check the commit messages for details and review and test them please. I'm not sure if GitHub allows you to actually post a review since it's your PR so just leave a comment in that case.

Makes sense, thanks! Tested and reviewed 👍

anttimaki (approved):

anttimaki:
I'll merge this to the earlier PR and will squash-merge it when the time is right.

PR #1422, opened 2024-08-29
Clean up ThunderstoreDownloaderProvider

anttimaki:
Remove references to settings and active game from the download provider, as these were mostly used to only access the setting telling whether to ignore the cached mod downloads. Instead pass the settings as a boolean argument.

Implement related changes to provider implementation and their callsites.

In downloadAndSave, switch to checking the variable for ignoring the cache first, before checking if the mod actually exists on the disk, as this might reduce the number of unnecessary file operations.


VilppeRiskidev (approved):
LGTM.

PR #1426, opened 2024-09-03
Prevent closing Profile importing modal when a Profile is being imported

VilppeRiskidev:
Closing the modal while importing is in progress used to cause confusing UX, as the data state changes unexpectedly. This change prevents closing the modal.


anttimaki:
@VilppeRiskidev this still leaves the close modal icon visible but broken on top right corner of the screen.

anttimaki (approved):
👌🏻

PR #1430, opened 2024-09-07
Profile import summary

ebkr:
This PR aims to allow users to see which mods are going to be installed when importing a profile

image

image

### DIFF
--- a/src/components/profiles-modals/ImportProfileModal.vue
+++ b/src/components/profiles-modals/ImportProfileModal.vue
@@ -25,12 +25,13 @@ import ZipProvider from "../../providers/generic/zip/ZipProvider";
 import ThunderstoreMod from "../../model/ThunderstoreMod";
 import ThunderstoreVersion from "../../model/ThunderstoreVersion";
 import ManagerInformation from "../../_managerinf/ManagerInformation";
+import OnlineModList from 'components/views/OnlineModList.vue';
 
 let fs: FsProvider;
 
 
 @Component({
-    components: {ModalCard}
+    components: { OnlineModList, ModalCard}
 })
 export default class ImportProfileModal extends mixins(ProfilesMixin) {
     private importUpdateSelection: "IMPORT" | "UPDATE" | null = null;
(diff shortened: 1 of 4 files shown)


src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -120,25 +134,26 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
-        }
-
         let read: string | null = await this.readProfileFile(files[0]);
 
         if (read !== null) {
-            const parsed: ExportFormat = await this.parseYamlToExportFormat(read);
-
-            const localListenerId = this.listenerId + 1;
-            this.listenerId = localListenerId;
-            document.addEventListener('created-profile', ((event: CustomEvent) =>
```
anttimaki, line 134: Moving this stuff to `installProfileHandler` makes the comment above this function outdated.

anttimaki (commented):
Overall the code changes look fine. Couple of possible issues:

- Any local mods in the imported profile are not shown on the preview
- https://github.com/ebkr/r2modmanPlus/blob/develop/src/r2mm/downloading/BetterThunderstoreDownloader.ts#L209 has this checkup for importing profiles that contains no mods found in Thunderstore's API. Usually this is due to importing the profile to a wrong game, but could be due to mod takedowns as well. Currently the preview is empty and if user proceeds the error is shown as before. It would be better UX to handle the "empty profile" case already on the preview phase and not let the user proceed.

I think these are somewhat niche cases so IDK if they should be considered blockers. The feature itself is probably valuable for most of the users as it is now.

ebkr:
> Any local mods in the imported profile are not shown on the preview

Locally imported mods aren't included anyway. We could potentially indicate which ones but I don't think it's a concern and is honestly probably better to inform the original exporter rather than the person importing the profile.

> https://github.com/ebkr/r2modmanPlus/blob/develop/src/r2mm/downloading/BetterThunderstoreDownloader.ts#L209 has this checkup for importing profiles that contains no mods found in Thunderstore's API

That's a good spot that I didn't consider - I'll add a new path to say that there are no resolvable mods to import for that code.

ebkr:
If running locally, this can be tested using: `0191d789-b860-5ed9-8a64-f3778f0f760e`

anttimaki:
> Locally imported mods aren't included anyway. We could potentially indicate which ones but I don't think it's a concern and is honestly probably better to inform the original exporter rather than the person importing the profile.

Okay so what tripped me up on this is that the locally imported mods are included in the `mods.yml` and shown on local mod list, although the actual files for the mod aren't installed. Since locally importing the mod added it into the cache, the icon even worked for me on the mod list. But that's probably a niche issue and not related to changes in this PR, so not a blocker IMO.

> That's a good spot that I didn't consider - I'll add a new path to say that there are no resolvable mods to import for that code.

Yeah, this is now a lot nicer experience for the user :+1:

anttimaki (approved):

anttimaki:
Hope I'm not jumping the gun when I merge this to get the changes to TSMM release.

PR #1437, opened 2024-09-11
WIP: mod version hiding

anttimaki:
Some of the communities share a common BepInEx package. If a newer
version is needed for one community, it's risky to just publish the
new version, since it would trigger the auto update feature for all
these communities, possibly breaking something. To prevent this, the
Thunderstore ecosystem schema will eventually include an optional
field for defining which version a community should use.


anttimaki:
Closing the PoC branch due to lack of traction.

PR #1438, opened 2024-09-13
Fix ThunderstoreMod date field typing

anttimaki:
The date fields denoting creation and last update date are in fact strings, not dates. IDK if it has always been this way or if e.g. moving the online mod list from disk to IndexedDB changed this.

Luckily the date fields are only used for newer/older than comparisons, and the ISO 8601 date strings yield the same results for those operations as the Date objects would.


VilppeRiskidev (approved):
Nice 👌

PR #1439, opened 2024-09-13
Fix watching online mod list for changes

anttimaki:
Online mod list was watching a wrong variable, and wouldn't fire an autoupdate if the main mod list changed, i.e. if there was a background update from Thunderstore API while user had the online mod list open.

This was broken some 6 months ago by yours truly while moving the mod list into Vuex store.


VilppeRiskidev (approved):
Cool!

PR #1447, opened 2024-09-17
Disable MacOS from CI pipeline

anttimaki:
GitHub actions dropped the support for MacOS 11, causing all CI jobs to fail. Updating to MacOS 12 would require updating 3rd party dependencies, namely electron-builder (see PR #1409).

MacOS isn't officially supported by this project, and it was added to CI pipeline just because doing so was easy and only required a few lines of config code. Now keeping MacOS on the CI pipeline would require testing that updating electron-builder has no ill side-effects. That work seems to have a low return of investment, so changes are it won't get done anytime soon.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,9 +16,10 @@ jobs:
           - kind: windows
             os: windows-latest
             platform: win
-          - kind: mac
-            os: macos-11
-            platform: osx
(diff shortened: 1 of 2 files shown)


anttimaki:
Merging this without review. It's easy enough to undo if it turns out to be a bad idea.

PR #1448, opened 2024-09-17
Time bubble

anttimaki:
- Proof of concept for filtering the list of available mods based on dates
- There's still no proper specification for how we want the feature to actually work and look like


anttimaki:
Closing the PoC branch due to lack of traction.

PR #1450, opened 2024-09-19
Separate Zip extraction function from Profile Import Ui component

VilppeRiskidev:


src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -26,6 +26,7 @@ import ThunderstoreVersion from "../../model/ThunderstoreVersion";
 import ManagerInformation from "../../_managerinf/ManagerInformation";
 import OnlineModList from "../views/OnlineModList.vue";
 import * as PackageDb from '../../r2mm/manager/PackageDexieStore';
+import { extractZippedProfileFile } from "../../utils/ProfileUtils";
```
anttimaki, line 29: This is fine now but we'll likely end up importing all the stuff from ProfileUtils, so this would be cleaner as `import * as ProfileUtils`. This can be changed in the next PR.

src/utils/ProfileUtils.ts
```diff
@@ -0,0 +1,29 @@
+import ZipProvider from "../providers/generic/zip/ZipProvider";
```
anttimaki, line 1: The imports in this project generally is a mess as there's no linting to autosort them. Regardless I personally like to at least try to get them right when creating new files. By "right" I mean "what I'm used to", which is the Python way, i.e. "standard library imports" (e.g. path) separated by a blank line from 3rd party imports (e.g. yaml) separated by a blank line from local imports. Each section should be sorted alphabetically based on the import path, e.g. the part after "from".

That's my personal preference, so this won't be a blocker now or later. If you choose, this can be fixed in the next PR.

anttimaki (approved):
Approved with comments that can be taken into account in the next PR.

PR #1451, opened 2024-09-20
Separate parseYamlToExportFormat function from ImportProfileModal UI …

VilppeRiskidev:
…component

### DIFF
--- a/src/components/profiles-modals/ImportProfileModal.vue
+++ b/src/components/profiles-modals/ImportProfileModal.vue
@@ -1,32 +1,32 @@
 <script lang="ts">
-import { Component } from 'vue-property-decorator';
-import { ModalCard } from "../all";
-import R2Error from "../../model/errors/R2Error";
-import { ProfileImportExport } from "../../r2mm/mods/ProfileImportExport";
-import ExportMod from "../../model/exports/ExportMod";
 import path from "path";
-import Profile from "../../model/Profile";
-import FileUtils from "../../utils/FileUtils";
-import FsProvider from "../../providers/generic/file/FsProvider";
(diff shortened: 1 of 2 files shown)


src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -1,32 +1,32 @@
 <script lang="ts">
+import path from "path";
```
anttimaki, line 2: I very much appreciate the thought of ordering imports here in the "old" file too! Ideally it would be done in a separate commit since now it's kinda hard to say what changes are actually due to moving the helper function what are just reordering. Also the ordering here seems to be "import X" instead of "from X"? Not a blocker.

src/utils/ProfileUtils.ts
```diff
@@ -1,6 +1,12 @@
-import ZipProvider from "../providers/generic/zip/ZipProvider";
 import path from "path";
+
+import * as yaml from "yaml";
+
 import Profile from "../model/Profile";
+import ExportFormat from "src/model/exports/ExportFormat";
```
anttimaki, line 6: **Blocker**: these imports will break builds on TSMM. Use relative `../` paths instead.

anttimaki (changes requested):
You fell victim to one of the classic blunders (automatic imports using absolute paths).

anttimaki (approved):
@VilppeRiskidev the PR is approved, but unlike we agreed earlier, I won't merge it yet. This is because Mythic wants to keep the "release pipeline clean" this week in case we need to do a hotfix update. So chain the next PR on this branch instead.

PR #1454, opened 2024-09-26
Separate installModAfterDownload function from UI component

VilppeRiskidev:
Separate installModAfterDownload function from UI component and use early return pattern consistently.


anttimaki (commented):
@VilppeRiskidev The changes in the PR look fine, but the part about throwing errors instead of returning them is missing. Is this an oversight? If it's intentional, I'd like to hear why you chose not to implement that part.

anttimaki (approved):
LGTM 👍🏻

PR #1459, opened 2024-10-01
ImportProfileModal improvements

VilppeRiskidev:


src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -100,20 +100,22 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
-            }],
-            buttonLabel: 'Import'
-        }).then((value: string[]) => {
-            if(value.length === 0) {
-                this.closeModal();
-            }
-            this.importProfileHandler(value);
-        })
+        this.activeStep = 'IMPORT_FILE';
+        setImmediate(() => {
```
anttimaki, line 112: - So based on some googling, this is a Node.js thing. I don't think it's available at least in TSMM (it *might* be)
- MDN says it's deprecated and shouldn't be used
- Why this works in TSMM *might* be due to a NPM shim package that's present via 3rd party dependency. If those packages drop the dependency, this *might* break
- For the uncertainty caused by the aforementioned bullet points, I think it's a safer bet to use `this.$nextTick`, since it's guaranteed to be present in Vue and not to stop working unexpectedly
- Although I am interested why you added this int the first place? It seems it wasn't needed before, so why is it needed now?

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -273,7 +280,6 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
             if (installResult instanceof R2Error) {
                 this.$store.commit('error/handleError', installResult);
                 keepIterating = false;
-                this.isProfileBeingImported = false;
```
anttimaki, line 274: Should we call `this.closeModal()` here? Or will that cause some issues?

anttimaki (changes requested):
- Overall looks nice
- With "better safe than sorry" sentiment, I think we shouldn't use the `setImmediate`
- Also note the separate question about closing the modal

src/components/profiles-modals/ImportProfileModal.vue
VilppeRiskidev: Weird vue behaviour, it doesn't want to rerender/update the modal when the file selector is opened, IDK if there's some better way to do this, but at least waiting for nextTick seemed to fix it. (The whole IMPORT_FILE modal wasn't being used before, so it didn't get caught not working)

VilppeRiskidev: actually `this.$nextTick` didn't seem to work either, but `process.nextTick` works

src/components/profiles-modals/ImportProfileModal.vue
VilppeRiskidev: yeah, why not 👍

anttimaki (approved):
Would've been cleaner if the fixes to the first and second commit would've been done to those commits, or a separate commit, instead of bundling them into the third commit. Doesn't change the outcome though.

Issue #1051, opened 2023-07-30
[FEATURE REQUEST] - Support Steam in a Flatpak


JoshuaFern:
**What problem does the feature request solve?**
Sometimes Linux gamers don't install Steam natively but instead opt for the Flatpak version of Steam.

**Example mock-up (optional)**
The path for Flatpak'd Steam would look something like `~/.var/app/com.valvesoftware.Steam/.steam/steam/steamapps/common/Valheim`. It could be implemented as another selectable option, i.e. Xbox, Steam, Steam (Flatpak)

**Additional information (optional)**
A workaround before this is implemented could be simply linking the Steam native directory to the Steam Flatpak directory. But that doesn't work for me because I have both a native and a Flatpak version of Steam installed.

Note, this is not a duplicate of #729 because this issue is relating to using r2modman as an AppImage with Flatpak'd Steam, whereas  #729 is relating to *distributing* r2modman as a Flatpak, potentially useful on the Steam Deck.

rehashedsalt:
This is required for compatibility with immutable OSes such as Fedora Silverblue/Kinoite

TheJanzap:
I'm running Flatpak Steam and installed r2modman via AUR and r2modman detected my Steam directory as `~/.var/app/com.valvesoftware.Steam/.local/share/Steam` and the game dir was also detected properly. However, upon launching the game, I get the error `You are missing the following 32-bit libraries, and Steam may not run:
libc.so.6`, so it is probably not running Steam correctly?

mscharley:
I managed to cobble together something that works from answers in a few places:

1. Use flatseal to grant Steam access to `xdg-config/r2modmanPlus-local` under "Other files".
1. Use the AppImage for r2modman normally to install and manage mods and profiles.
3. Use the help tab inside r2modman while on a profile and copy the launch parameters at the bottom of the page under "Launching the game from outside the mod manager" to get Steam to launch the mod profile.

The "Start modded" and "Start vanilla" buttons in r2modman don't appear to be working as they seem to be trying to run steam outside the flatpak sandbox, but this appears to be the only concession till there's proper support implemented. All the game and data folders all got autodetected correctly for me even though I have my games installed inside the flatpak data folder (ie, in `~/.var/app/com.valvesoftware.Steam`)

brug-dev:
> I managed to cobble together something that works from answers in a few places:
> 
[...]

PR #1462, opened 2024-10-02
Option to use Flatpaks under Linux

### LINKED ISSUE #1144: [BUG] - Flatpak steam detected but cannot "Start Modded"
**Describe the bug**
I am attempting to use r2modman with the flatpak version of steam on linux. I can install mods but when I click "Start Modded" nothing happens.

**To Reproduce**
1. Go to r2modman
2. Click on "Start Modded" with a game
3. Observe the game not opening

**Expected behavior**
I expected the game to open from steam

**Additional Context**
I realize this might not be able to be resolved on the r2modman side, but providing instructions to get it working with flatpak steam would be helpful. It seems this might be because of the steam flatpak's sandbox. Perhaps instructions could be added somewhere on github?

RossBrunton:
This adds a configuration setting "linuxUseFlatpak" which controls whether Steam should be launched as a Flatpak or not. By default this is detected automatically based on the path, but an option in the settings can override the behaviour.

If this option is enabled, we first check if we are able to read from our config directory. If not, we display an error and ask the user to give us permission. If we do, then we launch Steam through flatpak rather than the binary directly.

Closes: #1144, #1051

### DIFF
--- a/src/components/settings-components/SettingsView.vue
+++ b/src/components/settings-components/SettingsView.vue
@@ -299,6 +299,23 @@ let settingsList = [
         'fa-file-alt',
         () => emitInvoke('ShowDependencyStrings')
     ),
+    new SettingsRow(
+        'Other',
+        'Launch as Flatpak (Linux)',
+        'Launch Steam via flatpak',
+        async () => {
+            switch (settings.value.getContext().global.linuxUseFlatpak) {
+                case null:
+                    return 'Automatic';
+                case true:
+                    return 'Flatpak';
+                case false:
+                    return 'Native';
+            }
+        },
+        'fa-exchange-alt',
+        () => emitInvoke('ToggleLinuxUseFlatpak')
+    ),
 ];
 
 watch(search, () => {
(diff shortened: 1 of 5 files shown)


RossBrunton:
So I did all this work, and then decided to check if someone else had done it, and it seems both @wash2 ( https://github.com/ebkr/r2modmanPlus/pull/798 ) and @PedroVH ( https://github.com/ebkr/r2modmanPlus/pull/1064 ) have done this earlier. Guess I should have checked first.

Anyway, I think hopefully this should be more sophisticated, since it doesn't assume any hardcoded paths, allows the user to configure things and shows up an error if it isn't configured correctly.

Also, full disclosure, I've only tested this on Lethal Company, and I'm super unfamiliar with typescript.

RossBrunton:
Is there anything blocking this from being accepted?

gegoxaren:
Why has this not been merged yet?

ebkr:
Apologies for the huge delay on this.

The reason it took so long was two-fold.
1. I didn't want to support Flatpak Steam without also having a Flatpak build of the manager. It's pretty intuitive that both go together. The next release will make that properly available.
2. I don't have a Flatpak Steam installation to test this easily and it sucks having to ruin my current installation just to do so. I'll obviously take the hit at some point but this will still be further delayed for a little bit as I don't have a huge amount of storage space to dual boot.

PR #1464, opened 2024-10-03
Remove remnants of the old r2x format

VilppeRiskidev:


VilppeRiskidev:
WIP: I have to figure out how the exporting works, then remove .r2x stuff from exporting.

PR #1481, opened 2024-10-11
Unify folder term

VilppeRiskidev:


anttimaki (approved):
Well that's a lot more changes than I anticipated but I guess the logic for choosing folder over directory is still valid. The commit message could have contained a bit more information about why this change was done, but that's not critical.

I'll hold the merge for now due to TSMM being in the middle of rollout process.

PR #1482, opened 2024-10-11
Decouple download provider from full list of ThunderstoreMod objects

anttimaki:


src/r2mm/manager/PackageDexieStore.ts
```diff
@@ -123,6 +126,40 @@ export async function getPackagesByNames(community: string, packageNames: string
...
 }
 
+/**
+ * @param community Community which package listings should be used in the lookup.
+ * @param dependencies Lookup targets as Thunderstore dependency strings.
+ * @param useLatestVersion Ignore the version number in dependencyString and return the latest known version.
+ * @returns ThunderstoreCombo[], silently omitting unknown packages and versions.
+ */
+export async function getCombosByDependencyStrings(
+    community: string,
```
ebkr, line 136: It would be better for this to take a `Game` type and for this method to handle how to extract that. I don't think that the download provider should need to know that an internal folder name is what drives the community identifier.

src/r2mm/manager/PackageDexieStore.ts
```diff
@@ -123,6 +126,40 @@ export async function getPackagesByNames(community: string, packageNames: string
...
+    const split = dependencyStrings.map(splitToNameAndVersion);
+    const keys = split.map((d): [string, string] => [community, d[0]]);
+
+    // Dexie's anyOfIgnoreCase doesn't support compound indexes.
+    const packages = await db.packages.where('[community+full_name]').anyOf(keys).toArray();
+    const versionMap = Object.fromEntries(split);
+
+    return packages.map((rawPackage) => {
+        const rawVersion = useLatestVersion
+            ? rawPackage.versions[0]
```
ebkr, line 149: Is version 0 guaranteed to be the latest? Do we sort this on insertion?

I'm aware that we _can_ use the first version and it's unlikely to change, but I don't think it's good practice to rely on this behaviour unless we're accounting for it beforehand. It might be better to either sort, or reduce, to get the latest version instead.

src/r2mm/manager/PackageDexieStore.ts
```diff
@@ -123,6 +126,40 @@ export async function getPackagesByNames(community: string, packageNames: string
...
+
+        if (!rawVersion) {
+            return undefined;
+        }
+
+        const combo = new ThunderstoreCombo();
+        combo.setMod(ThunderstoreMod.parseFromThunderstoreData(rawPackage));
+        combo.setVersion(ThunderstoreVersion.parseFromThunderstoreData(rawVersion));
+        return combo;
+    }).filter((c): c is ThunderstoreCombo => c !== undefined);
```
ebkr, line 160: Two points on this:
1. Not a huge fan of this syntax. Do we need to use `(c): c is ThunderstoreCombo` instead of `(c: ThunderstoreCombo?) => c !== undefined`
2. Do we need to cast it? `c => c !== undefined)` is arguably more concise and gets the point across clearer. There may be an ESLint rule blocking this, so I'm happy for that to be the answer.

If neither of these strike you too much, I'm also not opposed to a util method to simply change this to something like: `.filter(ObjectUtils.onlyDefinedValues)`

src/utils/DependencyUtils.ts
```diff
@@ -0,0 +1,9 @@
+export const splitToNameAndVersion = (dependencyString: string): [string, string] => {
+    const parts = dependencyString.split('-');
+
+    if (parts.length !== 3) {
+        throw new Error(`Invalid dependency string "${dependencyString}"`);
```
ebkr, line 5: Careful, this is invalid. Author names can have hyphens iirc.

You can assume the dependency string **IS** valid because we're supplied it by TS. If a user attempts to import a local mod and gets it wrong then we just won't resolve it.

You can instead do (pseudo):
```
splits = dependencyString.split('-');
splits.last() => version
splits.secondLast() => mod name
splits.rest().join("-") => author name
```

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -14,66 +13,36 @@ import { ImmutableProfile } from '../../model/Profile';
...
-                    foundDependencies.push(tsCombo);
-                }
-            }
-        })
+    public async buildDependencySetUsingLatest(mod: ThunderstoreVersion, builder: ThunderstoreCombo[]): Promise<void> {
+        const useLatestVersion = true;
+        await this._buildDependencySet(mod, builder, useLatestVersion);
+    }
+
+    private async _buildDependencySet(mod: ThunderstoreVersion, builder: ThunderstoreCombo[], useLatestVersion=false): Promise<void> {
```
ebkr, line 30: This is more of a preference thing but it would be nicer to not have this defaulted. It could be more readable to create an Enum and we can utilise `DependencySetBuilderStyle.USE_LATEST` as a more readable equivalent wherever needed.

Default behaviour is weird IMO and we'll just prevent issues by being explicit.

ebkr (changes requested):

src/r2mm/manager/PackageDexieStore.ts
anttimaki: Makes sense.

For the background, when I whipped up the Dexie schema, I would've preferred to use the same community slug as is used in Thunderstore, but it wasn't available (aside from stripping from the API URL). I didn't want to add a new field to the GameManager mega list which already contains various formats of the spelling the game's name, so I just picked one that seemed most suitable. So although the variable is named "internalFolderName", what I use it for is "community identifier".

src/r2mm/manager/PackageDexieStore.ts
anttimaki: The current implementation sure does depend on zero index containing the latest version. So no new sorting or confirmations were added as part of this PR chain. It could be done as a separate task, but IDK how much overhead that would add to Lethal Community. Might be negligible 🤷🏻

src/r2mm/manager/PackageDexieStore.ts
anttimaki: It's not about ESLint. Does the repo even use ESLint? 

Anyway, semantically speaking, I think the user-defined type guard is considered "type narrowing", and not brute force "type casting". AFAIK using "x is y" type guard functions is perfectly fine and TypeScriptic way of doing things like this.

In this particular case it's optional (due to simplicity of the case), and can be left out. At the same time the code base uses things like `private packageUrl: string = '';` where the `string` part is fully optional and not needed by TypeScript. But someone has considered it clearer to include these extra type descriptions only for the readers benefit. I don't think using the type guard here is any different. And when discussion is based on personal preferences, the proper solution would be to introduce linting to automate forcing the conventions instead of discussing them in code reviews. But that's outside the scope.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
anttimaki: Generally speaking I don't consider parameter default values a negative thing. Of course one can go overboard with them, and e.g. using mutable default values is a trap. Constants and enums also have their use, especially to avoid typos on magic strings. I'm not sure how well that applies to booleans though.

In this particular case, we're talking about a private method, which is only called by the two named wrapper methods above. So I'd argue that having the named methods is more readable than having _buildDependencySet use enum parameter. (Also the named methods are public although they're only used internally, so to get rid of them would require changing the interface. Which might not be a bad thing in itself, but is a bit of scope creep.)

src/r2mm/manager/PackageDexieStore.ts
ebkr: Yeah I'm happy for it to be the community identifier, it's mostly just to separate functional concerns

src/r2mm/manager/PackageDexieStore.ts
ebkr: You can keep it as-is unless you want to do the change. No requirement to sort if we didn't before.

src/r2mm/manager/PackageDexieStore.ts
ebkr: > It's not about ESLint. Does the repo even use ESLint?

Potentially not. It did at one point and I think my configuration still has remnants of it.

> Anyway, semantically speaking, I think the user-defined type guard is considered "type narrowing", and not brute force "type casting". AFAIK using "x is y" type guard functions is perfectly fine and TypeScriptic way of doing things like this.

I'm fine with type guards where we actually need them, I don't think we've really used them much on the project and they don't appear to really add any benefit here as we don't care about the type of the filter. Supposedly we can do
```ts
.filter<ThunderstoreCombo>(c => c !== undefined)
``` 
which might read a bit nicer?

> At the same time the code base uses things like private packageUrl: string = ''; where the string part is fully optional and not needed by TypeScript

That looks like remnants of ESLint, granted I do prefer type declarations on properties themselves.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
ebkr: The problem with default values is that it's extremely easy to miss them out. I'd much rather we were explicit with our signatures. 

A boolean is IMO still a magic variable.
If you were to see the following on its own with no other context:
```ts
_buildDependencySet(mod, dependencies, true)
```
The `true` value doesn't really mean anything.

Equally, if you were to see:
```ts
_buildDependencySet(mod, dependencies, true)
...
_buildDependencySet(mod, dependencies)
...
_buildDependencySet(mod, dependencies, false)
```
You don't really know what's being done differently without context of the parameter's name

src/r2mm/manager/PackageDexieStore.ts
anttimaki: Changed to accept [Game object instead of id string](https://github.com/ebkr/r2modmanPlus/pull/1482/commits/9d71ae7c907b9b0ca24911a070847b4cc03b4974#diff-2122463b05e5c9342979622ca58d48970d706008d1d9df423063e69e07055988R137).

src/r2mm/manager/PackageDexieStore.ts
anttimaki: Did not change.

src/r2mm/manager/PackageDexieStore.ts
anttimaki: > `.filter<ThunderstoreCombo>(c => c !== undefined)`

Fails for me on dev server build: `TS2345: Argument of type '(c: ThunderstoreCombo | undefined) => boolean' is not assignable to parameter of type '(value: ThunderstoreCombo | undefined, index: number, array: (ThunderstoreCombo | undefined)[]) => value is ThunderstoreCombo'.`

Did not change.

src/utils/DependencyUtils.ts
anttimaki: Changed the implementation to [accept more than two hyphens](https://github.com/ebkr/r2modmanPlus/pull/1482/commits/9d71ae7c907b9b0ca24911a070847b4cc03b4974#diff-485ac28d8d642a9d12bce5c6751707214cc35527c52eac5c156d29640449b882R1).

ebkr (approved):

PR #1483, opened 2024-10-11
Don't require complete package list to install mods via URL protocol

anttimaki:
Check the existence of the target package and version directly from the IndexedDB cache. After the change the Manager component no longer needs access to the complete package list of the selected community.

### DIFF
--- a/src/pages/Manager.vue
+++ b/src/pages/Manager.vue
@@ -141,7 +141,6 @@ import Vue from 'vue';
 import Component from 'vue-class-component';
 import { Hero, Link, Modal, Progress } from '../components/all';
 
-import ThunderstoreMod from '../model/ThunderstoreMod';
 import ThunderstoreCombo from '../model/ThunderstoreCombo';
 import ProfileModList from '../r2mm/mods/ProfileModList';
 import PathResolver from '../r2mm/manager/PathResolver';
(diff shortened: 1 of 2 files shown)


src/pages/Manager.vue
```diff
@@ -619,8 +614,9 @@ import ModalCard from '../components/ModalCard.vue';
 			this.launchParametersModel = this.settings.getContext().gameSpecific.launchParameters;
 			const ignoreCache = this.settings.getContext().global.ignoreCache;
 
-			InteractionProvider.instance.hookModInstallProtocol(async data => {
-                const combo: ThunderstoreCombo | R2Error = ThunderstoreCombo.fromProtocol(data, this.thunderstoreModList);
+            InteractionProvider.instance.hookModInstallProtocol(async (protocolUrl) => {
+                const community = this.$store.state.activeGame.internalFolderName;
```
ebkr, line 618: We should be able to remove this as part of the comment on the previous PR

ebkr (approved):

src/pages/Manager.vue
anttimaki: If this is referring to using the Game object rather than the folder name string, I think I'd rather still pass the game object as parameter here. As a part of the recent refactorings, we've moved the state management out of various business logic and helper files in preparation for moving the business logic to be handled by the CLI. Although this particular case is somewhat in the grey area, I think it would be a step backwards to introduce more dependencies on GameManager where it can be easily avoided.

ebkr: Oh yeah, this was meant as we refactor to just pass the game through.

anttimaki: Refactored to [pass a Game object instead of a string](https://github.com/ebkr/r2modmanPlus/pull/1483/commits/0df0f5f9d8d695be42477e8dc412d0dcf2596495#diff-30efcca841db0d1e022ad42806e25a57c7b530c1844f6c8bf4d20581803b8b93R618-R619).

PR #1485, opened 2024-10-11
Don't depend on ThunderstoreMod containing all ThunderstoreVersions in components

anttimaki:


src/r2mm/manager/PackageDexieStore.ts
```diff
@@ -174,3 +184,24 @@ export async function isLatestPackageListIndex(community: string, hash: string)
 export async function setLatestPackageListIndex(community: string, hash: string) {
     await db.indexHashes.put({community, hash, date_updated: new Date()});
 }
+
+async function getDexiePackage(community: string, packageName: string) {
```
ebkr, line 188: Not a huge fan of this naming. It makes it sound Dexie related rather than data related

src/r2mm/manager/PackageDexieStore.ts
```diff
@@ -174,3 +184,24 @@ export async function isLatestPackageListIndex(community: string, hash: string)
...
+    const pkg = await db.packages.where({community, full_name: packageName}).first();
+
+    if (!pkg) {
+        throw new Error(`Couldn't find package "${packageName}" in ${community} packages`);
+    }
+
+    return pkg;
+}
+
+async function getDexieVersion(community: string, packageName: string, versionNumber: string) {
```
ebkr, line 198: Similar for here except the problem is worse. This makes it sound like you'd be getting the currently installed version of the Dexie library.

ebkr (commented):
I'm not sure I'm following on this PR, or at least the title isn't matching the content of the PR.

How have we removed ThunderstoreVersions from ThunderstoreMod?
`getDexiePackage` is returning the ThunderstoreMod with all the versions, and `getDexiePackage` gets a specific package. We are still however requiring every package to be contained to determine all selectable version numbers.

This was mainly triggered by the line:
```
this.selectedVersion = this.thunderstoreMod.getLatestVersion().getVersionNumber().toString();
```
which implies that every version is still there.

We are then doing:
```
this.versionNumbers = await PackageDb.getPackageVersionNumbers(
  this.activeGame.internalFolderName,
  this.thunderstoreMod.getFullName()
);
```
which currently feels like a redundant call.

Am I missing something?

src/r2mm/manager/PackageDexieStore.ts
anttimaki: Note that `getDexiePackage` and `getDexieVersion` aren't exported, so I'd consider them internal implementation details of the module. The naming refers to the (also internal) interfaces of the same name. That being said, I'm open for suggestion for better names for both the interfaces and the functions.

anttimaki:
> I'm not sure I'm following on this PR, or at least the title isn't matching the content of the PR.
> 
> How have we removed ThunderstoreVersions from ThunderstoreMod? ...

Background: when I started this PR stack, I wasn't sure if we could remove the `ThunderstoreVersions` completely from the `ThunderstoreMod`, or if it would be more practical to keep a reference to the latest one. So the "all" in the PR title refers to the fact that after these changes, the list of all versions is not directly accessed from the outside, but the reference to the latest version is still used. E.g. changes were made where `this.thunderstoreMod.getVersions()[0]` was substituted with `this.thunderstoreMod.getLatestVersion()`. So while `getLatestVersion` still internally depends on the whole list being there, none of the external call sites do.

The actual removal is done in the next, last PR. On hindsight this would probably been clearer if I didn't split these last two PRs into separate ones. Or worded the title differently

ebkr (approved):

src/r2mm/manager/PackageDexieStore.ts
anttimaki: Didn't change the naming as no alternatives were provided.

ebkr: getPackage? getPackageFromDatabase?

anttimaki: Renamed [the function](https://github.com/ebkr/r2modmanPlus/pull/1485/files#diff-2122463b05e5c9342979622ca58d48970d706008d1d9df423063e69e07055988R190).

src/r2mm/manager/PackageDexieStore.ts
anttimaki: Renamed [the function](https://github.com/ebkr/r2modmanPlus/pull/1485/files#diff-2122463b05e5c9342979622ca58d48970d706008d1d9df423063e69e07055988R200).

PR #1486, opened 2024-10-11
Don't depend on ThunderstoreMod containing all ThunderstoreVersions in Vuex

anttimaki:


src/store/modules/ProfileModule.ts
```diff
@@ -282,6 +289,16 @@ export default {
             await dispatch('resolveConflicts', params);
         },
 
+        // Return ThunderstoreCombos pointing to the latest available version.
+        async getModsWithUpdates({getters, rootState}): Promise<ThunderstoreCombo[]> {
```
ebkr, line 293: We should differentiate these names a bit

`modsWithUpdates` returns `ThunderstoreMod[]`
`getModsWithUpdates` returns `ThunderstoreCombo[]`

ebkr (approved):
Approved pending rename

src/store/modules/ProfileModule.ts
anttimaki: Well both `ThunderstoreMod` and `ThunderstoreCombo` present a "mod" so in that sense it's not incorrect. Nor does it differ from the rest of the code base where any variable named `mods` or `modList` may contain `ThunderstoreMod`s, `ThunderstoreCombo`s, `ManifestV2`s, or `ExportMods`s.

Generally speaking I'd appreciate suggestions when a naming change is requested. As the types seemed the be the issue here, I changed `getModsWithUpdates` to `getCombosWithUpdates`. I'm not too fond of this sort of Hungarian notation when using a typed language, but that argument is diminished by the fact that the version of Vuex used mangles the type info anyway.

PR #1493, opened 2024-10-15
Limit profile name length to prevent file system level errors

VilppeRiskidev:


VilppeRiskidev:
Added a maxlength to profile name input. I can't seem to get it to fail silently anymore. The error message if/when it _does_ fail is still not very user friendly, but I didn't figure out a way to reliably tell what caused the error.

anttimaki:
@VilppeRiskidev closing this now due to path lengths being handled via changes done in other PRs.

PR #1496, opened 2024-10-16
Improve error handling when importing profiles via code

VilppeRiskidev:


src/components/mixins/ProfilesMixin.vue
```diff
@@ -33,5 +33,10 @@ export default class ProfilesMixin extends Vue {
         return sanitize(nameToSanitize);
     }
 
+    isProfileCodeValid(profileImportCode: string): boolean {
+        const REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
```
anttimaki, line 37: Defining the regexp inside the function means an object gets created each time the function is called. Given that it's relatively simple case and probably won't be called that many times(*), the overhead caused by this might be negligible. However, it would be better to define it on the "root" level of the module, i.e. before this class definition.

(*): it's hard to estimate how many times Vue calls this though. Might be a lot?

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -267,13 +267,14 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
             <span class="tag is-dark" v-if="profileImportCode === ''">You haven't entered a code</span>
+            <span class="tag is-danger" v-else-if="!isProfileCodeValid(profileImportCode)">Invalid code, check for typos</span>
             <span class="tag is-success" v-else>You may import the profile</span>
         </template>
         <template v-slot:footer>
             <button
                 id="modal-import-profile-from-code-invalid"
                 class="button is-danger"
-                v-if="profileImportCode === ''">
+                v-if="profileImportCode === '' || !isProfileCodeValid(profileImportCode)">
```
anttimaki, line 277: Nitpick: IMO this would be clearer without the `profileImportCode === '' ||` prefix. I guess one could argue that the perhaps unnecessary function call adds overhead as well, but I would prioritize the clean code in this case.

src/r2mm/profiles/ProfilesClient.ts
```diff
@@ -19,6 +19,19 @@ function formatApiError<T>(e: T, genericTitle: string): R2Error | T {
                 message,
                 "You were rate limited by the server, wait for a while and try again."
             );
+        } else if (e.response.status === 404) {
+            // Note that this error text is only applicable when importing, but it should be OK as creating a profile shouldn't respond with 404 anyway.
+            return new R2Error(
+                genericTitle,
+                `Server responded with \"404: Not Found\" while trying to download the profile`,
```
anttimaki, line 26: Nitpick: template literals (using backticks when defining a string) are usually used only when embedded expression (`${myVariable}`) is used. Often a linting tool enforces such behaviour. But linting is not currently present, so I don't consider that an enforceable rule. The code base already mixes `"` and `'`, so I wouldn't want to add a third option into the mix, though.

(Also if you end up using the backticks anyway, you don't need to escape quotation marks inside them.)

src/r2mm/profiles/ProfilesClient.ts
```diff
@@ -19,6 +19,19 @@ function formatApiError<T>(e: T, genericTitle: string): R2Error | T {
...
+                `The profile import code entered might either be expired or contain typos.`
+            );
+        } else if (e.response.status.toPrecision(1) === "5e+2") { // Status code is 5xx
```
anttimaki, line 29: - CDN issues do not return 500 error code, or any error code for that matter, since the destination server is not reached at all. They manifest as `Network Error`s
- These are moot points due to the above, but I wrote them before the first one so FYI:
    - While what's done hear is neat (and a TIL moment for me), I think the code isn't that readable without the comment accompanying it. I would prefer `(e.response.status > 499 && e.response.status < 600)` for readability's sake.
    - The error message says "while trying to download", but I assume uploads might return 500 errors as well

anttimaki (commented):
Mostly nitpicking but the fact that CDN error doesn't return a 500 series of response code a major one. I guess this is hard for you to test naturally if you don't encounter the problem. I think you can fake the problem by turning the connection to offline in dev tool's network tab - at least this results in a similar error for me.

VilppeRiskidev:
Previous comments should be now fixed and the PR is completely ready for review.

src/r2mm/mods/ProfileImportExport.ts
```diff
@@ -49,8 +52,33 @@ async function saveDownloadedProfile(profileData: string): Promise<string> {
...
-    const response = await ProfileApiClient.getProfile(profileCode);
-    return saveDownloadedProfile(response.data);
+    try {
+        let is404: boolean = false;
+        const response: AxiosResponse<string> = await retry(
+            () => ProfileApiClient.getProfile(profileCode),
+            5,
+            1000,
+            () => { return !is404 },
+            (e: any) => {
```
anttimaki, line 62: The signature for this argument is `(e: Error|unknown)`, not `(e: any)`. Using `any` here masks hides a bug: `e.message` might be undefined, so calling `e.message.startsWith()` causes another error.

I think best way in this situation would be to do something like:
```
(e: unknown) => {
    console.error(e);
    if (R2Error.fromThrownValue(e).message.startsWith("404")) {
        is404 = true;
    }
}
```

src/r2mm/mods/ProfileImportExport.ts
```diff
@@ -49,8 +52,33 @@ async function saveDownloadedProfile(profileData: string): Promise<string> {
...
+                console.error(e);
+                if (e.message.startsWith("404")) {
+                    is404 = true;
+                }
+            },
+            true
+        );
+        return saveDownloadedProfile(response.data);
+    } catch (e: any) {
```
anttimaki, line 71: You might want to consider getting rid of `any` here too. While the current implementation doesn't contain bugs, it increases a change that someone editing this creates one. If you expect the `e` be a regular `Error` here, something like this could suffice:

```
} catch (e: unkown) {
    if (e instance of Error && e.message === "Network Error") {
        ...
```

anttimaki (approved):
Detected two minor issues still but will fix them in another PR to get this merged and released.

PR #1503, opened 2024-10-22
Remove last uses of Profile from ProfileInstallerProvider

anttimaki:
ImmutableProfile objects will now be used with uninstallation operations, as changing the target profile in the middle of the operation would result in disaster.


VilppeRiskidev (approved):
Yes 👍

PR #1514, opened 2024-11-01
GDWeave support

### LINKED ISSUE #1512: [FEATURE REQUEST] - GDWeave support
[GDWeave](https://github.com/NotNite/GDWeave) is a mod loader for Godot games. It supports Godot 3.5.2 for the game [WEBFISHING](https://store.steampowered.com/app/3146520/WEBFISHING/). More support for other Godot versions may be added in the future. (It was developed for the WEBFISHING community but has the potential to branch out to other games, essentially.)

Disclosure: I am the developer of GDWeave.

## File format info

```
game.exe
winmm.dll                           # Rust loader
GDWeave/
  GDWeave.log                       # does not contain game logs, only loader logs
  core/
    GDWeave.dll                     # C# core
    <other dependencies of GDWeave>
  mods/
    <mod id>/                       # folder name is not enforced but encouraged
      manifest.json                 # not a Thunderstore manifest
      <other mod files>
  configs/
    <mod id>.json                   # unstructured JSON object
```

`winmm.dll` must be present in the game directory (like BepInEx using Unity Doorstop). GDWeave will expect the GDWeave folder to exist next to the running executable, this will break profiles. I am willing to add a `GDWEAVE_FOLDER_OVERRIDE` environment variable or such.

The GDWeave mod manifest is not important to the mod manager - it contains an ID and list of dependencies to load after.

## Technical info

GDWeave is written in C# and contains a Rust DLL for serving as an entrypoint. Users are required to install .NET 8 to use GDWeave. GDWeave will automatically prompt the user to install .NET 8 upon launch if it is missing, so r2modman does not need to handle this for the user.

GDWeave uses Reloaded hooking libraries and may set off Thunderstore anti-spam filters.

## State on Thunderstore

GDWeave mods live on Thunderstore ([WEBFISHING](https://thunderstore.io/c/webfishing/)), but GDWeave itself is not on Thunderstore yet. I need to get GDWeave uploaded as a package, then other mods must depend on GDWeave.

GDWeave will likely need to live in "cross-community limbo" like r2modman itself and BepInExPack. I will need to talk to Thunderstore moderators more about this.

The only community currently using GDWeave on Thunderstore is [WEBFISHING](https://thunderstore.io/c/webfishing/).

## Current mod ecosystem

We are still in progress of migrating mods to Thunderstore from [a custom mod index](https://github.com/NotNite/webfishing-mods). The community currently uses [Hook, Line, & Sinker](https://hooklinesinker.lol/) for mod management (which will/does support the Thunderstore API).

NotNite:
Closes #1512.

- Formatting is all over the place because the formatting was all over the place when I got here, so I just disabled format-on-save in VSCode and pretended I was Prettier.
- I haven't tested this super well, but it does generate a GDWeave log file. Mods seem to be loading fine.
- My installer logic is probably wrong. The entire `GDWeave` folder seems to be copied into the game install from the profile, so I think something is definitely wrong there.
- I need to make a GDWeave 2.0.13 release with the changes required for this (new CLI args). If you do not see this update, this PR won't work at all!

Sidenote: I had the most amusing dev env setup of my life. You can use Volta + Python 2 that I already had installed + manually installing the VS2015 Build Tools and Windows 8.1 SDK. Man I hate node-gyp sometimes.


NotNite:
Things left to solve before this can be merged:

- It seems that r2mm copies the entire profile folder (even a dummy .txt I manually created for testing) when clicking "Start modded". This is unwanted behavior but I'm not sure how to fix it (pretty much nothing needs to be copied in the GDWeave folder).
- Unsure if profile sharing and whatnot works.
- Can't run the sync script because I don't have fetch with this node version. Welp.
- A lot of tests are failing but I don't know if it's my fault?

NotNite:
Fixed the copy issue (thanks to Ebkr for helping [in the Thunderstore Discord](https://canary.discord.com/channels/809128887366975518/1300648467256836116/1302713558320152588)), and excluded GDWeave from profile exports. This excludes *everything*, configs included, but it seems to match the behavior of the other profile exports.

This PR is now ready for review. Please scrutinize the installer code - while it works well for me, I don't entirely understand what I wrote very well, and I mainly was looking at BepInEx/Lovely's implementation.

Given that this is a new mod loader for a new game, I'm willing to buy it on Steam for anyone who will be reviewing (lol). You can test easily by installing GDWeave, and either a mod that modifies the main menu (e.g. TackleBox will add a "Mods" button) or by checking the GDWeave log file (GDWeave/GDWeave.log).

src/installers/GDWeaveInstaller.ts
```diff
@@ -0,0 +1,82 @@
...
+            const srcPath = path.join(args.packagePath, item);
+            const destPath = args.profile.joinToProfilePath(item);
+            if ((await FsProvider.instance.stat(srcPath)).isFile()) {
+                await FsProvider.instance.copyFile(srcPath, destPath);
+            } else {
+                await FsProvider.instance.copyFolder(srcPath, destPath);
+            }
+        }
+
+        await addToStateFile(args.mod, fileRelocations, args.profile);
```
anttimaki, line 32: I don't think is needed as we don't need to track any extra state to be able to uninstall this. Also I'm not 100% how GDWeave should react to being disabled, but on a quick glance seems state isn't used for that either?

src/installers/GDWeaveInstaller.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+async function searchForManifest(fs: FsProvider, dir: string): Promise<string | null> {
```
anttimaki, line 36: Nitpick: It could look cleaner if this took `const files` from below as the argument and handled it all. Now the impementation is sort of splitted between the PluginInstaller and this function. Also IMO passing the FsProvider singleton as an argument is sort of an antipattern.

src/installers/GDWeaveInstaller.ts
```diff
@@ -0,0 +1,82 @@
...
+    return null;
+}
+
+export class GDWeavePluginInstaller implements PackageInstaller {
+    async install(args: InstallArgs) {
+        const fs = FsProvider.instance;
+        const installDir = path.join("GDWeave", "mods", args.mod.getName());
+        const files = await FsProvider.instance.readdir(args.packagePath);
+
+        // Packaging is all over the place. Find a folder with a manifest.json
```
anttimaki, line 57: I'm not against this "sniffing out the mod folder" pattern as long as it's acknowledged that using it is a compromise which trades "no changes to existing packages now" to "having potential problems in the future". Providing "do it exactly like this" instructions may be less confusing for new mod authors? E.g. if someone splits their implementation to two sibling folders, only one of them will get copies. This might be unnecessary future proofing, I do that sometimes 🤷🏻

src/r2mm/installing/default_installation_rules/game_rules/InstallRules_GDWeave.ts
```diff
@@ -0,0 +1,32 @@
+import type { CoreRuleType } from '../../InstallationRules';
+import * as path from 'path';
+import { GAME_NAME } from '../../profile_installers/ModLoaderVariantRecord';
+import { RuleSubtype } from "../../InstallationRules";
+
+export function buildGDWeaveRules(gameName: GAME_NAME, extraRules?: RuleSubtype[]): CoreRuleType {
```
anttimaki, line 6: This whole definitions shouldn't be needed, as GDWeaveInstaller.ts provides custom installation logic for the mod loader (and mods). Note the ReturnOfModding example though: the PackageInstaller implementations can call IntallRules internally if it makes implementation easier. In GDWeave's case the process is so straightforward though, that it might not be needed.

anttimaki (commented):
Here are some preliminary comments. This is not a full review, I'll do that once the current issues are addressed. And feel free to ask if there's any confusion, I'm not the top expert or authority when it comes to mod installation support, so I might have gotten something wrong.

src/installers/GDWeaveInstaller.ts
```diff
@@ -0,0 +1,82 @@
+import { InstallArgs, PackageInstaller } from "./PackageInstaller";
+import path from "path";
+import FsProvider from "../providers/generic/file/FsProvider";
+import { addToStateFile } from "./InstallRuleInstaller";
+import FileUtils from "../utils/FileUtils";
+
+const basePackageFiles = ["manifest.json", "readme.md", "icon.png"];
+
+// Adapted from the BepInEx & Lovely installers
+export class GDWeaveInstaller implements PackageInstaller {
```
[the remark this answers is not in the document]
NotNite: I've started on this, but I'm unsure how to handle the uninstall. ReturnOfModding deletes everything that isn't a folder and isn't `mods.yml` - I'll just copy this behavior for now, but let me know if I should do it differently.

src/installers/GDWeaveInstaller.ts
NotNite: Yeah, I wasn't sure how to use the state file, but knowing that there's a better way to accomplish it I'll be moving towards it.

src/installers/GDWeaveInstaller.ts
NotNite: I do have a suggested format for users, which is `GDWeave/mods/ModIDGoesHere` (documented [here](https://github.com/NotNite/GDWeave/blob/main/MODS.md#publishing-to-thunderstore)). Admittedly it would be better if it was a standard name `mod` directory like discussed prior, but I'm not sure if it's feasible to break this with over 100 packages at this point.

NotNite:
I rewrote the installer to be more like the ReturnOfModding one. Because I can't flat delete the GDWeave folder without deleting all of the mods inside of it, I had to special case `winmm.dll` and `GDWeave/core`. A state file might be better here? Let me know what you think!

I have tested installing and uninstalling mods and it seems to work fine. Apologies about that!

src/installers/GDWeaveInstaller.ts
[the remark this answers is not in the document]
anttimaki: ReturnOfModding author/community had clear ideas how they wanted things to work. For GDWeave, it's better to keep things as simple as possible If there's no reason to do otherwise. My comment was a bit misleading in that sense, sorry.

Anyway, there's now this discrepancy between the install and uninstall methods: install copies everything, but uninstall only removes `winmm.dll` and `GDWeave/core`. If you're fairly certain these two will suffice for GDWeave for now, I'd suggest the installation also only copies these two and ignores everything else.

src/installers/GDWeaveInstaller.ts
```diff
@@ -0,0 +1,138 @@
...
+        const { profile } = args;
+
+        try {
+            // ReturnOfModding deletes all files in the root and ignores directories.
+            // We can't do that because GDWeave/core exists, but blanket deleting GDWeave would delete all mods in the process.
+            // This would probably be handled better by a state file, but /shrug.
+
+            const toDelete = [
+                profile.joinToProfilePath('winmm.dll'),
+                profile.joinToProfilePath('GDWeave/core'),
```
anttimaki, line 48: `joinToProfilePath('GDWeave', 'core')` would be slightly safer.

src/installers/GDWeaveInstaller.ts
```diff
@@ -0,0 +1,138 @@
...
+        if (!modFolder) {
+            throw new Error('Could not find mod folder');
+        }
+
+        const root = profile.joinToProfilePath(
+            'GDWeave',
+            'mods',
+            mod.getName()
+        );
+        await FileUtils.ensureDirectory(root);
```
anttimaki, line 111: I think `FsProvider.instance.copyFolder(modFolder, root);` might be sufficient to replace lines 111-119?

src/pages/Manager.vue
```diff
@@ -490,9 +490,12 @@ import ModalCard from '../components/ModalCard.vue';
                 case PackageLoader.MELON_LOADER:
                     logOutputPath = path.join(this.profile.getProfilePath(), "MelonLoader", "Latest.log");
                     break;
-				case PackageLoader.RETURN_OF_MODDING:
```
anttimaki, line 493: Nitpick: this file unfortunately mixes tabs and spaces and no one's dared to refactor it. Anyway the indentation looks a bit wonky now.

src/r2mm/launching/runners/linux/SteamGameRunner_Linux.ts
```diff
@@ -80,7 +83,7 @@ export default class SteamGameRunner_Linux extends GameRunnerProvider {
 
     }
 
-    private async ensureWineWillLoadBepInEx(game: Game): Promise<void | R2Error>{
+    private async ensureWineWillLoadBepInEx(game: Game, proxyDll: string): Promise<void | R2Error>{
```
anttimaki, line 86: After the additional parameter, the method name is now a bit misnomer. Would `ensureWineWillLoadDoorstop` be more suitable?

Also I've no way to test this works on Linux but I assume you've done so?

anttimaki (commented):
In addition to the comments below:

- Another PR was merged and this needs rebasing
- HLS should be added to mod exclusion file(s)
- Disabling mods doesn't currently do anything. For ReturnOfModding for example this is handled by the usage of InstallRuleInstaller internally (I think). But that feels like a legacy way of doing it. I would rather take a look what adding the support into PackageInstaller would require. I *think* it should be something I can implement rather quickly. If it's not quick, I'm not sure if it needs to be a blocker for this PR anyway, as long as the community is okay with it.

anttimaki:
> I rewrote the installer to be more like the ReturnOfModding one. Because I can't flat delete the GDWeave folder without deleting all of the mods inside of it, I had to special case `winmm.dll` and `GDWeave/core`. A state file might be better here? Let me know what you think!
> 
> I have tested installing and uninstalling mods and it seems to work fine. Apologies about that!

I think just copying/deleting `winmm.dll` and `GDWeave/core` is completely fine. AFAIK the state management file is used only by MelonLoader, and they're working on some changes that might allow us to do away with that. So I'd rather not add any new code that depends on the feature. (This is my understanding of the situation at least.)

No need for apologies, the process is undocumented and confusing due to the legacy and newer code. In fact I think you've gotten pretty close in just a short time 👍🏻

anttimaki:
Based on a quick check the enable/disable thing I mentioned seems a quick change indeed, so don't worry about that, I'll handle it.

src/r2mm/launching/runners/linux/SteamGameRunner_Linux.ts
NotNite: Doorstop is a Unity-specific thing, so I'm going with "ensureWineWillLoadDllOverride".

PR #1517, opened 2024-11-04
Fix sorting online mod list by download counts

anttimaki:
This had broken when ThunderstoreMod object dropped a reference to a list of ThunderstoreVersions to reduce memory usage. ThunderstoreMod defined a separate getter for package's total download count, but the sorting method kept using the getter inherited by ThunderstoreMod from ThunderstoreVersion, always returning 0. TypeScript didn't catch the error, since due to inheritance, it was a valid getter to call.

Rather than rename the sorting method to use the correct getter, drop the duplicate getter to ensure similar errors don't occur in the future. Both ThunderstoreMod and ThunderstoreVersion now have the same getter, while former just returns the combined download counts of all the versions.

### DIFF
--- a/src/components/views/OnlineModList.vue
+++ b/src/components/views/OnlineModList.vue
@@ -46,7 +46,7 @@
                 <DonateButton :mod="key"/>
             </template>
             <div class='card-footer-item non-selectable'>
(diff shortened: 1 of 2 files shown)


VilppeRiskidev (approved):
Nice! 👍

PR #1519, opened 2024-11-04
Upgrade dependencies. 

JonathanSteininger:
This was a pretty simple change in the dependencies to upgrade the electron version to past v24. This is to improve Linux support especially under Wayland and NVIDIA.

#### Things that need editing.

- Reverting the version number increase that I had made.
- Validating the build instructions (removing OSX instructions)

#### Testing

I have only tested in the real world for Linux. 

The tests provided passed the same amount before and after these changes on my machine.

#### Notes

Due to the dependencies upgrading, the build environments also need to be newer. (versions should be in the build instructions)

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -8,9 +8,6 @@
 #### Linux Build
 - rpm builder
 
-#### OSX
-- dmg-license (from npm or yarn)
-
 ### Setting up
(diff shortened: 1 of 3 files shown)


JonathanSteininger:
Removed the version number changes that I had made for the build.

PR #1523, opened 2024-11-05
Add error handling for when the profile file yaml parsing fails

VilppeRiskidev:


src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -121,7 +121,15 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
 
         if (read !== null) {
             this.profileImportFilePath = files[0];
-            this.profileImportContent = await ProfileUtils.parseYamlToExportFormat(read);
+            try {
+                this.profileImportContent = await ProfileUtils.parseYamlToExportFormat(read);
+            } catch (e: unknown) {
+                if(e instanceof R2Error) {
```
anttimaki, line 127: Well probably want to catch all errors here, including the once possibly thrown by `yaml.parse` or any future additions to the helper function. So I would recommend skipping the `if` check and just using `R2Error.fromThrownValue`.

src/utils/ProfileUtils.ts
```diff
@@ -125,6 +125,27 @@ export async function installModsToProfile(
...
 export async function parseYamlToExportFormat(yamlContent: string) {
     const parsedYaml = await yaml.parse(yamlContent);
+    if (!parsedYaml) {
+        throw new R2Error(
+            'Failed to parse yaml contents.',
+            'Yaml parsing failed when trying to import profile via file (The contents of export.r2x file are invalid).',
+            'Ensure that the profile import file isn\'t corrupted.'
+        )
+    }
+    if (!parsedYaml.profileName || typeof parsedYaml.profileName !== 'string') {
```
anttimaki, line 135: Nitpick: `typeof parsedYaml.profileName !== 'string'` would suffice here since `undefined` is not a string.´?

src/utils/ProfileUtils.ts
```diff
@@ -125,6 +125,27 @@ export async function installModsToProfile(
...
+        throw new R2Error(
+            'Failed to read profile name.',
+            'Reading the profile name after parsing the yaml failed (export.r2x is missing the profileName field).',
+            'Ensure that the profile import file isn\'t corrupted.'
+        )
+    }
+    if (!parsedYaml.mods || typeof parsedYaml.mods !== 'object') {
```
anttimaki, line 142: Looking at old L130/new L151 it looks like we're expecting an array, and while object might be an array, it might not. Using `if(!Array.isArray(parsedYaml.mods) {` would be more accurate, shorter, and make the intent clearer to the reader.

anttimaki (changes requested):
Looks pretty good but can be made more bullet proof with small changes.

PR #1525, opened 2024-11-05
Proof of concept Vue 2.7 composition API bridge

ebkr:


src/components/v2/Hero.vue
ebkr, line 1: The equivalent template is:
```html
<template>
    <section :class="['hero', heroType]" ref="section">
        <div class="hero-body non-selectable">
            <div class="container">
                <h1 class="title" ref="title">
                    {{title}}
                </h1>
                <h2 class="subtitle" ref="subtitle">
                    {{subtitle}}
                </h2>
            </div>
        </div>
    </section>
</template>

<script lang="ts">
import Vue from 'vue';
import { Component, Prop } from 'vue-property-decorator'

@Component
export default class Hero extends Vue {
    @Prop({default: ''})
    heroType: string | undefined;

    @Prop({default: ''})
    title: string | undefined;

    @Prop({default: ''})
    subtitle: string | undefined;
}
</script>
```

src/components/v2/Hero.vue
```diff
@@ -0,0 +1,24 @@
...
+                <h2 class="subtitle" ref="subtitle">
+                    {{subtitle}}
+                </h2>
+            </div>
+        </div>
+    </section>
+</template>
+
+<script lang="ts" setup>
+export interface HeroProps {
```
anttimaki, line 17: Is this exported because Vue demands it, or is it just in case some call site will need it?

src/pages/GameSelectionScreen.vue
```diff
@@ -164,7 +164,7 @@
 import { Component, Vue } from 'vue-property-decorator';
 import Game from '../model/game/Game';
 import GameManager from '../model/game/GameManager';
-import Hero from '../components/Hero.vue';
+import Hero from '../components/v2/Hero.vue';
```
anttimaki, line 167: Is the intention to create a v2 of all components, or can components that are currently used only in one place be replaced in-place?

anttimaki:
@VilppeRiskidev this moving to new style of components is something you need to be aware of too.

src/pages/GameSelectionScreen.vue
ebkr: It's primarily just to map what we have remaining.
Once we have all components inside a v2 folder, we'll migrate them back out. with a big bang replace of `/v2/` to `/`.

src/components/v2/Hero.vue
ebkr: Looks like we don't need it. Will remove.

VilppeRiskidev (commented):
Cool, nice to know! 👍

PR #1528, opened 2024-11-06
Error handling read profile file

VilppeRiskidev:


src/utils/ProfileUtils.ts
```diff
@@ -180,17 +180,23 @@ export async function populateImportedProfile(
...
+    let read: string | undefined;
     if (file.endsWith('.r2x')) {
         read = (await FsProvider.instance.readFile(file)).toString();
     } else if (file.endsWith('.r2z')) {
-        const result: Buffer | null = await ZipProvider.instance.readFile(file, "export.r2x");
-        if (result === null) {
-            return null;
-        }
-        read = result.toString();
+        await ZipProvider.instance.readFile(file, "export.r2x").then(
```
anttimaki, line 188: Combining `await` and `.then()` is quite confusing so I wouldn't use it unless it for some reason is absolutely necessary for the error handling. If combining the approaches, or even if just switching from `await` to `.then()` is for the error handling's sake, I'd leave a comment explaining why it's so. Otherwise someone like me tries to "fix it" later.

src/utils/ProfileUtils.ts
```diff
@@ -180,17 +180,23 @@ export async function populateImportedProfile(
...
+            (value) => { read = value?.toString() },
```
anttimaki, line 189: This won't compile for me (and the CI pipeline) so I can't test it. I assume you can't use the optional chaining here yet.

src/utils/ProfileUtils.ts
```diff
@@ -180,17 +180,23 @@ export async function populateImportedProfile(
...
+            () => { return null; }
```
anttimaki, line 190: As far as I can tell this server no purpose, and it's an optional argument so you could as well leave it out?

src/utils/ProfileUtils.ts
```diff
@@ -180,17 +180,23 @@ export async function populateImportedProfile(
...
+        );
+    }
+
+    if (!read) {
+        throw new R2Error(
+            'Error when reading file contents',
+            'Reading the .r2z file contents failed. File contents be empty or corrupted.',
```
anttimaki, line 197: > Reading the .r2z file contents failed.
It's not really the r2z that's the problem, but r2x?

> File contents be empty or corrupted.
This sounds a bit "street" way to phrase it? 😂 Should it be "are" instead of "be"?

src/utils/ProfileUtils.ts
```diff
@@ -180,17 +180,23 @@ export async function populateImportedProfile(
...
+            'Ensure that the imported profile file is valid.'
```
anttimaki, line 198: I'm not sure this is very helpful for someone trying to import a profile e.g. via code? Should we use something like "Ask the person who exported the profile to try again"? Or maybe just leave it empty?

anttimaki (changes requested):

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -117,9 +117,17 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
+            read = await ProfileUtils.readProfileFile(files[0]);
+        } catch (e: unknown) {
+            const err = R2Error.fromThrownValue(e);
+            this.$store.commit('error/handleError', err);
+            this.closeModal();
+            return;
+        }
 
-        if (read !== null) {
+        if (read) {
```
anttimaki, line 130: I think this `if`-check might not be needed, as the previous "step" should have returned if `read` is falsy? And if it somehow is false, we silently skip setting two important looking state variables, which might make any bugs harder to track? Unless you've got counter arguments, I'd prefer it removed.

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -129,14 +137,14 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
             }
+        }
 
-            if (this.profileToOnlineMods.length === 0) {
-                this.activeStep = 'NO_PACKAGES_IN_IMPORT';
-                return;
-            }
-
-            this.activeStep = 'REVIEW_IMPORT';
+        if (this.profileToOnlineMods.length === 0) {
```
anttimaki, line 142: Nitpick: could be onelined:
`this.activeStep = this.profileToOnlineMods.length ? 'REVIEW_IMPORT' : 'NO_PACKAGES_IN_IMPORT';`

anttimaki (changes requested):
The changes looks good. Couple more small things on the call site side. Let me know if you think they aren't needed, in which case I can approve this.

src/utils/ProfileUtils.ts
anttimaki: @VilppeRiskidev Did something happen here during the force push? I could swear this was fixed before, but now the `?` operator is back?

VilppeRiskidev: yeah, weirdly the changes were lost at some point :( I found some of them from IDE's local history. Remade the changes and should be good now.

src/utils/ProfileUtils.ts
```diff
@@ -201,17 +201,23 @@ export async function populateImportedProfile(
...
-        }
-        read = result.toString();
+        ZipProvider.instance.readFile(file, "export.r2x").then(
+            (value) => {
+                read = value ? value.toString() : null;
+            },
+        );
+    }
+
+    if (!read) {
```
anttimaki, line 216: This approach doesn't work for me at all. At this point the callback hasn't fired and `read` is always `undefined` for valid r2x files.

anttimaki (changes requested):

anttimaki (approved):
Approved 👍🏻

PR #1529, opened 2024-11-06
Update AdmZipProvider.ts

x753:
Ensure zips extract to intended paths

### DIFF
--- a/src/providers/generic/zip/AdmZipProvider.ts
+++ b/src/providers/generic/zip/AdmZipProvider.ts
@@ -8,6 +8,7 @@ export default class AdmZipProvider extends ZipProvider {
 
     async extractAllTo(zip: string | Buffer, outputFolder: string): Promise<void> {
         const adm = new AdmZip(zip);
+        outputFolder = outputFolder.replace(/\\/g, '/');
         adm.extractAllTo(outputFolder, true);
     }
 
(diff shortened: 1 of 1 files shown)


src/providers/generic/zip/AdmZipProvider.ts
```diff
@@ -27,16 +23,11 @@ export default class AdmZipProvider extends ZipProvider {
...
-    }
-
-    private async sanitizedExtraction(entry: IZipEntry, outputPath: string): Promise<void> {
-        const sanitizedTargetName = entry.entryName.split('\\').join('/');
-        await FileUtils.ensureDirectory(path.dirname(path.join(outputPath, sanitizedTargetName)));
-        if (entry.isDirectory)
-            await FileUtils.ensureDirectory(path.join(outputPath, sanitizedTargetName));
-        else
-            await FsProvider.instance.writeFile(path.join(outputPath, sanitizedTargetName), entry.getData());
+        if(!path.posix.normalize(path.join(outputPath, target)).startsWith(outputPath))
```
ebkr, line 26: This supposedly doesn't normalize `\` characters. So zips created on Windows have the potential to be invalid on Linux if I'm not mistaken?

x753: I believe I tested it on Linux with a Windows-created zip, but here's a new commit to address edge cases just in case it does matter.

ebkr (approved):

PR #1532, opened 2024-11-07
Disable modal buttons when in progress

VilppeRiskidev:


anttimaki (changes requested):
Brownie points for the separation of commits 👍🏻 

Create/rename forms can be submitted with enter, and as they do an extra check before calling the method that sets the `xInProgress` variable, they can be triggered twice. I realise that double-submit with enter is less likely than with a mouse click, but for completeness sake we could start the `keyup.enter` methods with `!xInProgress`?

anttimaki (approved):
Not exactly what I suggested but works for me so let's go.

PR #1536, opened 2024-11-08
Prevent disabling/enabling mod loaders

anttimaki:
Before the changes, mod loaders could be disabled, but in most cases it didn't really do anything. Disabling e.g. BepInEx and running the game modded still loaded the mods.

Defining how each mod loader should be actually disabled isn't trivial, and the need to actually disable them is low, as user can always just choose to run the vanilla game. Therefore it was decided that just hiding the disable option for package loaders was the best approach.

This adds 0-0.1ms rendering delay for each mod. While not ideal, it doesn't affect even the bigger profiles too badly.


VilppeRiskidev (approved):
LGTM 👍

PR #1539, opened 2024-11-09
MelonLoader Improvements

ebkr:
This PR fixes how we handle MelonLoader to prevent always passing `--agfregenerate` on MelonLoader 0.6.x.
Additionally we remove the "recommended" version for BONELAB as that should always be using the latest 0.6.x.

This functionality is pending a release of MelonLoader 0.6.6 (the next version) to work correctly. Once this has been released, we should expect to see correct behaviour with MelonLoader and BONELAB.


anttimaki (approved):
I'm not fully sure what's this is about but what I see makes sense to me. I assume it's intentional for BONEWORKS to keep using 0.5.4 as the recommended version?

ebkr:
Yeah it's intentional. It doesn't work with the newer versions of ML.

The reason for the change is that the manager keeps passing --agfregenerate which causes an extremely slow startup time every time `Start modded` is clicked. This reduces that and makes it a much nicer experience by only regenerating when the file doesn't exist.

PR #1542, opened 2024-11-11
Replace "preloader fix" with Steam game installation validation

anttimaki:


src/components/settings-components/SettingsView.vue
```diff
@@ -388,6 +380,14 @@ import CdnProvider from '../../providers/generic/connection/CdnProvider';
                         },
                         'fa-folder-open',
                         () => this.emitInvoke('ChangeSteamDirectory')
+                    ),
+                    new SettingsRow(
+                        'Debugging',
+                        'Validate Steam game installation',
```
ebkr, line 386: `Validate` is probably the wrong term? Makes it sound like it would just start Steam's validate functionality which we do, but it's not the only thing.

Options:
- `Clean and validate <game> installation`
- `Reset <game> installation`
- `Something else?`

Would be good to have the game name dynamic, similar to how it is for the `Change <game> folder` setting. It makes it easier to ensure that people don't click it for the wrong game.

src/pages/Help.vue
```diff
@@ -71,7 +71,7 @@
                 <p>That's because you don't legally own the game. The manager only supports legal copies.</p>
                 <hr/>
                 <h2 class='title is-5'>A text window appears and closes immediately.</h2>
-                <p>Try running the preloader fix on the Settings screen.</p>
+                <p>Try running "Validate Steam game installation" on the Settings screen.</p>
```
ebkr, line 74: Not necessary for this PR as the functionality is the same - Might be nice to make it an action that can actually run the setting? If it's high effort then not worth the hassle.

src/pages/Manager.vue
```diff
@@ -35,25 +35,29 @@
 			<button class="modal-close is-large" aria-label="close"
 			        @click="showRor2IncorrectDirectoryModal = false"></button>
 		</div>
-		<ModalCard :is-active="fixingPreloader" @close-modal="closePreloaderFixModal" :can-close="true">
+		<ModalCard :is-active="isValidatingSteamInstallation" @close-modal="closeSteamInstallationValidationModal" :can-close="true">
 			<template v-slot:header>
-				<h2 class='modal-title'>Attempting to fix preloader issues</h2>
+				<h2 class='modal-title'>Attempting to fix issues with game files</h2>
```
ebkr, line 40: Might be better as: `Clearing the <game> installation directory`

src/pages/Manager.vue
```diff
@@ -35,25 +35,29 @@
...
-					<p>You will not not be able to launch the game until Steam has verified the integrity of the
-						game.
+					<p>
+						You will not not be able to launch the game until
+						Steam has verified the integrity of the game files.
 					</p>
 				</div>
-				<p>Steam will be started, and will attempt to verify the integrity of {{ activeGame.displayName }}.</p>
+				<p>
+					Steam will be started, and will attempt to verify the
```
ebkr, line 50: Nit: We can remove the comma

src/pages/Manager.vue
```diff
@@ -35,25 +35,29 @@
...
 				<br/>
-				<p>Please check the Steam window for validation progress. If the window has not yet appeared, please be
-					patient.
+				<p>
+					Please check the Steam window for validation progress.
+					If the window has not yet appeared, please be patient.
 				</p>
 			</template>
 			<template v-slot:footer>
-				<button v-if="dependencyListDisplayType === 'view'" class="button is-info"
```
ebkr, line 55: This was interesting (being tied to `dependencyListDisplayType`), I wonder if it caused any issues before?

src/r2mm/manager/SteamInstallationValidator.ts
```diff
@@ -0,0 +1,64 @@
...
+
+        const gameFolder = await GameDirectoryResolverProvider.instance.getDirectory(game);
+        if (gameFolder instanceof R2Error) {
+            return gameFolder;
+        }
+
+        // Sanity check we're about to delete something resembling a valid game folder.
+        let exeFound = false;
+        for (let exeName of game.exeName) {
+            if (await FsProvider.instance.exists(path.join(gameFolder, exeName))) {1
```
ebkr, line 30: There's a trailing `1` here

ebkr (changes requested):

src/components/settings-components/SettingsView.vue
anttimaki: I went with `Reset <game> installation` as I find "reset" suggests kind of "factory reset" type of thing which is what we do here.

src/pages/Help.vue
anttimaki: Didn't include this now. Triggering the process would be simple enough, but I feel we'd need to describe what the actions does in more detail to avoid unexpected data loss, i.e. duplicate what the settings page currently says about the feature. Also we can't currently trigger the modal used by the settings page, as the help page isn't a subpage of the manager page.

src/pages/Manager.vue
anttimaki: The issue was that the buttons were not shown on the preloader fix modal and export-as-code modal. It wasn't a big deal as they were just "ok, close the modal" type of buttons and user had other means to close the modals.

src/r2mm/manager/SteamInstallationValidator.ts
anttimaki: What in tarnation?

ebkr (approved):

PR #1545, opened 2024-11-13
Download refactor separate callback

VilppeRiskidev:


src/utils/ModDownloadUtils.ts
```diff
@@ -23,3 +32,58 @@ export async function downloadCompletedCallback(profile: Profile, downloadedMods
         }
     });
 }
+
+export async function downloadProgressCallback (
```
anttimaki, line 36: Having this as a separate method in the vue component makes sense, as it removed duplicated code. However, I'm not so sure having it as a utility function is a good idea. The fact that this takes 11 arguments which mostly deal with managing the component's internal state means that it's not really a standalone in anyway. It's so tightly coupled with the current component, I'd argue that having it in a separate file actually makes the code harder to read. Also The Big Picture stuff argues against moving this here.

PR #1547, opened 2024-11-13
Simplify ProfileUtils.readProfileFile

anttimaki:
Users can't pick r2x files anymore, so supporting only r2z files will suffice. Also renamed the argument and a variable to be more descriptive.


anttimaki:
I think the error handling has been improved in the source since the PR was opened and this would regress it. Closing this now so it doesn't get merged until I've confirmed it's ok.

PR #1551, opened 2024-11-14
PoC of mod downloading through Vuex store

VilppeRiskidev:


src/store/modules/ModDownloadModule.ts
```diff
@@ -0,0 +1,97 @@
...
+            });
+
+            return new Promise((resolve, reject) => {
+                ThunderstoreDownloaderProvider.instance.download(
+                    params.profile,
+                    params.mod,
+                    params.version,
+                    true,
+                    (progress, modName, status, err) => {
+                        const index = state.allDownloads.findIndex((modProgressItem) => modProgressItem.mod.getName() === modName);
```
anttimaki, line 77: `findIndex` will return the first matching element, but active download is assumed to be the last item in the array. So if same mod is downloaded twice (e.g. download failure or reinstall after uninstallation), this will update wrong ProgressItem. Either reverse the list before searching, or change the approach so `allDownloads` is "newest first" (which might come handy in DownloadMonitor component).

src/store/modules/ModDownloadModule.ts
```diff
@@ -0,0 +1,97 @@
...
+
+                        if (status === StatusEnum.FAILURE) {
+                            if(err) {
```
anttimaki, line 80: What happens if I have multiple downloads in progress, and one in the background fails? I don't think we can necessarily throw these - we should try to store the error in the progress object and read/watch for it in the components. I'm not sure how to actually cancel the failed download right now.

src/store/modules/ModDownloadModule.ts
```diff
@@ -0,0 +1,97 @@
...
+                                reject(err);
+                            }
+                        }
+                        if (status === StatusEnum.PENDING) {
+                            if (index !== -1) {
```
anttimaki, line 85: I think this error should be handled earlier. Sort of those "this should never happen" situations.

src/store/modules/ModDownloadModule.ts
```diff
@@ -0,0 +1,97 @@
...
+                                state.allDownloads[index].downloadProgress = { progress: progress, status: status };
```
anttimaki, line 86: IIRC only mutations should, well, mutate the state, otherwise the change might not propagate correctly or there could be race conditions. So that's why this whole callback should be a mutation. (I know there's one getter is TsModsModule that also does this but it's clearly labeled a bad idea and done because unlike actions, getters don't have access to mutations.

src/store/modules/ModDownloadModule.ts
```diff
@@ -0,0 +1,97 @@
...
+            }
+        ): Promise<boolean> {
+            state.allDownloads.push({
+                mod: params.mod,
+                version: params.version,
+                installProgress: { progress: 0, status: StatusEnum.PENDING },
+                downloadProgress: { progress: 0, status: StatusEnum.PENDING }
+            });
+
+            return new Promise((resolve, reject) => {
```
anttimaki, line 70: > If combining await and .catch() is needed in the current state of refactoring, fine. But to manage expectations, I expect the final product to handle this in try { await ... } catch (e) { ... } finally { ... } manner.

In the same vein don't return a Promise object from here. After refactoring just await the different steps and return undefined (automatically turned into promise as this an async method.

anttimaki (changes requested):

src/store/modules/ModDownloadModule.ts
VilppeRiskidev: added a TODO comment about removing the completedCallback.

src/store/modules/ModDownloadModule.ts
VilppeRiskidev: added a TODO comment about displaying the error(s).

VilppeRiskidev:
Added a few TODO comments, which (I think) should be addressed as separate PRs. Mainly:
- Dropping the completedCallback and awaiting instead
- Do something with the downloaded dependencies (track/display them somehow/somewhere)
- Error displaying (when downloads fail)
- Closing the modal (which I think will be done naturally when the completedCallback is removed and the installation is also implemented)

src/components/views/DownloadModModal.vue
```diff
@@ -136,6 +136,11 @@ let assignId = 0;
 
         static allVersions: [number, DownloadProgress][] = [];
 
+
+        get activeDownload() {
```
anttimaki, line 140: To build on my earlier similar comment about a progress-getter returning undefined or number: due to the poor type support of the Vuex version we're using, this will now return `any` type, which makes it difficult to detect errors that occur if the value is actually `undefined`. So

a) always check that these getters have a properly inferred return type. If Vuex is involved, changes are it doesn't. If the inferred type is wrong, provide correct return type manually (it's IMO ok to always do this to avoid bugs from future changed
b) Unless a getter like this can provide a sane fallback value, make sure the return type include `undefined` (or whatever the correct type is) so incorrect usage elsewhere causes an error.

src/store/index.ts
```diff
@@ -128,6 +129,7 @@ export const store = {
         profile: ProfileModule,
         profiles: ProfilesModule,
         tsMods: TsModsModule,
+        modDownload: ModDownloadModule,
```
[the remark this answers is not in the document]
anttimaki: It's a bit confusing to have 👍🏻 reaction but not have anything changed. If this falls under wontfix, please be more explicit of that intention.

src/store/modules/ModDownloadModule.ts
```diff
@@ -0,0 +1,97 @@
...
+import ThunderstoreCombo from '../../model/ThunderstoreCombo';
+import ThunderstoreMod from '../../model/ThunderstoreMod';
+import ThunderstoreVersion from '../../model/ThunderstoreVersion';
+import ThunderstoreDownloaderProvider from '../../providers/ror2/downloading/ThunderstoreDownloaderProvider';
+
+import { State as RootState } from '../index';
+
+interface Progress {
+    progress: number;
+    status: number; // StatusEnum
```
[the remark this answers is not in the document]
anttimaki: Since the Enum is added it should be used for typing here.

src/store/modules/ModDownloadModule.ts
```diff
@@ -0,0 +1,112 @@
...
+        downloads: [],
+    }),
+
+    getters: <GetterTree<State, RootState>>{
+        activeDownload(state): ProgressItem | undefined {
+            return state.downloads.slice(-1)[0]; // Last element of the array
+        },
+    },
+    mutations: {
+        reset(state: State) {
```
anttimaki, line 44: Nitpick: this is currently unused and I'd wager won't be needed in the future either as we probably want to keep downloads going on even if user changes profile/game (if that's practical to implement). A use case for something similar could be "clear all completed downloads" on DownloadMonitor view, but in that case we probably want to call the method something that doesn't get confused with the other Vuex reset methods.

src/store/modules/ModDownloadModule.ts
```diff
@@ -0,0 +1,112 @@
...
+            state.downloads.push({
+                modName: modName,
+                installProgress: 0,
+                downloadProgress: 0,
+                status: StatusEnum.PENDING,
+                error: null
+            });
+        },
+        updateDownloadProgress(state: State, params: { progress: number, modName: string, status: number, err: R2Error | null }) {
+            let downloadMod = state.downloads.find((progressItem) => progressItem.modName === params.modName);
```
anttimaki, line 57: As mentioned in the earlier reviews, .find() returns the first matching item which doesn't work if we have multiple ProgressItems for the same mod (e.g. first one has failed or user installs a different version). To think of it, we therefore probably need some different way of separating the ProgressItems than a simple mod name.

src/store/modules/ModDownloadModule.ts
```diff
@@ -0,0 +1,112 @@
...
+
+            // Doesn't do anything when the downloaded mod is not found in the array (which means it is a dependency).
```
anttimaki, line 59: Seem dangerous to assume it's always a dependency instead of just a bug. Also based on gut feeling I think we're going to need track all the combos in the progress item, not just the "main download". This includes dependencies when downloading a single mod, and multiple mods with their dependencies when updating all installed mods to latest versions.

src/store/modules/ModDownloadModule.ts
```diff
@@ -0,0 +1,112 @@
...
+            // TODO: figure out if/how/where to track/display these dependencies' progresses.
```
anttimaki, line 60: My intuition says postponing planning how this is going to be handled before implementing stuff is like painting yourself towards a corner. Overall I think, despite having the "Big Picture" plan of the changes, we're going to need more detailed plan of the steps before progressing the implementation instead of just trying different things and seeing how it goes.

src/store/modules/ModDownloadModule.ts
```diff
@@ -0,0 +1,112 @@
...
+
+        //TODO: Do the installation
+        async downloadAndInstallMod(
+            {dispatch, getters, state, commit},
+            params: {
+                profile: ImmutableProfile,
+                mod: ThunderstoreCombo,
+            }
+        ) {
+            state.downloads.push({
```
anttimaki, line 84: As per an earlier review comment, don't mutate state in the actions, use commits instead. In fact you have a mutator for this exact thing, but then it's not used for some reason.

src/store/modules/ModDownloadModule.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+interface State {
+    downloads: ProgressItem[];
+}
+
+/**
+ * State for Mod Downloading/Installation.
+ */
+export default {
```
anttimaki, line 31: Nitpick: use named exports whenever possible. Default exports can lead to errors when imported carelessly.

anttimaki (changes requested):
Added comments which contain things worth keeping in mind. However, they might be a bit moot as the my main gripe with the PR is that it's not what I meant yesterday when I said to have a small PR that can be approved even if it's not merged until more work is done in the branch.

What I meant was to have small PRs that do one thing which takes the code base from one working state to another working state. E.g. what we did with ImportModModal at first. A good rule of thumb for this could be that the PR improves the code even if it would be the only change we did. I.e. it could do one or more of the following:

- Remove business logic from the Vue component
- Extract a nice standalone part from a long function to a separate well named function
- Remove duplicate code by refactoring it into a function
- Clean up a messy part of implementation

It seems to me this PR is still too ambitious - it does a bit of this and a bit of that, and the result is lot of TODO comments and a state that doesn't really work anymore. If we continue to develop the feature on top of this PR, I think we'll have to constantly keep it in mind and double check the future changes, instead of having a stack of PRs we consider done and don't have to think about anymore.

Let's continue the discussion in the daily.

src/store/index.ts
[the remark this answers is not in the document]
VilppeRiskidev: oops, my mistake

PR #1552, opened 2024-11-14
Reduce time spent on the splash screen

anttimaki:
Users may want to get to the manager screen quickly so they can just launch the game with existing mods. Having to wait for the latest mod list update can be frustrating.

Splash screen now only checks the API for a fresh list if no local cache is found at all. Handling different error scenarios is also omitted, and user is moved to profile selection screen even if no mod list can be loaded. Error screen offering a chance to retry loading the mod list is dropped. The idea is that the background update or a manually triggered mod list update will handle the mod list eventually.

Note that the changes only affect the splash screen. Background updates still work the same way they did before.

In the future the splash screen can be hastened even further e.g. by handling the loading of the exclusion list in the background and reducing the timeouts/retry attempts of the API calls.

### DIFF
--- a/src/components/mixins/SplashMixin.vue
+++ b/src/components/mixins/SplashMixin.vue
@@ -2,7 +2,6 @@
 import Vue from 'vue';
 import Component from 'vue-class-component';
 
-import R2Error from '../../model/errors/R2Error';
 import RequestItem from '../../model/requests/RequestItem';
 import type { PackageListChunks, PackageListIndex } from '../../store/modules/TsModsModule';
 
@@ -11,7 +10,6 @@ import type { PackageListChunks, PackageListIndex } from '../../store/modules/Ts
 export default class SplashMixin extends Vue {
     heroTitle = '';
     heroType: string = 'primary';
-    isOffline = false;
     loadingText = '';
     requests: RequestItem[] = [];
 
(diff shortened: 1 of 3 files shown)


src/components/mixins/SplashMixin.vue
```diff
@@ -54,32 +52,49 @@ export default class SplashMixin extends Vue {
...
+
+        if (!hasPriorCache) {
+            const packageListIndex = await this.fetchPackageListIndex();
+            const packageListChunks = await this.fetchPackageListChunksIfUpdated(packageListIndex);
+            this.getRequestItem('ThunderstoreDownload').setProgress(100);
+            hasUpdatedCache = await this.writeModsToPersistentCacheIfUpdated(packageListIndex, packageListChunks);
         }
+
+        if (hasPriorCache || hasUpdatedCache) {
+            await this.readModsToVuex();
```
ebkr, line 68: This naming could be better by not referring to the store implementation but rather what we're doing with it.

readMods <- implies the behaviour is to be parsed, not written to.

Might be better to call it something along the lines of: `triggerStoreModListUpdate`

src/components/mixins/SplashMixin.vue
```diff
@@ -133,52 +148,31 @@ export default class SplashMixin extends Vue {
...
+    async readModsToVuex(): Promise<void> {
         try {
             await this.$store.dispatch('tsMods/updateMods');
-            isModListLoaded = true;
         } catch (e) {
-            this.$store.commit(
-                'error/handleError',
-                R2Error.fromThrownValue(e, `${this.loadingText} failed`)
-            );
+            console.error('SplashMixin failed to read mods to Vuex.', e);
```
ebkr, line 170: Following the above comment, the `read` here is misleading

src/components/mixins/SplashMixin.vue
anttimaki: Changed as suggested.

src/components/mixins/SplashMixin.vue
anttimaki: Comment updated.

PR #1553, opened 2024-11-14
Add ModListUpdateBanner

anttimaki:
The banner will show on top of an empty online mod list, explaining that the list is empty because splash screen failed to load it. A one-click option for trying to update the mod list is offered.

This change is adjustment to the fact that the splash screen no longer stops on errors and user is more likely to reach the manager view with no mod list loaded into Vuex.

### DIFF
--- a/src/components/ModListUpdateBanner.vue
+++ b/src/components/ModListUpdateBanner.vue
@@ -1,0 +1,73 @@
+<script lang="ts">
+import { mixins } from 'vue-class-component';
+import { Component, Watch } from 'vue-property-decorator';
+
+import UtilityMixin from './mixins/UtilityMixin.vue';
+
+@Component({})
+export default class ModListUpdateBanner extends mixins(UtilityMixin) {
+    updateError = '';
+
+    get isModListLoaded(): boolean {
+        return this.$store.state.tsMods.modsLastUpdated !== undefined;
+    }
+
+    get isUpdateInProgress(): boolean {
+        return this.$store.state.tsMods.isBackgroundUpdateInProgress;
+    }
+
+    @Watch('isUpdateInProgress')
+    onLoadingProgressChange(newVal: boolean, oldVal: boolean) {
+        // React to background update as well as the manual update from the banner.
+        if (!oldVal && newVal) {
+            this.updateError = '';
+        }
+    }
+
+    async updateModList() {
+        if (this.isUpdateInProgress) {
+            return;
+        }
+
+        this.$store.commit('tsMods/startBackgroundUpdate');
+        this.updateError = '';
+
+        try {
+            // Invalidate hash to force a refresh. Otherwise a scenario where
+            // the latest index hash is already present in IndexedDB but loading
+            // the package list into Vuex store has failed would cause the banner
+            // to just disappear when refreshThunderstoreModList skips the actual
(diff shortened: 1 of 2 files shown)


src/components/ModListUpdateBanner.vue
```diff
@@ -0,0 +1,78 @@
+<script lang="ts">
+import { mixins } from 'vue-class-component';
+import { Component, Watch } from 'vue-property-decorator';
+
+import UtilityMixin from './mixins/UtilityMixin.vue';
+import ManagerInformation from '../_managerinf/ManagerInformation';
+
+@Component({})
```
ebkr, line 8: Minor nit: I don't believe we need the `{}`

ebkr, line 8: It could be worth moving to the Vue 3 Composition API - I won't make it a blocker but it's something to consider

src/components/ModListUpdateBanner.vue
```diff
@@ -0,0 +1,78 @@
...
+
+<template>
+    <div v-if="!isModListLoaded" id="mod-list-update-banner" class="margin-bottom">
+        <div class="notification is-warning margin-right">
+            <span v-if="isUpdateInProgress">
+                Updating mod list from Thunderstore...
+            </span>
+            <span v-else-if="updateError">
+                Error updating the mod list: {{ updateError }}<br />
+                {{ appName }} will keep trying to update the mod list in the background.
```
ebkr, line 65: Nit: We could generify this by using `The manager` or `We`

To counter the design a little, we could have:

```
There was an error updating the mod list. <a>Show the error</a>.
{{ appName }} will keep trying to update the mod list in the background.
```
The thinking is that the `a` tag would open a modal and we can keep the normal error preview.

src/components/ModListUpdateBanner.vue
```diff
@@ -0,0 +1,78 @@
...
+            </span>
+            <span v-else>
+                It seems like the latest mod list hasn't been loaded from Thunderstore yet.
```
ebkr, line 68: I'd refrain from uncertainty because we know what occurred to get into this state. 
We can just say something like:
```
An error occurred when updating the mod list. Would you like to <a>try again?</a>
```

src/components/ModListUpdateBanner.vue
anttimaki: One argument (options) is required.

anttimaki: The last time I looked into Composition API I couldn't figure out how to use Vuex store in the new components, and I still don't. So I'd rather not get sidetracked with it now. Unless you have some pointers on how to get it working easily?

src/components/ModListUpdateBanner.vue
anttimaki: Changed to `The manager`.

I can do another PR for the error modal suggestion, but do you think it still matters what we do here if the profile import modal (in #1598 ) will show the error "inline" like here? I'd rather not use the error modal there, as that would interfere with the progress made on import modal.

src/components/ModListUpdateBanner.vue
anttimaki: Changed the texts along the lines of the suggestion.

PR #1554, opened 2024-11-14
Update package exclusion list in Splash screen only if it's needed

anttimaki:


ebkr (approved):
Approving. I'd expect a change later to update the exclusion list alongside the "Failed to update mod list" banner though, or better, whenever we refresh the mod list. 

It might already do this in which case feel free to correct me and disregard it entirely.

PR #1571, opened 2024-11-29
Separate Version Selction Modal from DownloadModModal.vue

VilppeRiskidev:


VilppeRiskidev:
Progress modal doesn't open when mod downloading is triggered from the version selection modal. It works when doing the update all mods. Probably caused by Vue not updating, I'll look into this.

src/components/mixins/DownloadMixin.vue
```diff
@@ -0,0 +1,32 @@
...
+import Vue from 'vue';
+import Component from 'vue-class-component';
+
+import ThunderstoreMod from "../../model/ThunderstoreMod";
+import Profile from "../../model/Profile";
+
+
+@Component
+export default class DownloadMixin extends Vue {
+    get ignoreCache(): boolean {
```
anttimaki, line 11: This is currently used only in DownloadModModal which redefines the method. Seems it's not needed in the mixin.

src/components/views/DownloadModModal.vue
```diff
@@ -15,56 +15,14 @@
...
-                            {{selectedVersion}} is an outdated version
-                        </span>
-                    </div>
-                </div>
-            </template>
-            <template v-slot:footer>
-                <button class="button is-info" @click="downloadThunderstoreMod()">Download with dependencies</button>
-            </template>
-        </ModalCard>
+        <DownloadModVersionSelectModal
```
anttimaki, line 18: To clarify, when we discussed separating the modals, I assumed we we're talking about making the components independent, whereas the version select modal is now a subcomponent. But I consider the subcomponent a step forwards, so it's fine for this PR. Just keep in mind we might (or might not) want to separate these even more in the future.

src/components/views/DownloadModModal.vue
```diff
@@ -15,56 +15,14 @@
...
+            :versionNumbers="versionNumbers"
+            :selected-version="selectedVersion"
+            :currentVersion="currentVersion"
+            :recommendedVersion="recommendedVersion"
```
anttimaki, line 22: This and the next comment are about the same thing but I write them separately to make it clearer what my train of thought was. Read them both before doing changes.

It looks to me that these four props are passed to child component somewhat unnecessarily. If you look at what `getModVersions` does, it seems that it should be a part of the child component, not the parent. If the watched method is moved to the child, only thing the parent cares about is the selected version.

Also the way `selectedVersion` and `setSelectedVersion` is passed here is a bit React-ish. It will probably work without problems, but using emits would be more in line with Vue. You could take a look at how `CategoryFilterModal` and `CategorySelectorModal` currently interact. It's not an exact match so don't copy it as-is, and read the Vue docs about events/emits to get the whole picture.

src/components/views/DownloadModModal.vue
```diff
@@ -15,56 +15,14 @@
...
+            :download-thunderstore-mod="downloadThunderstoreMod"
```
anttimaki, line 23: Now, if you followed the idea from the earlier comment, and it worked, it's quite easy to see that the only reason the parent component is interested in `selectedVersion` is that it's needed by `downloadThunderstoreMod`. Furthermore, the parent component doesn't care about `downloadThunderstoreMod` as it's only passed to the child component, which means that it should reside in the child component, not the parent.

If that also works without problems, the next step in chain would be `downloadHandler`, which is actually also only used by child component. However, the internal implementation of `downloadHandler` is tightly coupled with the parent component, so it might be a good spot limit the scope of this PR and leave it into the parent.

That then leaves the question of how `DownloadModVersionSelectModal.downloadThunderstoreMod` should trigger `DownloadModModal.downloadHandler`? For this PR I'd say the subcomponent should emit an event with a `ThunderstoreCombo` as a payload, and the parent should catch the event and call `downloadHandler`. If you're thinking that Vuex could be used for that as well, you're right, but lets leave that for another PR.

src/components/views/DownloadModVersionSelectModal.vue
```diff
@@ -0,0 +1,82 @@
...
+        ModalCard
+    },
+})
+export default class DownloadModVersionSelectModal extends mixins(DownloadMixin) {
+    @Prop({ required: true }) versionNumbers!: string[];
+    @Prop({ required: true }) selectedVersion!: string | null;
+    @Prop({ required: true }) currentVersion!: string;
+    @Prop({ required: true }) recommendedVersion!: string;
+    @Prop({ required: true }) downloadThunderstoreMod!: Function;
+    @Prop({ required: true }) setSelectedVersion!: Function;
```
anttimaki, line 75: Moot point if my earlier comments check out, but these could use more accurate typing, e.g. `(selectedVersion: string) => void` for `setSelectedVersion`.

src/components/views/DownloadModVersionSelectModal.vue
```diff
@@ -0,0 +1,82 @@
...
+
+    handleChange(event: Event) {
+        this.$props.setSelectedVersion((event.target as HTMLSelectElement).value);
```
anttimaki, line 78: I don't think the `.$props` part is necessary and it might affect the typing.

anttimaki (changes requested):
Overall this PR looks nice but I'd argue it can be made better.

src/components/views/DownloadModModal.vue
VilppeRiskidev: Should be better now (in the next commit).
I banged my head against the wall trying to do this exact thing last night for hours, but it just wouldn't work. Well, at least it works now.

src/components/views/DownloadModModal.vue
VilppeRiskidev: yeah, I think the downloadHandler should be _handled_ 🥁 by vuex once it is time for that.

src/components/views/DownloadModModal.vue
```diff
@@ -98,17 +97,17 @@ let assignId = 0;
...
-
         get ignoreCache(): boolean {
             const settings = this.$store.getters['settings'];
             return settings.getContext().global.ignoreCache;
         }
 
-        setSelectedVersion(newSelectedVersion: string) {
-            this.selectedVersion = newSelectedVersion;
+        selectedVersionChanged(event: Event) {
+            if (!(event.target instanceof HTMLSelectElement)) {
```
anttimaki: Moot point now, but it would be nicer if this was done in the child the component that emits, as that's where the markup and thus the event target is defined. The parent component is only interested in the version string, so we should pass only that and hide the internal implementation of the child component.

anttimaki (commented):
I'm not entirely sure why but you skipped the part 1 of my Big Comment and implemented only the part 2. The changes you did for part 2 look good, but we don't really get the benefits unless we also implement the part 1. So I added a commit that implements the part 1. At least for me this seems to work nicely, but please test it thoroughly.

I also added a second cleanup commit that probably shouldn't be a part of this PR but I couldn't resist as the changes of me implementing it separately later are slim.

So, Uno reverse card time, please review and test my changes.

VilppeRiskidev (commented):
Very good (can't approve this myself). IDK how I missed the part in your comment about what you did with these commits, but yeah, makes sense, much cleaner this way. Nice cleanup also.

anttimaki (approved):
Approved based on my earlier and Vilppe's later reviews.

PR #1576, opened 2024-12-03
Remove obsolete converter interface

VilppeRiskidev:


anttimaki (changes requested):
My bad for not mentioning this on the ticket, but the current changes only remove the interface and not the class methods the interface made mandatory. The point was to remove `.fromReactive()` methods and their usages as well, as they're the part that creates unnecessary overhead, unnecesasrily casting the object to the same object again.

src/r2mm/mods/ProfileModList.ts
```diff
@@ -43,7 +43,7 @@ export default class ProfileModList {
             try {
                 const value = (yaml.parse((await fs.readFile(profile.joinToProfilePath('mods.yml'))).toString()) || []);
                 for(let modIndex in value){
-                    const mod = new ManifestV2().fromReactive(value[modIndex]);
+                    const mod = value[modIndex];
```
VilppeRiskidev, line 46: This could probably be refactored a bit (by just using `value[modIndex]` instead of creating the local `mod` variable, but idk might be that it needs a copy for some reason. Would need testing that it works, so I decided to not use time to do that.)

anttimaki: A bit of testing would always be a good idea. In this case it would've caught that the current changes spectacularly break things. A good lesson in "never believe what Antti says, always verify it yourself" 😂 

The issue is that `.fromReactive` was meant for fixing the objects that get passed through a Vue component (e.g. `v-for="(mod) in mods"`). What I failed to notice is that the usage here is completely different - someone has just decided to use the existing method since it has worked for their use case. Here the loop is one where ManifestV2[] has been cast to Yaml, written to disk, read as Yaml from the disk, and parsed into a plain JS object. So here we actually do need to cast it into a ManifestV2 again.

I'm not exactly sure what the best way to solve this would be, as I can think of a few ways. They're outside the scope of the task though, so I'd suggest you restore `ManifestV2.fromReactive` but rename it into something more descriptive like `fromJsObject`.

anttimaki: BTW assigning an object into a variable in JS doesn't copy the object, the variable only references the existing object (its location in memory). But yeah, the variable is used here probably for readability reasons and could arguably be removed.

src/model/VersionNumber.ts
```diff
@@ -1,6 +1,4 @@
-import ReactiveObjectConverterInterface from './safety/ReactiveObjectConverter';
-
-export default class VersionNumber implements ReactiveObjectConverterInterface {
+export default class VersionNumber {
```
anttimaki, line 1: Didn't the earlier version of the PR also remove `fromReactive` from here? Why was that change dropped, I think the method should still go from this class.

src/model/ManifestV2.ts
```diff
@@ -108,7 +107,7 @@ export default class ManifestV2 implements ReactiveObjectConverterInterface {
         return this;
     }
 
-    public fromReactive(reactive: any): ManifestV2 {
+    public fromJsObject(reactive: any): ManifestV2 {
```
anttimaki, line 110: NGL I'm a bit irked about the argument name still being "reactive". Not a blocker but wink wink nudge nudge.

anttimaki (changes requested):

src/model/VersionNumber.ts
VilppeRiskidev: Whoops, I did an oopsie when reverting changes

anttimaki (approved):

PR #1583, opened 2024-12-11
Add RecursiveMelonLoaderInstaller

anttimaki:
MelonLoader has added support for recursive loading. This means mods, plugins and userLibs can all be placed in the same folder inside the profile folder instead of needing to copy each type to a specific subfolder. This in turn means we can namespace each installed mod into its own subfolder, which makes uninstalling the mods straightforward and removes the need to track installation state on a separate file.

The mod author still need to place plugins and userLibs to specific folders in their Thunderstore packages though.

In the profile, package's UserData folder (if any) gets copied to profile/UserData/[packageName] and everything else gets copied to profile/Mods/[packageName]. MelonLoader handles the rest.

This will require MelonLoader v0.6.7 or newer to work.

### DIFF
--- a/src/installers/RecursiveMelonLoaderInstaller.ts
+++ b/src/installers/RecursiveMelonLoaderInstaller.ts
@@ -1,0 +1,138 @@
+import path from 'path';
+
+import {
+    disableModByRenamingFiles,
+    enableModByRenamingFiles,
+    InstallArgs,
+    PackageInstaller
+} from './PackageInstaller';
+import FileWriteError from '../model/errors/FileWriteError';
+import R2Error from '../model/errors/R2Error';
+import VersionNumber from '../model/VersionNumber';
+import FsProvider from '../providers/generic/file/FsProvider';
+import FileUtils from '../utils/FileUtils';
+
+/**
+ * Handles (un)installation of MelonLoader v0.7.0 and above.
+ */
+export class RecursiveMelonLoaderInstaller implements PackageInstaller {
+    private static readonly TRACKED = ['MelonLoader', 'version.dll'];
+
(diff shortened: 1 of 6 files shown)


anttimaki:
This is a draft PR as the MelonLoader needs a new release for things to work as expected.

As for the naming, if we end up having two ML installer implementations at the same time, the new one could be renamed to RECURSIVE_MELON_LOADER, or the old one could be renamed to LEGACY_MELON_LOADER. If there's going to be only one then simple MELON_LOADER will do.

Currently (confirmtion pending) the idea is that mod packages would be structured as:

```
# mod.zip
  - Mods
    - my_first_mod.dll
  - Plugins
    - my_plugin.dll
  - UserLibs
    - helper_lib.dll
  - UserData
    - <custom folders and files>
  - my_second_mod.dll
  - <required files>
```

The files would get copied to profiles as:

```
# <profile>/
  - Mods
    - <packageName>
      - Mods
        - my_first_mod.dll
      - Plugins
        - my_plugin.dll
      - UserLibs
        - helper_lib.dll
      - my_second_mod.dll
      - <required files>   
  - UserData
    - <PackageName>
      - <custom folders and files>
```

The `<required files>` are copied to Mods folder as a `manifest.json` is needed to trigger the recursive loading of mods/plugins/user libs.

anttimaki:
- Draft status no longer warranted
- Renamed the new version to RecursiveMelonLoaderInstaller
- Supports v0.7.0+
  - v0.7.0 requires manifest.json in the folder to trigger recursive loading, but this requirement will be removed in v0.7.1. The update doesn't require changes to mod manager
- Existing games will keep using the older implementation until the respective communities have declared themselves ready to migrate

PR #1587, opened 2024-12-12
Move mod version downloading logic and state management to Vuex state

VilppeRiskidev:


src/components/views/DownloadModModal.vue
```diff
@@ -92,28 +88,36 @@ let assignId = 0;
...
+                        const assignIndex = store.state.download.allVersions.findIndex(([number, val]: [number, DownloadProgress]) => number === currentAssignId);
                         if (status === StatusEnum.FAILURE) {
                             if (err !== null) {
-                                const existing = DownloadModModal.allVersions[assignIndex]
+                                const existing = store.state.download.allVersions[assignIndex]
                                 existing[1].failed = true;
-                                DownloadModModal.allVersions[assignIndex] = [currentAssignId, existing[1]];
+                                store.commit('download/updateDownloadObject', {
+                                    assignId: assignIndex,
+                                    downloadObject: [currentAssignId, existing[1]]
```
anttimaki, line 119: The argument type here is not what you've defined the downloadObject to be in the Vuex store (it expects only DownloadProgress object. This can be fixed either on the call sites or on the Vuex method. Do note that there are multiple call sites to fix if you choose to go that route.

I know the original code is confusing but be careful about the arguments of Vuex methods as the type support doesn't really exist.

src/store/modules/DownloadModule.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    mutations: {
+        setDownloadObject(state: State, downloadObject: DownloadProgress) {
+            state.downloadObject = Object.assign({}, downloadObject);
+        },
+        pushDownloadObjectToAllVersions(state: State, params: { assignId: number, downloadObject: DownloadProgress }) {
+            state.allVersions.push([params.assignId, params.downloadObject]);
+        },
+        updateDownloadObject(state: State, params: { assignId: number, downloadObject: DownloadProgress }) {
+            Vue.set(state.allVersions, params.assignId, params.downloadObject);
```
anttimaki, line 38: I think using the Vue method here is unnecessary, and not ideal from readability perspective (the reader might not know what exactly the set method does, but will probably know what the basic JS array methods do). And like above, just to be sure I'd rather create a new array rather than mutate the existing. Something like this (but do note the earlier comment about the arguments passed to this function not really matching the current argumen typing):

```
const newVersions = [...state.allVersions];
newVersions[params.assignId] = params.downloadObject;
state.allVersions = newVersions;
```

anttimaki (changes requested):
The main idea is as discussed but the argument typing should be fixed and preferrably the mutation of objects in the Vuex could be replaced with copying.

anttimaki (approved):
As a nitpick I'm not too sure about the `downloadVersion` argument name, but that'll probably get reworked in the next PR anyway, so I'd say this is good as it is :+1:

VilppeRiskidev:
This is now a part of https://github.com/ebkr/r2modmanPlus/pull/1594

PR #1588, opened 2024-12-13
Cleaning up LocalModInstaller

anttimaki:


anttimaki:
@VilppeRiskidev updated the middle commit. Earlier it completely removed the writing of the file as it seemed some obsolete legacy feature. It's now capsulated in a separate method that explains it's purpose instead.

PR #1594, opened 2024-12-17
Download callback refactor

VilppeRiskidev:


VilppeRiskidev:
I based this branch on the previous branches since those have been approved already and are just waiting to be merged.

src/components/views/DownloadModModal.vue
```diff
@@ -57,7 +57,7 @@ import Profile from '../../model/Profile';
 import { Progress } from '../all';
 import ConflictManagementProvider from '../../providers/generic/installing/ConflictManagementProvider';
 import ModalCard from '../ModalCard.vue';
-import { installModsToProfile } from '../../utils/ProfileUtils';
+import {installModsAndResolveConflicts, installModsToProfile} from '../../utils/ProfileUtils';
```
anttimaki: Nitpick: seems you've chosen to omit the spaces within braces here, although the convention in code base seems to favour them, at least in the context of imports.

src/utils/ProfileUtils.ts
```diff
@@ -56,6 +58,22 @@ async function extractConfigsToImportedProfile(
     }
 }
 
+export async function installModsAndResolveConflicts(
```
anttimaki, line 61: Food for a thought, not a change request:

Given that this is in the end a rather simple piece of code, it could be argued that having it in the utils is a bit of overkill, especially since we have another wrapper in the mixin that reduces duplication. Another thing that IMO speaks a bit against having it here is the fact that we need to pass the store as an argument. That's not inherently bad, but a weak signal that the split between UI component and helper library should be considered. For example, this helper is now not reusable if the call site doesn't have access to the store. Ideally we could import the store singleton to helpers but I _think_ that's currently unnecessarily difficult due to R2MM & TSMM having different store definitions.

I guess that my point is that I've no strong arguments one way or another, but my nitpick level opinion is that passing the store as an argument should be avoided if it's easy to do so.

src/utils/ProfileUtils.ts
```diff
@@ -56,6 +58,22 @@ async function extractConfigsToImportedProfile(
...
+    downloadedMods: ThunderstoreCombo[],
+    profile: Profile,
+    store: Store<any>
+): Promise<void> {
+    await ProfileModList.requestLock(async () => {
+        try {
+            const modList: ManifestV2[] = await installModsToProfile(downloadedMods, profile.asImmutableProfile());
+            await store.dispatch('profile/updateModList', modList);
+            throwForR2Error(await ConflictManagementProvider.instance.resolveConflicts(modList, profile));
```
anttimaki, line 70: Downside of having a long running the stack of branches is that you'll need to rebase the whole stack to avoid conflicts. E.g. here your earlier changes mean this method no longer accepts Profile but ImmutableProfile, meaning `installModsAndResolveConflicts` should only deal in ImmutableProfiles too.

I guess we should try to merge the approved branches after the holidays even if the refactoring as whole isn't completed yet. In the meanwhile you might just keep doing the rebases manually, or try Graphite*, or do this shuffle where you keep only the latest approved PR (which contains the commits from earlier PRs too) and drop the earlier PRs so you'll only have one approved PR the rebase (this explanation probably sucked, we can go through what I meant in the daily).

*: I've personally used Graphite to handle the branches locally on my dev environment, put handled the pushing with vanilla git and created the PRs manually, as I'm not sure if it's good idea to include Graphite stuff in a repo where most other authors don't use it.

src/utils/ProfileUtils.ts
```diff
@@ -16,6 +16,8 @@ import ZipProvider from "../providers/generic/zip/ZipProvider";
 import ProfileInstallerProvider from "../providers/ror2/installing/ProfileInstallerProvider";
 import * as PackageDb from '../r2mm/manager/PackageDexieStore';
 import ProfileModList from "../r2mm/mods/ProfileModList";
+import ConflictManagementProvider from "src/providers/generic/installing/ConflictManagementProvider";
+import {Store} from "vuex";
```
anttimaki, line 20: Nitpick: if we want to keep the import order logic, this should be located before yaml as it's a 3rd party import.

src/components/views/DownloadModModal.vue
```diff
@@ -214,7 +213,10 @@ interface DownloadProgress {
                         downloadVersion: [currentAssignId, obj]
                     });
                 }
-            }, this.downloadCompletedCallback);
+            }, (downloadedMods) => {
+                this.downloadCompletedCallback(downloadedMods);
+                this.downloadingMod = false;
```
anttimaki, line 218: Just mentioning this in the hopes we remember to do this in the future PRs: this should be something like `isDownloadProgressModalOpen` and probably to be handled into `ModalModule` store. Ideally in a way that when one of the three modals in this file is opened, the other two are automatically closed.

anttimaki (approved):
The changes look good 👍🏻 I still found some reasons to flap my gums so read the commits for future reference.

src/utils/ProfileUtils.ts
anttimaki: We decided to close the earlier PRs and point this PR to develop, so it now contains already approved changes from earlier PRs.

src/utils/ProfileUtils.ts
```diff
@@ -56,6 +58,22 @@ async function extractConfigsToImportedProfile(
...
+    downloadedMods: ThunderstoreCombo[],
+    profile: ImmutableProfile,
+    store: Store<any>
+): Promise<void> {
+    await ProfileModList.requestLock(async () => {
+        try {
+            const modList: ManifestV2[] = await installModsToProfile(downloadedMods, profile);
+            await store.dispatch('profile/updateModList', modList);
+            throwForR2Error(await ConflictManagementProvider.instance.resolveConflicts(modList, profile));
+        } catch (e: any) {
```
anttimaki, line 71: Minor: IDK why I didn't notice this earlier, probably due to the long comment above. Anyways, We now catch the exception to throw it as-is again, so basically the try-catch does nothing here. I'll leave it up to you if/when it should be fixed.

anttimaki (approved):

anttimaki (approved):

VilppeRiskidev:

PR #1596, opened 2024-12-18
Clean up the download state management Vuex store and it's usages

VilppeRiskidev:
- Remove assignId field
- Remove downloadObject field
- Rename some fields and functions to be more descriptive
- assignId is no longer passed unnecessarily when calling vuex store functions

### DIFF
--- a/src/components/views/DownloadModModal.vue
+++ b/src/components/views/DownloadModModal.vue
@@ -60,7 +60,6 @@ import ConflictManagementProvider from '../../providers/generic/installing/Confl
 import ProfileInstallerProvider from '../../providers/ror2/installing/ProfileInstallerProvider';
 import ThunderstoreDownloaderProvider from '../../providers/ror2/downloading/ThunderstoreDownloaderProvider';
 import ProfileModList from '../../r2mm/mods/ProfileModList';
-import { DownloadProgress } from "../../store/modules/DownloadModule";
 
     @Component({
         components: {
@@ -84,45 +83,31 @@ import { DownloadProgress } from "../../store/modules/DownloadModule";
             ignoreCache: boolean,
             store: Store<any>
         ): Promise<void> {
-            return new Promise((resolve, reject) => {
+            return new Promise(async (resolve, reject) => {
                 const tsMod = combo.getMod();
                 const tsVersion = combo.getVersion();
 
-                const currentAssignId = store.getters["download/currentDownload"]
-                    ? store.getters["download/currentDownload"].assignId + 1
-                    : 0;
-
-                const progressObject = {
-                    progress: 0,
-                    initialMods: [`${tsMod.getName()} (${tsVersion.getVersionNumber().toString()})`],
-                    modName: '',
-                    assignId: currentAssignId,
-                    failed: false,
-                };
-
-                store.commit('download/addDownload', progressObject);
+                const assignId = await store.dispatch(
+                    'download/addDownload',
+                    [`${tsMod.getName()} (${tsVersion.getVersionNumber().toString()})`]
+                );
 
                 setTimeout(() => {
                     ThunderstoreDownloaderProvider.instance.download(profile.asImmutableProfile(), tsMod, tsVersion, ignoreCache, (progress: number, modName: string, status: number, err: R2Error | null) => {
-                        if (status === StatusEnum.FAILURE) {
-                            if (err !== null) {
-                                const existing = store.state.download.allDownloads.find((dlObj: DownloadProgress) => {
-                                    return dlObj.assignId === currentAssignId;
-                                });
-                                existing.failed = true;
-                                store.commit('download/updateDownload', existing);
-                                DownloadModModal.addSolutionsToError(err);
-                                return reject(err);
+                        try {
+                            if (status === StatusEnum.FAILURE) {
+                                store.commit('download/updateDownload', {assignId, failed: true});
+                                if (err !== null) {
+                                    DownloadModModal.addSolutionsToError(err);
+                                    return reject(err);
+                                }
+                            } else if (status === StatusEnum.PENDING) {
+                                store.commit('download/updateDownload', {assignId, progress, modName});
                             }
-                        } else if (status === StatusEnum.PENDING) {
-                            const obj = {
-                                progress: progress,
-                                initialMods: [`${tsMod.getName()} (${tsVersion.getVersionNumber().toString()})`],
-                                modName: modName,
-                                assignId: currentAssignId,
-                                failed: false,
+                        } catch (e) {
+                            if (e instanceof R2Error) {
+                                store.commit('error/handleError', e);
                             }
-                            store.commit('download/updateDownload', obj);
                         }
                     }, async (downloadedMods: ThunderstoreCombo[]) => {
                         ProfileModList.requestLock(async () => {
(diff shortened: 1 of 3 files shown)


src/store/modules/DownloadModule.ts
```diff
@@ -23,34 +23,39 @@ export const DownloadModule = {
...
-            const newVersions = [...state.allVersions];
-            newVersions[params.assignId] = params.downloadVersion;
-            state.allVersions = newVersions;
+        updateDownload(state: State, downloadObject: DownloadProgress) {
+            const newVersions = [...state.allDownloads];
+            const index = newVersions.findIndex((oldDownloadObject: DownloadProgress) => {
+                return oldDownloadObject.assignId === downloadObject.assignId;
+            });
+            if (index === -1) {
+                throw new R2Error(
```
anttimaki, line 51: Nitpick: Did you test this? Is the error message shown to user in the error modal like it's written here, or does it get overwritten somewhere along the way? If this error is shown to the user, I think the middle line might be too technical while offering the user no actionable information, nor provides anything for us to debug the issue. Something like "DownloadProgress with id ${oldDownloadObject.assignId} was not found" would at least help in debugging if the value turns out to be undefined or something?

anttimaki (changes requested):
The changes match mostly what was planned, so in that sense this could be approved, but I think the suggested improvements are worthy of implementing right away.

One exception is that you didn't implement a getter for "next assign id" as was planned, nor was the plan augmented. Why's that?

Also generally speaking, while it's good to stick to the plan, I consider going a bit further than the plan if further improvement changes are discovered during implementation. Scope creep needs to be kept in mind though. I guess what I'm trying to say is to beware of the mentality where the plan is executed blindly, as it doesn't necessarily lead to best outcome.

VilppeRiskidev:
Improved based on the comments and rebased

src/components/views/DownloadModModal.vue
```diff
@@ -91,52 +83,31 @@ interface DownloadProgress {
...
                             }
-                        } else if (status === StatusEnum.PENDING) {
-                            const obj = {
-                                progress: progress,
-                                initialMods: [`${tsMod.getName()} (${tsVersion.getVersionNumber().toString()})`],
-                                modName: modName,
-                                assignId: currentAssignId,
-                                failed: false,
+                        } catch (e) {
+                            if (e instanceof R2Error) {
```
anttimaki, line 108: This should handle all caught errors regardless of the type, and in this specific method we should probably use `return reject(e)` as that's used in other places where errors are handled?

src/components/views/DownloadModModal.vue
```diff
@@ -167,112 +138,64 @@ interface DownloadProgress {
...
-                    const obj = {
-                        progress: progress,
-                        modName: modName,
-                        initialMods: modsWithUpdates.map(value => `${value.getMod().getName()} (${value.getVersion().getVersionNumber().toString()})`),
-                        assignId: currentAssignId,
-                        failed: false,
+                        this.$store.commit('download/updateDownload', {assignId, failed: true});
+                        if (err !== null) {
+                            DownloadModModal.addSolutionsToError(err);
+                            this.$store.commit('error/handleError', err);
```
anttimaki, line 154: Nitpick: as a personal preference, we could throw `err` here as we have the try-catch in place anyway. That way we don't have to duplicate the error handling. Applies to `downloadHandler` as well.

src/components/views/DownloadModModal.vue
```diff
@@ -167,112 +138,64 @@ interface DownloadProgress {
...
+                        }
+                    } else if (status === StatusEnum.PENDING) {
+                        this.$store.commit('download/updateDownload', {assignId, progress, modName});
                     }
-                    if (this.$store.state.download.downloadObject!.assignId === currentAssignId) {
-                        this.$store.commit('download/setDownloadObject', obj);
+                } catch (e) {
+                    if (e instanceof R2Error) {
```
anttimaki, line 160: Here too all errors should be handled. Applies to `downloadHandler` as well.

src/store/modules/DownloadModule.ts
```diff
@@ -23,34 +29,49 @@ export const DownloadModule = {
...
-        pushDownloadObjectToAllVersions(state: State, params: { assignId: number, downloadObject: DownloadProgress }) {
-            state.allVersions = [...state.allVersions, [params.assignId, params.downloadObject]];
-        },
-        updateDownloadObject(state: State, params: { assignId: number, downloadVersion: [number, DownloadProgress]}) {
-            const newVersions = [...state.allVersions];
-            newVersions[params.assignId] = params.downloadVersion;
-            state.allVersions = newVersions;
+        updateDownload(state: State, update: UpdateObject) {
+            const newDownloads = [...state.allDownloads];
+            if (newDownloads[update.assignId].assignId !== update.assignId) {
```
anttimaki, line 65: How did you test this? I ran two test cases:

1. The call site uses hard coded `assignId: 0` for the updates. The first download works as expected. Consequent downloads remains at 0%, while in the DownloadMonitor I can see that the first download's progress bar is being updated. No error is thrown as the first downloads assignId is in fact still 0.

2. The call site uses hard coded `assignId: 9999` for the updates. `newDownloads[update.assignId]` returns `undefined`, so `newDownloads[update.assignId].assignId` throws an error. Since the call site currently only expects `R2Errors` the error silently ignored. Even if the call site shows the error modal for all error types, the download actually continues in the background.

So I guess the best approach would be to:

- Use the `if (index === -1)` check from my earlier example
- If the `assignId` is unknown AND the update object indicates download failure, throw the error to show the error modal
- If the `assignId` is unknown AND the update object indicates non-error states, just ignore the error (perhaps `console.error it)`. The progress on the modal won't update, but the modal itself should close once the installation finishes
  - Or you could get fancy and show a text that says the progress is unknown but the installation is still likely in progress instead of the progress bar on the progress modal

Other approaches might work too so let me know if you have any ideas. And remember to test my ideas when implementing them as this short of stuff happens.

I consider this issue critical.

src/store/modules/DownloadModule.ts
```diff
@@ -23,34 +29,49 @@ export const DownloadModule = {
...
+                throw new R2Error(
+                    'Failed to update download status.',
+                    `DownloadProgress with id of ${update.assignId} didn't have the correct assignId.`,
+                    'Try initiating the download again or restarting the app.'
+                );
+            } else {
```
anttimaki, line 71: Nitpick: personal preference, but I see the `else` block a waste of lines and indentation when the `if` block always throws.

anttimaki (changes requested):

src/pages/DownloadMonitor.vue
```diff
@@ -42,10 +42,12 @@
 
 <script lang="ts">
 
+import Timeout = NodeJS.Timeout;
```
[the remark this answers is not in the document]
anttimaki: @VilppeRiskidev you could as a separate task/subtask for this so it's not forgotten (if you haven't already).

src/store/modules/DownloadModule.ts
VilppeRiskidev: Ok this should now be ok, but this is the most unsure part of the PR.
at least now it doesn't stop downloads from continuing.
The `console.warn()` thing feels a bit spammy, but it shouldn't be happening in the normal use cases after all (or more like never).

src/components/views/DownloadModModal.vue
```diff
@@ -167,112 +136,60 @@ interface DownloadProgress {
...
+                        }
+                    } else if (status === StatusEnum.PENDING) {
+                        this.$store.commit('download/updateDownload', {assignId, progress, modName});
                     }
-                    this.$store.commit('download/updateDownloadObject', {
-                        assignId: assignIndex,
-                        downloadVersion: [currentAssignId, obj]
-                    });
+                } catch (e) {
+                    this.$store.commit('error/handleError', e);
```
anttimaki, line 158: Now that the "if is R2Error" block is removed per my earlier request, `e` is `unknown` type and shouldn't be passed to `handleError` as is (there's no complaints about it due to the poor type support with this Vuex version). Instead, use `R2Error.fromThrownValue(e)`

src/store/modules/DownloadModule.ts
```diff
@@ -23,34 +29,65 @@ export const DownloadModule = {
...
-            state.allVersions = newVersions;
+        updateDownload(state: State, update: UpdateObject) {
+            const newDownloads = [...state.allDownloads];
+            const index = newDownloads.findIndex((old: DownloadProgress) => {
+                return old.assignId === update.assignId;
+            });
+
+            if (index === -1) {
+                // The DownloadProgress by the ID from the update wasn't found at all.
+                const err = new R2Error(
```
anttimaki, line 71: ~Minor: when we consider the situation where `err` is thrown and consequently shown to user in error modal, the message might not be as informative as it could be. If the action the user triggered is "download FartMod", they're probably not that interested in the download status or assignIds. Perhaps the `name` of the error could be something like "Failed to download a mod" and the `message` something like "[mod name] could not be downloaded". The `solution` is fine as it is. To show the mod name we will have to change `UpdateObject.modName` to required attribute though.~

Scratch the above. What we haven't taken into consideration is that this error will always overshadow the error provided by DownloadProvider (the one we pass to `addSolutionsToError` helper). I think we don't want to do that, so we shouldn't throw anything here. This is based on an assumption that DownloadProvider will include an Error in the callback whenever the status is set as failure. I think it's a relatively safe assumption, so trying to sniff out whether it did or didn't would probably be a wasted effort.

Secondly, logging the error message as it's now in the cases when the download hasn't failed isn't ideal either, as in case someone checks the console, we don't want them to "Try initiating the download again" as then they will have multiple downloads of the same files progressing.

Therefore I think this whole if-block should just console.warn something simple like "Could not find DownloadProgress object with assignId x" and `return`.

Thoughts?

src/store/modules/DownloadModule.ts
```diff
@@ -23,34 +29,65 @@ export const DownloadModule = {
...
+                // Only throw if the download failed, otherwise just console.log() it.
+                if (update.failed) {
+                    throw err;
+                }
+                console.warn(err);
+                return;
+            }
+
+            if (index !== update.assignId) {
+                // Just as a bonus, log if the DownloadProgress by the ID was located at the corresponding index of the array.
```
anttimaki, line 85: Doesn't the comment say exactly the opposite of what the code actually does? Anyway I think the code is simple enough so no comment is necessary here.

src/store/modules/DownloadModule.ts
```diff
@@ -23,34 +29,65 @@ export const DownloadModule = {
...
+                console.log(`There was a mismatch between download update\'s assign ID (${update.assignId}) and the index it was found at (${index}).`);
+            }
+
+            newDownloads[update.assignId] = {...newDownloads[update.assignId], ...update};
```
anttimaki, line 89: Critical(?): isn't this a bug? Since we can't trust that `index === update.assignId`, shouldn't we use `index` to index the arrays here?

anttimaki (changes requested):
The commits in this PR should probably be squashed. We can't do that during merging while using the "stack all the approved changes in the same PR" approach, so you should do it manually and force push the changes.

src/store/modules/DownloadModule.ts
VilppeRiskidev: Oops, it was supposed to be "was not"... anyway, makes sense to remove the comment.

src/store/modules/DownloadModule.ts
VilppeRiskidev: Yeah makes sense 👍

src/components/views/DownloadModModal.vue
```diff
@@ -167,112 +136,60 @@ interface DownloadProgress {
...
+                            }
+                        } else if (status === StatusEnum.PENDING) {
+                            this.$store.commit('download/updateDownload', {assignId, progress, modName});
                         }
-                        this.$store.commit('download/updateDownloadObject', {
-                            assignId: assignIndex,
-                            downloadVersion: [currentAssignId, obj]
-                        });
+                    } catch (e) {
+                        this.$store.commit('error/handleError', e);
```
anttimaki, line 189: This instance should also use `R2Error.fromThrownValue(e)`.

VilppeRiskidev:

PR #1597, opened 2024-12-20
Unify mod list update functionality under Vuex

anttimaki:


src/store/modules/TsModsModule.ts
```diff
@@ -49,17 +50,19 @@ export const TsModsModule = {
...
-        isBackgroundUpdateInProgress: false,
+        /*** Mod list is updated from the API automatically and by user action */
+        isThunderstoreModListUpdateInProgress: false,
         /*** All mods available through API for the current active game */
         mods: [],
         /*** When was the mod list last refreshed from the API? */
-        modsLastUpdated: undefined
+        modsLastUpdated: undefined,
+        /*** Error shown on UI after mod list refresh fails */
+        thunderstoreModListUpdateError: '',
```
ebkr, line 63: I think it'd be worth changing this to be an undefined/Error object. We can do more with the Error itself such as present a modal than we can with just the error message.

src/store/modules/TsModsModule.ts
```diff
@@ -148,16 +145,95 @@ export const TsModsModule = {
         setExclusions(state, payload: string[]) {
             state.exclusions = payload;
         },
-        startBackgroundUpdate(state) {
-            state.isBackgroundUpdateInProgress = true;
+        setThunderstoreModListUpdateError(state, error: string|unknown) {
+            if (typeof error === 'string') {
+                state.thunderstoreModListUpdateError = error;
```
ebkr, line 150: Correct me if I'm wrong but this block should never be hit.

We invoke this in one place, and that's part of a try/catch block. We should always hit the `instanceof Error` case.

src/store/modules/TsModsModule.ts
```diff
@@ -148,16 +145,95 @@ export const TsModsModule = {
...
+            commit('startThunderstoreModListUpdate');
+
+            try {
+                commit('setThunderstoreModListUpdateStatus', 'Checking for mod list updates from Thunderstore...');
+                let packageListIndex: PackageListIndex;
+
+                try {
+                    packageListIndex = await dispatch('fetchPackageListIndex');
+                } catch {
+                    // "Retry failed after 5 attempts" isn't very helpful error message.
```
ebkr, line 187: Noted and agreed, worth removing the comment now as it doesn't benefit the codebase

src/store/modules/TsModsModule.ts
anttimaki: Recap of places where this is used:

- ModListUpdateBanner: currently shows error as "inline text". As per my reply in the earlier PR, can be changed to show the error in modal
- ImportProfileModal: next PR in this stack will also show the error as "inline test". IMO using modal isn't an option here, as we're already showing one modal. We can of course extract the text from the Error object here, so that's not a problem
- SettingsView:  currently shows error as "inline text". Can be changed to use modal if we so choose
- UtilityMixin: throws the error so using Error object makes things a bit easier here

So in summary, I've no strong opinion this matter, given that the conversion from string to Error to string is trivial. If error modals is the way to go I'd rather do that (as well the change to Error/undefined) at a separate PR. That leaves the ImportProfileModal as an open issue though.

src/store/modules/TsModsModule.ts
anttimaki: Comment removed.

PR #1598, opened 2024-12-20
Refactor ImportProfileModal to support mod list updates

anttimaki:


src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -317,32 +306,83 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
+                </p>
+
+                <p class="margin-top">{{ profileMods.unknown.join(", ") }}</p>
+
+                <p v-if="profileMods.known.length" class="margin-top">
+                    These packages will not be installed.
+                </p>
+                <p v-else class="margin-top">
+                    Ensure the profile is intended for the currently selected game.
+                </p>
```
ebkr, line 325: This is probably fairly useless as an indicator since the BepInEx/MelonLoader packages can be shared across games.

Would it be better UX if we have another modal step, and in that step we re-show the game name and ask if it's correct?

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -317,32 +306,83 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
+                    </span>
+
+                    <br />
+
+                    <span v-if="$store.state.tsMods.isThunderstoreModListUpdateInProgress">
+                        {{ $store.state.tsMods.thunderstoreModListUpdateStatus }}
+                    </span>
+                    <span v-else-if="$store.state.tsMods.thunderstoreModListUpdateError">
+                        Error updating the mod list:
+                        {{ $store.state.tsMods.thunderstoreModListUpdateError }}.
```
ebkr, line 341: Do we have a similar concern here as with one of the previous PRs? Should we instead give the user the option to view the error in a separate error modal? Do they actually care about an inline variant of the message?

Generally the error modal is a fixed place for "something went wrong, we're informing you on what it is".
The actual purpose of this part is to allow the user to try again.

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -317,32 +306,83 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
-                @click="onProfileReviewConfirmed();">
-                Import
-            </button>
+            <div class="container is-flex is-justify-content-space-between">
+                <button
+                    id="modal-no-mods-to-import"
+                    class="button is-danger"
+                    @click="closeModal"
+                >
+                    Cancel import
```
ebkr, line 358: I think `Cancel` alone is fine? Not overly fussed.

We don't usually offer "Cancel" actions though. It's expected that the user would just click off the modal. Might be nice to make it consistent in other places if we're going to start doing this.

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -27,15 +30,14 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
         | 'IMPORT_FILE'
-        | 'ADDING_PROFILE'
+        | 'IMPORT_CODE'
         | 'REVIEW_IMPORT'
-        | 'NO_PACKAGES_IN_IMPORT'
+        | 'IMPORT_UPDATE_SELECTION'
+        | 'ADDING_PROFILE'
         | 'PROFILE_IS_BEING_IMPORTED'
-    = 'IMPORT_UPDATE_SELECTION';
+    = 'FILE_CODE_SELECTION';
```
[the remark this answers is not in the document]
anttimaki: My idea was to fail early. As the step where the import process is most likely to "fail" is the preview step, I wanted to take the user there without having to e.g. create a new profile folder first.

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -56,24 +58,22 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
-            .filter(mod => !!mod)
-    }
-
-    // Fired when user selects whether to import a new profile or update existing one.
-    onCreateOrUpdateSelect(mode: 'CREATE' | 'UPDATE') {
-        this.importUpdateSelection = mode;
-
-        if (mode === 'UPDATE') {
-            this.targetProfileName = this.activeProfileName;
+    get profileMods(): {known: ThunderstoreMod[], unknown: string[]} {
```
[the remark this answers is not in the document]
anttimaki: I'm not sure which files you refer to? The type isn't used elsewhere and doesn't need to be redefined even within the modal file.

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -317,32 +306,83 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
             <h2 class="modal-title">Packages to be installed</h2>
         </template>
         <template v-slot:body>
-            <OnlineModList :paged-mod-list="profileToOnlineMods" :read-only="true" />
+            <OnlineModList :paged-mod-list="profileMods.known" :read-only="true" />
+            <div v-if="!profileMods.known.length || profileMods.unknown.length" class="notification is-warning margin-top">
```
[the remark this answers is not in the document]
anttimaki: 🤷🏻

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -317,32 +306,83 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
+                <p v-if="!profileMods.known.length">
+                    None of the packages in the profile were found on Thunderstore:
+                </p>
+                <p v-else>
+                    Some of the packages in the profile were not found on Thunderstore:
+                </p>
+
+                <p class="margin-top">{{ profileMods.unknown.join(", ") }}</p>
```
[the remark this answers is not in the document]
anttimaki: 🤷🏻

src/components/profiles-modals/ImportProfileModal.vue
anttimaki: The way I see it is that the user might want to proceed with the import even if the update fails. Opening the error modal would prevent them from continuing where they left off, which to me is more awkward than the inline error text. I'm not completely against changing this though, if it's seen as an improvement.

src/components/profiles-modals/ImportProfileModal.vue
anttimaki: 🤷🏻

anttimaki:
🤷🏻 = Changes implemented in #1622

PR #1610, opened 2025-01-08
Separate update all

VilppeRiskidev:


src/components/views/UpdateAllInstalledModsModal.vue
```diff
@@ -0,0 +1,73 @@
...
+import R2Error from "../../model/errors/R2Error";
+import ThunderstoreDownloaderProvider from "../../providers/ror2/downloading/ThunderstoreDownloaderProvider";
+
+@Component({
+    components: {DownloadModVersionSelectModal, ModalCard}
+})
+export default class UpdateAllInstalledModsModal extends mixins(DownloadMixin)  {
+
+
+    async downloadLatest() {
```
anttimaki, line 42: Nitpick: the naming makes little sense in the standalone component. It could be `updateAll` or `updateMods` oslt.

anttimaki (approved):
> updating the downloadingMod doesn't properly propagate the update to all components

As I [mentioned during an earlier modal refactoring](https://discord.com/channels/877914952902516776/1312062521795936326/1312066082890059876), it's not supposed to. That's not how mixins work.

> One thing I thought about is if this should be in ModalsModule instead of DownloadModule

Currently I would choose the first, as that's already where `openDownloadModModal` is. Eventually I hope we can separate the three modals completely, so the two don't have to be subcomponents of the first one, so we can revisit the question at that point. In the meanwhile it's okay to keep them where they are.

Approving this as it works and is a good step on the refactoring path. I'd like to point out that moving `downloadLatest` to separate component makes it less obvious that it's quite similar to `downloadHandler`. At least the progress/completed callbacks could probably be moved into DownloadMixin in the next PR. It'll probably make it easier to reason how to rid ourselves from the callbacks altogether.

anttimaki (approved):
👍🏻

VilppeRiskidev:

PR #1611, opened 2025-01-08
Fix downloading a dependency for an already installed mod

anttimaki:
The implementation tried to read the mod list from a store location that hasn't existed in a while.


VilppeRiskidev (approved):
LOL nice one, LGTM 👍

PR #1259, opened 2024-03-10
Add MS Store version of Bomb Rush Cyberfunk

cspotcode:
I don't actually own the MS Store version of the game so I can't test this.

Additionally, the game is *not* on GamePass; it's on the Microsoft Store.  Does that affect the configuration?

I asked someone who has the MS Store version to get some info about the installed package.

```
PS C:\Users\usernameredacted> Get-AppxPackage | ? {$_.name -like '*Cyberfunk*'}

Name              : TeamReptile.BombRushCyberfunk
Publisher         : CN=9E62FE55-F651-4B2F-939E-E0C77E2B64FF
Architecture      : X64
ResourceId        :
Version           : 1.0.19991.0
PackageFullName   : TeamReptile.BombRushCyberfunk_1.0.19991.0_x64__e5dfpr1e2tmgw
InstallLocation   : C:\Program Files\WindowsApps\TeamReptile.BombRushCyberfunk_1.0.19991.0_x64__e5dfpr1e2tmgw
IsFramework       : False
PackageFamilyName : TeamReptile.BombRushCyberfunk_e5dfpr1e2tmgw
PublisherId       : e5dfpr1e2tmgw
IsResourcePackage : False
IsBundle          : False
IsDevelopmentMode : False
NonRemovable      : False
Dependencies      : {Microsoft.VCLibs.140.00.UWPDesktop_14.0.33519.0_x64__8wekyb3d8bbwe}
IsPartiallyStaged : False
SignatureKind     : Store
Status            : Ok
```

PR #1616, opened 2025-01-11
Remove MSStore from Bomb Rush Cyberfunk

cspotcode:
Effectively reverts #1259

At the time, nobody with the technical ability to verify if it would work had the MS Store copy, so we enabled it.  Since then, someone was able to get the MS Store version and verify that it is IL2CPP, so is incompatible with all our modding.  So now we know the MSStore / GamePass option should be removed from r2modman.


ebkr:
Apologies this got left open - This logic has been moved to https://github.com/thunderstore-io/ecosystem-schema for a while.

If it hasn't happened already, could you open a pull request there?

PR #1617, opened 2025-01-14
Download completed callback improvement

VilppeRiskidev:


VilppeRiskidev:
See also: https://github.com/thunderstore-io/thunderstore-mod-manager/pull/297

VilppeRiskidev:
Needs some testing from my end.
Also still need to move the callback(s) to the Mixin.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -1,3 +1,4 @@
+import ManifestV2 from "src/model/ManifestV2";
```
anttimaki, line 1: "src" in import

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -86,35 +87,23 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
                            ignoreCache: boolean,
                            callback: (progress: number, modName: string, status: number, err: R2Error | null) => void,
                            completedCallback: (modList: ThunderstoreCombo[]) => void) {
+
+        const modList = await ProfileModList.getModList(profile);
```
anttimaki, line 93: Good idea to fail early :+1:

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -86,35 +87,23 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
...
+        if (modList instanceof R2Error) {
+            return callback(0, mod.getName(), StatusEnum.FAILURE, modList);
+        }
+
         let dependencies: ThunderstoreCombo[] = [];
         await this.buildDependencySet(modVersion, dependencies, DependencySetBuilderMode.USE_EXACT_VERSION);
         this.sortDependencyOrder(dependencies);
         const combo = new ThunderstoreCombo();
```
anttimaki, line 102: Possible idea for future PR or another commit: this method could take a combo object as the argument instead of `mod` & `modVersion`. All the other methods in the provider already deal with combos, so that would be consistent.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -86,35 +87,23 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
...
-        // If not, get latest version of dependencies.
-        let isModpack = combo.getMod().getCategories().find(value => value === "Modpacks") !== undefined;
-        if (!isModpack) {
-            // If not modpack, get latest
-            dependencies = [];
-            await this.buildDependencySet(modVersion, dependencies, DependencySetBuilderMode.USE_LATEST_VERSION);
-            this.sortDependencyOrder(dependencies);
-            // #270: Remove already-installed dependencies to prevent updating.
-            dependencies = dependencies.filter(dep => modList.find(installed => installed.getName() === dep.getMod().getFullName()) === undefined);
-            downloadableDependencySize = this.calculateInitialDownloadSize(dependencies);
```
anttimaki, line 115: ~Seems you left this line out from your new method, which I think would normally be a bug. But, `calculateInitialDownloadSize` is stupid and its docstring is a lie. IMO we should yeet `downloadableDependencySize` and just refer to `dependencies.length`~ ah well it was addressed in the next commit

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -148,6 +137,19 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
         })
     }
 
+    // If combo is a modpack, use the modpack's dependency versions. If it isn't, get the latest versions.
```
anttimaki, line 144: Extracting this to a separate function is a good idea, but the I think the implementation could be better:

1. It should be private as there's no need to call it from outside
2. I don't think `modVersion` argument is needed since it's already available in `combo`?
3. I think `modList` isn't needed as a parameter either since aside from returning early in case of error it's needed only inside this method. However, it's probably better to wait this change until we've changed the `.download()` method enough so that we can throw the error instead of returning it
4. The method takes `dependencies` as an arg, maybe mutates it, and returns it. I find this to be an anti-pattern, since array arguments are passed as references. In other words whatever we do to `dependencies` here affects the `dependencies` in `download` whether we return the value or not. Ideally we wouldn't want to mutate the `dependencies` of `download` here as that makes the code hard to track. Instead, ideally, this should return a separate array that can then be merged to other `dependencies` in `download`. However, the next point renders this point moot
5. The implementation is still inefficient. The old implementation first gathers the dependencies with `USE_EXACT_VERSION`, *then* checks if the mod is a modpack, and if it is, does it all over again using `USE_LATEST_VERSION`. Why not check for modpack first, and then call `buildDependencySet` with a correct argument, and do all that follows only once, in this method. Then this method can just return the array of dependencies and we don't have to worry about the stuff in the point 4 above. And we can do the reading of `modList` inside this method and return early in case of error as decsribed in point 3.

src/components/views/DownloadModModal.vue
```diff
@@ -133,10 +132,9 @@ import ProfileModList from '../../r2mm/mods/ProfileModList';
                     } catch (e) {
                         this.$store.commit('error/handleError', R2Error.fromThrownValue(e));
                     }
-                }, async (downloadedMods) => {
-                    await this.downloadCompletedCallback(downloadedMods);
-                    this.$store.commit('download/setIsModProgressModalOpen', false);
                 });
+                this.$store.commit('download/setIsModProgressModalOpen', false);
```
anttimaki, line 136: Why switch the order? Now the modal is closed before install finishes, and the user might start the game modded before all mods are copied into profile folder.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -185,7 +189,7 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
 
             try {
                 const response = await this._downloadCombo(combo, singleModProgressCallback);
-                await this._saveDownloadResponse(response, combo, singleModProgressCallback);
+                await this._saveDownloadResponse(response, combo, singleModProgressCallback); //TODO: track installation separately
```
anttimaki, line 192: Not sure what you mean by this comment? `downloadImportedMods` already has this covered.

src/components/views/DownloadModModal.vue
VilppeRiskidev: oops, wasn't supposed to.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
VilppeRiskidev: I did this change now, seemed like a simple change that removes extra lines.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
VilppeRiskidev: Should be better now

src/components/views/DownloadModModal.vue
```diff
@@ -60,31 +60,30 @@ import ProfileModList from '../../r2mm/mods/ProfileModList';
...
-                setTimeout(() => {
-                    ThunderstoreDownloaderProvider.instance.download(profile.asImmutableProfile(), tsMod, tsVersion, ignoreCache, (progress: number, modName: string, status: number, err: R2Error | null) => {
-                        try {
-                            if (status === StatusEnum.FAILURE) {
-                                store.commit('download/updateDownload', {assignId, failed: true});
-                                if (err !== null) {
-                                    DownloadMixin.addSolutionsToError(err);
-                                    return reject(err);
+                setTimeout(async () => {
+                    try {
```
anttimaki, line 70: Generally speaking, we'd either want to have the try-blocks cover as little code as possible to have error specific error handling. Or we'd want to have one try-block for the whole thing if we handle all the errors the same way.

IDK exactly what is the optimal way to handle this case though.

- My earlier comment referred to catch errors thrown from the "core" of `.download()`, i.e. not the callback. But now this wraps both `.download()` and the install part. The latter already has its own error handling.
- It might be clearer if the `.download()` had it's own try-catch, to make it more clearer a distinct step from the installation?
- The error in the added catch block is shown on the error modal while the inner error handling `reject`s - I don't think it should do that.

That being said, I'd recommend changing the profile list loading part in `.download` to always throw and then seeing what's the simplest way we can handle the errors here.

src/components/views/DownloadModModal.vue
```diff
@@ -117,26 +118,34 @@ import ProfileModList from '../../r2mm/mods/ProfileModList';
...
-                    } catch (e) {
-                        this.$store.commit('error/handleError', R2Error.fromThrownValue(e));
-                    }
-                }, async (downloadedMods) => {
+                    });
                     await this.downloadCompletedCallback(downloadedMods);
                     this.$store.commit('download/setIsModProgressModalOpen', false);
-                });
+                } catch (e) {
+                    this.$store.commit('error/handleError', R2Error.fromThrownValue(e));
```
anttimaki, line 147: Should this close the progress modal before opening the error modal, and/or do some other cleanup?

src/providers/ror2/downloading/ThunderstoreDownloaderProvider.ts
```diff
@@ -49,17 +49,17 @@ export default abstract class ThunderstoreDownloaderProvider {
...
+     * @param profile                   The profile the mod is downloaded for (needed to prevent dependencies from updating existing mods).
+     * @param combo                     The combo to be downloaded.
+     * @param ignoreCache               Download mod even if it already exists in the cache.
+     * @param totalProgressCallback     Callback to show the combined state of all the downloads.
      */
-    public abstract download(profile: ImmutableProfile, mod: ThunderstoreMod, modVersion: ThunderstoreVersion,
-                    ignoreCache: boolean,
-                    callback: (progress: number, modName: string, status: number, err: R2Error | null) => void,
-                    completedCallback: (modList: ThunderstoreCombo[]) => void): void;
+    public abstract download(
```
anttimaki, line 57: Remember to clean up unused imports

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -82,70 +82,67 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
...
-        if (!isModpack) {
-            // If not modpack, get latest
-            dependencies = [];
-            await this.buildDependencySet(modVersion, dependencies, DependencySetBuilderMode.USE_LATEST_VERSION);
-            this.sortDependencyOrder(dependencies);
-            // #270: Remove already-installed dependencies to prevent updating.
-            dependencies = dependencies.filter(dep => modList.find(installed => installed.getName() === dep.getMod().getFullName()) === undefined);
-            downloadableDependencySize = this.calculateInitialDownloadSize(dependencies);
-        }
+        let dependencies = await this.getDependenciesWithCorrectVersions(combo, modList);
```
anttimaki, line 98: Nitpick: use `const` instead.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -82,70 +82,67 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
...
+            } catch(e) {
+                throw R2Error.fromThrownValue(e, `Failed to download mod ${combo.getVersion().getFullName()}`);
+            }
+        }
+        return allModsToDownload;
+    }
+
+    // If combo is a modpack, use the modpack's dependency versions. If it isn't, get the latest versions.
+    private async getDependenciesWithCorrectVersions(combo: ThunderstoreCombo, modList: ManifestV2[]) {
+        let dependencies: ThunderstoreCombo[] = [];
```
anttimaki, line 138: Nitpick: probably could use `const` here as well.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -82,70 +82,67 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
...
+        const isModpack = combo.getMod().getCategories().find(value => value === "Modpacks") !== undefined;
```
anttimaki, line 139: Nitpick: `some` instead of `find` would be shorter and IMO clearer.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -82,70 +82,67 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
...
+        const versionMode = isModpack ? DependencySetBuilderMode.USE_EXACT_VERSION : DependencySetBuilderMode.USE_LATEST_VERSION;
+
+        await this.buildDependencySet(combo.getVersion(), dependencies, versionMode);
+        this.sortDependencyOrder(dependencies);
+        // #270: Remove already-installed dependencies to prevent updating.
+        return dependencies.filter(dep => modList.find(installed => installed.getName() === dep.getMod().getFullName()) === undefined);
```
anttimaki, line 145: Nitpick: `some` instead of `find` would be shorter and IMO clearer.

anttimaki (changes requested):

src/components/views/DownloadModModal.vue
VilppeRiskidev: Yes, closing the modal makes sense.
_Maybe_ it would be wise to move this modal open state tracking to the `ModalsModule` and close all modals (or at least the progress modal) when an error modal is opened...

src/components/views/DownloadModModal.vue
VilppeRiskidev: should be better now 🤔

anttimaki: I'm not sure what was changed here? It seems it's still a mixture of small, action specific try-catches, and an outer -wrap-all try-catch? And the wrap-all still handles the error instead of rejecting it like the other places?

I think the easiest solution would be to just drop the wrap-all one altogether. The callsite (in Manager.vue) will then handle the error as it sees fit. But it's still a bit hard to reason about the flow of the code just by looking at it, so this should be tested to see if it actually works.

src/components/views/DownloadModModal.vue
anttimaki: We should probably `return` from the `catch` to avoid the rest of the code getting executed after failure.

anttimaki (changes requested):
This passes the "review" state, but on manual testing, it seems that errors don't properly set the state of the download tracker. I.e:

1. Start download
2. Before it finishes, use dev tools to go offline
3. Error modal is shown
4. In the DownloadMonitor view, the download is still shown as in progress

I think the catch-block in `downloadHandler` should call Vuex's `download/updateDownload` with failure state, but do test this and other error cases thoroughly yourself too.

VilppeRiskidev:
> This passes the "review" state, but on manual testing, it seems that errors don't properly set the state of the download tracker. I.e:
> 
> 1. Start download
> 2. Before it finishes, use dev tools to go offline
> 3. Error modal is shown
> 4. In the DownloadMonitor view, the download is still shown as in progress
> 
> I think the catch-block in `downloadHandler` should call Vuex's `download/updateDownload` with failure state, but do test this and other error cases thoroughly yourself too.

Should be fixed now. The progress failed status is now updated when the download fails.

Displaying the error in DownloadMonitor will be improved in the future.

anttimaki (changes requested):
image

As discussed in yesterday's daily, this is not what we want. This should unambiguously show the download failed. Even if the improvements we discussed are done separately, this here is a bug that should've been caught in your testing.

anttimaki (changes requested):
image

I think it's worse now (this is a failed download). How are you testing this? Maybe there's a difference between errors?

VilppeRiskidev:
At least I can't get it to fail without it showing the correct error state in ModDownloadMonitor anymore.

src/components/mixins/DownloadMixin.vue
```diff
@@ -46,6 +52,35 @@ export default class DownloadMixin extends Vue {
         }
     }
 
+    downloadProgressCallback(assignId: number, progress: number, modName: string, status: number, err: R2Error | null) {
```
anttimaki, line 55: These now have some repetition I think is addressed in #1625

src/components/views/DownloadModModal.vue
```diff
@@ -60,49 +59,48 @@ import ProfileModList from '../../r2mm/mods/ProfileModList';
...
+                        for (const combo of downloadedMods) {
+                            try {
+                                await DownloadModModal.installModAfterDownload(profile, combo.getMod(), combo.getVersion());
+                            } catch (e) {
+                                return reject(
+                                    R2Error.fromThrownValue(e, `Failed to install mod [${combo.getMod().getFullName()}]`)
+                                );
+                            }
+                        }
+                        const modList = await ProfileModList.getModList(profile.asImmutableProfile());
```
anttimaki, line 96: As the following issues existed before this PR I think they're out of scope but please take a note that we should check if at least some of the can **and** should be addressed:

- Any error from `.getModList()` is silently ignored
- Depending on how `.resolveConflicts()` is used elsewhere, and how it's implemented internally it could make sense to not pass `modList` as an argument but read it internally (it seems only profile is required to do that). However, keeping `ProfileModList` and  the implemenation of `ConflictManagementProvider` loosely coupled might be the correct idea too
-  `.resolveConflicts()` is called separately for each installed mod in this loop, and it might be unnecessary if calling it only once in the end with the final `modList` yields the same results. However, if we want to fail early to keep the profile from corrupting somehow, the current approach might be the correct idea

src/pages/DownloadMonitor.vue
```diff
@@ -19,18 +19,21 @@
                         <div class="border-at-bottom pad pad--sides">
                             <div class="card is-shadowless">
                                 <p><strong>{{ downloadObject.initialMods.join(", ") }}</strong></p>
-                                <p v-if="downloadObject.progress < 100">Downloading: {{ downloadObject.modName }}</p>
-                                <p>{{Math.min(Math.floor(downloadObject.progress), 100)}}% complete</p>
-                                <Progress v-if="!downloadObject.failed"
+                                <p v-if="downloadObject.failed">Downloading failed</p>
+                                <p v-else-if="downloadObject.progress < 100">Downloading: {{ downloadObject.modName }}</p>
```
anttimaki, line 23: This doesn't update `modName` when dependencies are being downloaded but I think that is addressed in #1625

anttimaki (approved):
I'm approving this now but please ensure the newest comments are addressed in #1625 and on a separate task. Note that this makes merging this branch sort of conditional on having #1625 merge-ready.

src/components/mixins/DownloadMixin.vue
VilppeRiskidev: correct 👍

src/components/views/DownloadModModal.vue
VilppeRiskidev: I'll fix these in the in https://github.com/ebkr/r2modmanPlus/pull/1625

anttimaki: For future reference, all of these were in fact not fixed in #1625.

PR #1618, opened 2025-01-15
Fix version number stringification when updating all installed mods

anttimaki:


anttimaki:
FYI @VilppeRiskidev . Merging this now as it's needed for a TSMM release.

PR #1619, opened 2025-01-16
Improve "reset cached package list" feature

anttimaki:


VilppeRiskidev (approved):
Code looks good and doesn't seem to break anything 👌

PR #1625, opened 2025-01-23
Mod installation progress tracking

VilppeRiskidev:


src/components/mixins/DownloadMixin.vue
```diff
@@ -44,15 +44,15 @@ export default class DownloadMixin extends Vue {
         return this.$store.getters['profile/activeProfile'];
     }
 
-    async downloadCompletedCallback(downloadedMods: ThunderstoreCombo[]): Promise<void> {
+    async downloadCompletedCallback(downloadedMods: ThunderstoreCombo[], assignId: number): Promise<void> {
```
anttimaki, line 47: Idea for another task: renaming the `assingId` to something more descriptive. If I read just this file "assignId" doesn't really tell me what this number is. `downloadId` or `progressId` oslt might be more helpful.

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -203,18 +203,20 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
         const isUpdate = this.importUpdateSelection === 'UPDATE';
 
         try {
             const comboList = await ProfileUtils.exportModsToCombos(mods, game);
-            await ThunderstoreDownloaderProvider.instance.downloadImportedMods(comboList, ignoreCache, progressCallback);
-            await ProfileUtils.populateImportedProfile(comboList, mods, targetProfileName, isUpdate, zipPath, progressCallback);
+            await ThunderstoreDownloaderProvider.instance.downloadImportedMods(comboList, ignoreCache, (progress) => {
+                progressCallback(`Downloading mods: ${Math.floor(progress)}%`)
+            });
+            await ProfileUtils.populateImportedProfile(comboList, mods, targetProfileName, isUpdate, zipPath, (progress) => {
```
anttimaki, line 217: image

Needs more testing.

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -203,18 +203,20 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
+                progressCallback(`Copying mods to profile: ${progress}%`)
```
anttimaki, line 218: Should this also use `Math.floor`?

src/components/views/DownloadModModal.vue
```diff
@@ -4,11 +4,18 @@
...
+                    <h3 class='title'>Downloading and installing {{$store.getters['download/currentDownload'].modName}}</h3>
+                    <p>Downloading: {{Math.floor($store.getters['download/currentDownload'].downloadProgress)}}% complete</p>
                     <Progress
                         :max='100'
-                        :value="$store.getters['download/currentDownload'].progress"
+                        :value="$store.getters['download/currentDownload'].downloadProgress"
+                        :className="['is-dark']"
+                    />
+                    <p v-if="$store.getters['download/currentDownload'].installationProgress">Installation: {{$store.getters['download/currentDownload'].installationProgress}}% complete</p>
+                    <p v-else>Installation: waiting for download to finish</p>
```
anttimaki, line 15: Nitpick: using "Installing" instead of "Installation" on these two lines would be more in line with "Downloading" above.

src/pages/DownloadMonitor.vue
```diff
@@ -19,14 +19,28 @@
...
-                                <p v-if="downloadObject.progress < 100">Downloading: {{ downloadObject.modName }}</p>
-                                <p>{{Math.min(Math.floor(downloadObject.progress), 100)}}% complete</p>
-                                <Progress v-if="!downloadObject.failed"
-                                    :max='100'
-                                    :value='downloadObject.progress'
-                                    :className="['is-info']"
-                                />
-                                <Progress v-else-if="downloadObject.failed"
+
+                                <div v-if="downloadObject.downloadProgress < 100">
```
anttimaki, line 23: Now the download monitor does the annoying thing where progress bar resets to zero when a next phase starts. (This isn't even my final form.)

I was wondering if showing two progress bars side by side would work in this view?

src/pages/DownloadMonitor.vue
```diff
@@ -19,14 +19,28 @@
...
+                                    <p>Downloading: {{ downloadObject.modName }}</p>
+                                    <p>{{(downloadObject.failed ? "Failed at " : "") + Math.min(Math.floor(downloadObject.downloadProgress), 100)}}% complete</p>
+                                    <Progress v-if="!downloadObject.failed"
+                                        :max='100'
+                                        :value='downloadObject.downloadProgress'
+                                        :className="['is-info']"
+                                    />
+                                </div>
+
+                                <div v-else>
```
anttimaki, line 33: image

State fails when installation throws an error. It's worth checking if errors thrown from other steps as well might cause issues.

image

Installing the same mod twice causes the state go awry in some fashion. I know it's a niche case but if there's an issue like this it might happen in other scenarios as well so better to figure out what's going on.

src/store/modules/DownloadModule.ts
```diff
@@ -7,7 +7,8 @@ interface DownloadProgress {
     assignId: number;
     initialMods: string[];
     modName: string;
-    progress: number;
+    downloadProgress: number;
+    installationProgress: number;
```
anttimaki, line 11: Nitpick: "installProgress" would be shorter and match "download" better methinks.

src/utils/ProfileUtils.ts
```diff
@@ -54,17 +54,20 @@ async function extractConfigsToImportedProfile(
...
-        progressCallback(`Copying configs to profile: ${progress}%`);
+        progressCallback(progress);
     }
 }
 
 export async function installModsAndResolveConflicts(
     downloadedMods: ThunderstoreCombo[],
     profile: ImmutableProfile,
-    store: Store<any>
+    store: Store<any>,
```
anttimaki, line 64: IIRC I've mentioned earlier that this helper maybe shouldn't be a part of ProfileUtils since the need to pass the Store implies we're making the coupling too tight and not getting the code reusability that's often nice for helper functions.

I think it's even more clear now that we have to pass the assignId here too. I think it's an implementation detail the helpers in this file shouldn't be concerned with. It would make sense to me to move installModsAndResolveConflicts to DownloadMixin.

src/utils/ProfileUtils.ts
```diff
@@ -200,8 +203,12 @@ export async function populateImportedProfile(
```
anttimaki, line 203: Nitpick: I lowkey don't like it how all `progressCallbacks` used to have the same function signature but now others take a number while this one still takes a string. Consistency would be nice.

anttimaki (changes requested):

src/utils/ProfileUtils.ts
VilppeRiskidev: Should be more consistent again. An extra parameter still had to be added to `installModsToProfile`'s `progressCallback`.

src/components/mixins/DownloadMixin.vue
VilppeRiskidev: Created a task for this.

src/components/mixins/DownloadMixin.vue
```diff
@@ -42,14 +47,59 @@ export default class DownloadMixin extends Vue {
...
+    downloadProgressCallback(assignId: number, downloadProgress: number, modName: string, status: number, err: R2Error | null) {
+        try {
+            if (status === StatusEnum.FAILURE) {
+                this.setIsModProgressModalOpen(false);
+                this.$store.commit('download/updateDownload', {assignId, failed: true});
+                if (err !== null) {
+                    DownloadMixin.addSolutionsToError(err);
+                    throw err;
+                }
+            } else if (status === StatusEnum.PENDING || status === StatusEnum.SUCCESS) {
```
anttimaki, line 68: Why does this differ from L83? I think these two implementations should be identical, unless there's a very good reason for them not to be.

~In fact, if they should be identical, I think this should by DRYed up by calling `DownloadMixin.downloadProgressCallback` in the try-catch here. (Also, is the try-catch required here? It might be, I can't recall.)~

In fact, I think the place where this is called should call the static method instead. The try-catch seems (please verify) unnecessary since the call site already has error handling. But do check if `|| status === StatusEnum.SUCCESS` should also be applied to the static method too.

src/components/mixins/DownloadMixin.vue
```diff
@@ -42,14 +47,59 @@ export default class DownloadMixin extends Vue {
...
+            }
+        } else if (status === StatusEnum.PENDING) {
+            store.commit('download/updateDownload', {assignId, downloadProgress, modName});
+        }
+    }
+
+    async installModsAndResolveConflicts(
+        downloadedMods: ThunderstoreCombo[],
+        profile: ImmutableProfile,
+        store: Store<any>,
```
anttimaki, line 91: No need to pass store as an argument to this method.

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -212,17 +212,17 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
     async importProfile(targetProfileName: string, mods: ExportMod[], zipPath: string) {
         this.activeStep = 'PROFILE_IS_BEING_IMPORTED';
         this.importPhaseDescription = 'Downloading mods: 0%';
-        const progressCallback = (progress: number|string) => typeof progress === "number"
```
anttimaki, line 215: IDK if these changes are really needed after the last round changes in other files?

src/pages/DownloadMonitor.vue
```diff
@@ -19,18 +19,45 @@
...
-                                    :value='downloadObject.progress'
-                                    :className="['is-info']"
-                                />
-                                <Progress v-else-if="downloadObject.failed"
-                                    :max='100'
-                                    :value='100'
-                                    :className="['is-danger']"
-                                />
+
+                                <div class="row" v-if="!downloadObject.failed">
```
anttimaki, line 23: Nitpick: I prefer the order of "If x ... else ..." instead of "if not x ... else"

src/pages/DownloadMonitor.vue
```diff
@@ -19,18 +19,45 @@
...
+
+                                    <div class="col">
+                                        <p v-if="downloadObject.downloadProgress < 100">Downloading: {{ downloadObject.modName }}</p>
+                                        <p v-else-if="downloadObject.installProgress < 100 && !downloadObject.failed">Downloading:</p>
```
anttimaki, line 27: 🐛 I think there's a bug somewhere, the `modName` never updates when downloading dependencies.

`&& !downloadObject.failed` is unnecessary as this `<div class="row">` is already conditional to it. Same thing a few rows later

src/pages/DownloadMonitor.vue
```diff
@@ -19,18 +19,45 @@
...
+                                        <p>{{Math.min(Math.floor(downloadObject.downloadProgress), 100)}}% complete</p>
+                                        <Progress
+                                            :max='100'
+                                            :value='downloadObject.downloadProgress'
+                                            :className="['is-info']"
+                                        />
+                                    </div>
+
+                                    <div v-if="downloadObject.installProgress < 100 && !downloadObject.failed" class="col">
+                                        <p v-if="downloadObject.installProgress < 100">Installing: {{ downloadObject.modName }}</p>
```
anttimaki, line 37: Do we even attempt to update the `.modName` during installation phase? If not, showing this is a bit silly?

src/pages/DownloadMonitor.vue
```diff
@@ -19,18 +19,45 @@
...
+                                            :max='100'
+                                            :value='downloadObject.installProgress'
+                                            :className="['is-info']"
+                                        />
+                                    </div>
+
+                                </div>
+
+                                <div v-else-if="downloadObject.failed">
+                                    <p>{{downloadObject.failed ? "Download failed" : `${Math.min(Math.floor(downloadObject.installProgress), 100)}% complete`}}</p>
```
anttimaki, line 50: This is only rendered if download has failed so the logic here is unnecessary.

src/pages/DownloadMonitor.vue
```diff
@@ -19,18 +19,45 @@
...
+                                    <Progress
+                                        :max='100'
+                                        :value='100'
+                                        :className="['is-danger']"
+                                    />
+                                </div>
+
+                                <div v-else>
```
anttimaki, line 58: Is this ever executed? `downloadObject.failed` is boolean, so one of the previous `if`s is bound to execute. Or it can be `undefined`, in which case `v-if="!downloadObject.failed"` will execute. And IMO it makes no sense to spefically check against undefined, as showing "status unknown" looks odd when we can't event show `downloadObject.initialMods` above goes assumably it's also `undefined`.

src/utils/ProfileUtils.ts
```diff
@@ -58,18 +58,6 @@ async function extractConfigsToImportedProfile(
     }
 }
 
-export async function installModsAndResolveConflicts(
```
anttimaki, line 61: Remember to clean up unnecessary imports.

src/utils/ProfileUtils.ts
```diff
@@ -78,8 +66,8 @@ export async function installModsAndResolveConflicts(
 export async function installModsToProfile(
     comboList: ThunderstoreCombo[],
     profile: ImmutableProfile,
-    disabledModsOverride?: string[],
-    progressCallback?: (status: string) => void
+    progressCallback: (status: string, progress: number) => void,
```
anttimaki, line 69: Should the second argument be optional?

Also IDK why this was changed to a required argument?

src/utils/ProfileUtils.ts
```diff
@@ -134,6 +122,7 @@ export async function installModsToProfile(
     }
 
     throwForR2Error(await ProfileModList.saveModList(profile, profileMods));
+    progressCallback('', 100);
```
anttimaki, line 125: Would be a lot safer to use `"Copying mods to profile: 100%"` instead of an empty string.

anttimaki (changes requested):

src/components/mixins/DownloadMixin.vue
VilppeRiskidev: Now uses the static function for both 👍

src/pages/DownloadMonitor.vue
VilppeRiskidev: good catch, this never worked to begin with, but should be fixed now. I think one reason for this was that there were two different variables called `combo` so it was assumed they were the same :D

src/pages/DownloadMonitor.vue
VilppeRiskidev: The whole logic of the changes to this UI component were a total mess, should be fixed now.

src/pages/DownloadMonitor.vue
VilppeRiskidev: This one made me laugh when I read what I had coded LOL

src/utils/ProfileUtils.ts
VilppeRiskidev: 1. Sure
2. No reaseon
✅ 👍

src/utils/ProfileUtils.ts
VilppeRiskidev: Sure, even if it's not used at this point, it is better for the future developer who might assume it having the correct status string value.

anttimaki (commented):
In addition to these there was some buggy behaviour shared in voice chat.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -99,7 +99,7 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
         const allModsToDownload = [...dependencies, combo];
 
         let downloadCount = 0;
-        const singleModProgressCallback = (progress: number, status: number, err: R2Error | null) => {
+        const singleModProgressCallback = (modName: string, progress: number, status: number, err: R2Error | null) => {
```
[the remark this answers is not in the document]
VilppeRiskidev: Makes sense 👍

src/pages/DownloadMonitor.vue
```diff
@@ -19,21 +19,42 @@
...
+                                        <Progress
+                                            :max='100'
+                                            :value='downloadObject.downloadProgress'
+                                            :className="['is-info']"
+                                        />
+                                    </div>
+
+                                    <div v-if="downloadObject.installProgress < 100" class="col">
+                                        <p>Installing:</p>
+                                        <p v-if="downloadObject.downloadProgress < 100">Installing: waiting for download to finish</p>
```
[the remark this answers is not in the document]
VilppeRiskidev: Now it tracks the modName in the installation too and the logic is updated.

VilppeRiskidev:
Buggy behaviour should be at least better now, I'll have to test a bit more to see if it still fails in some conditions.

VilppeRiskidev:
https://github.com/ebkr/r2modmanPlus/pull/1617/files#r1939046141 "Any error from .getModList() is silently ignored" has been fixed

VilppeRiskidev:
https://github.com/ebkr/r2modmanPlus/pull/1617/files#r1939046141 
- Depending on how `.resolveConflicts()` is used elsewhere, and how it's implemented internally it could make sense to not pass `modList` as an argument but read it internally (it seems only profile is required to do that). However, keeping `ProfileModList` and  the implemenation of `ConflictManagementProvider` loosely coupled might be the correct idea too
-  `.resolveConflicts()` is called separately for each installed mod in this loop, and it might be unnecessary if calling it only once in the end with the final `modList` yields the same results. However, if we want to fail early to keep the profile from corrupting somehow, the current approach might be the correct idea

These two seem a bit trickier, I think we shuld create a new ticket for these if it is something we want to do.

src/components/views/DownloadModModal.vue
```diff
@@ -4,11 +4,18 @@
             <div class="modal-background" @click="setIsModProgressModalOpen(false);"></div>
             <div class='modal-content'>
                 <div class='notification is-info'>
-                    <h3 class='title'>Downloading {{$store.getters['download/currentDownload'].modName}}</h3>
-                    <p>{{Math.floor($store.getters['download/currentDownload'].progress)}}% complete</p>
+                    <h3 class='title'>Downloading and installing {{$store.getters['download/currentDownload'].modName}}</h3>
```
anttimaki, line 7: I think having this this way was my earlier suggestion but now that the installation actually updates the modName this seems a bit awkward as installation starts to loop the same mod names again. This could maybe show only "Downloading" or "Installing" based on which phase is active?

src/components/views/DownloadModModal.vue
```diff
@@ -4,11 +4,18 @@
...
+                    <p>Downloading: {{Math.floor($store.getters['download/currentDownload'].downloadProgress)}}% complete</p>
                     <Progress
                         :max='100'
-                        :value="$store.getters['download/currentDownload'].progress"
+                        :value="$store.getters['download/currentDownload'].downloadProgress"
+                        :className="['is-dark']"
+                    />
+                    <p v-if="$store.getters['download/currentDownload'].installProgress">Installing: {{$store.getters['download/currentDownload'].installProgress}}% complete</p>
```
anttimaki, line 14: Should this use Math.floor as well?

src/pages/DownloadMonitor.vue
```diff
@@ -19,18 +19,50 @@
...
+                                        :className="['is-danger']"
+                                    />
+                                </div>
+
+                                <div v-else class="row">
+
+                                    <div class="col">
+                                        <p v-if="downloadObject.downloadProgress < 100">Downloading: {{ downloadObject.modName }}</p>
+                                        <p v-else>Downloading:</p>
+                                        <p>{{Math.min(Math.floor(downloadObject.downloadProgress), 100)}}% complete</p>
```
anttimaki, line 37: To match with "Download failed", this could also simply say "Download complete" on one row without the percentage, once both download and installation is done.

src/pages/DownloadMonitor.vue
```diff
@@ -19,18 +19,50 @@
...
+                                            :value='downloadObject.downloadProgress'
+                                            :className="['is-info']"
+                                        />
+                                    </div>
+
+                                    <div v-if="downloadObject.installProgress < 100" class="col">
+                                        <div v-if="downloadObject.downloadProgress < 100">
+                                            <p>Installing:</p>
+                                            <p v-if="downloadObject.downloadProgress < 100">Waiting for download to finish</p>
+                                            <Progress
```
anttimaki, line 49: Personally I'm not a fan of using the indeterminate progress bar here. I don't think it adds much value as the user can see from the other progress bar what's currently happening. Also when the UI slows down, it lags and annoys me 😆

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -98,36 +98,39 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
...
             } else if (status === StatusEnum.SUCCESS) {
-                totalProgress = this.generateProgressPercentage(100, downloadCount, allModsToDownload.length);
+                totalDownloadProgress = this.generateProgressPercentage(100, downloadCount, allModsToDownload.length);
                 downloadCount += 1;
             } else {
                 console.error(`Ignore unknown status code "${status}"`);
                 return;
             }
-            totalProgressCallback(totalProgress, combo.getMod().getName(), status, err);
+            totalProgressCallback(Math.round(totalDownloadProgress), modInProgressName, status, err);
```
anttimaki, line 118: I think other places use Math.floor, any reason to use .round instead here? (If changed, should be changed to TSMM sibling PR as well.)

src/utils/ProfileUtils.ts
```diff
@@ -89,10 +75,12 @@ export async function installModsToProfile(
     const installedVersions = profileMods.map((m) => m.getDependencyString());
     const disabledMods = disabledModsOverride || profileMods.filter((m) => !m.isEnabled()).map((m) => m.getName());
     let currentMod;
```
anttimaki, line 77: After the changes currentMod isn't really used and there's no point tracking the same thing twice, so might as well yeet it.

I do like the empty line before try-block though, so please don't yeet it.

anttimaki (commented):
I don't consider any of the comments alone a blocker, but given there's a few of them, consider if some should be implemented now.

Ideas for future tasks:
- Prevent background mod list updates while there are active downloads
- Allow clearing one/all completed downloads via DownloadMonitor
- Allow retrying failed downloads via DownloadMonitor
- Use the actual file size information in ThunderstoreVersion to give different weights to different size mods when determining the progress

src/r2mm/downloading/BetterThunderstoreDownloader.ts
VilppeRiskidev: The reason was that the `generateProgressPercentage()` sometimes returns like 99.9999999 instead of 100 (floating point number calculations are sometimes a tiny bit inaccurate) and therefore the download progress got stuck at 99%. Should the `generateProgressPercentage` actually only return integers (round the number before returning)?

src/pages/DownloadMonitor.vue
VilppeRiskidev: Sure, I don't have a strong opinion so let's go with that.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
VilppeRiskidev: If the change is OK, should/could the rounding of the progress percentage be removed from the UI components or is it a best practice to round just in case?

src/pages/DownloadMonitor.vue
anttimaki: I'm probably not the right person to judge whether the completed progress bar should be green or stay blue. But changing the failed download bar to green is mixed signals and should IMO be avoided.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
anttimaki: In hindsight this whole thing should probably be designed so that it's the Vuex module that receives updates of each mods progress and internally keeps tracks how many mods is completed out of how many mods total, and what's the process of currently downloading mod. And it should then be the one that provides a getter or some other mechanic for providing the total progress to UI (rounded in one place and in one consistent way). That would also allow us to show it as percentage for progress bars, or as "Downloading mod 5 out of 6" for other status indicators.

Anyway, I'd be careful in changing `generateProgressPercentage()` as it's called from elsewhere too. If you think this change improves the overall situation, feel free to go ahead with it, but do test it thoroughly (all functions that call `generateProgressPercentage()` and related UI parts).

src/pages/DownloadMonitor.vue
```diff
@@ -19,18 +19,58 @@
...
+                                <div v-if="downloadObject.failed">
+                                    <p>Download failed</p>
+                                    <Progress
+                                        :max='100'
+                                        :value='100'
+                                        :className="['is-success']"
+                                    />
+                                </div>
+
+                                <div v-else-if="downloadObject.downloadProgress === 100 && downloadObject.installProgress === 100">
```
anttimaki, line 32: Separate task: update the progress object to have status enum instead of single `failed` boolean flag. Having statuses like "downloading", "installing", "done", and "failed" makes places like this much more readable.

src/pages/DownloadMonitor.vue
```diff
@@ -19,18 +19,58 @@
...
+                                        <p>Installing:</p>
+                                        <p>Waiting for download to finish</p>
+                                        <Progress
+                                            :max='100'
+                                            :value='0'
+                                            :className="['is-info']"
+                                        />
+                                    </div>
+                                    <div v-else class="col">
+                                        <p>Installing: {{ downloadObject.modName }}</p>
```
anttimaki, line 64: Seems `modName` in installing phase shows "teamName-modName" while downloading part only shows the "modName". The latter is IMO cleaner.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -98,36 +98,39 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
...
+        const singleModProgressCallback = (downloadProgress: number, status: number, err: R2Error | null) => {
             if (status === StatusEnum.FAILURE) {
                 throw err;
             }
 
-            let totalProgress: number;
+            let totalDownloadProgress: number;
             if (status === StatusEnum.PENDING) {
-                totalProgress = this.generateProgressPercentage(progress, downloadCount, allModsToDownload.length);
+                totalDownloadProgress = this.generateProgressPercentage(downloadProgress, downloadCount, allModsToDownload.length);
```
anttimaki, line 110: Not having "download is 80%, extracting 20%" multiplier here like there's in `downloadImportedMods` leads to situation where downloading e.g. "TV Loader Anime Openings" shows download being completed and installation "waiting for download" for two minutes while the videos are extracted from the zip.

anttimaki (commented):
Bonus: it seems at some point the order of downloads has changed. Develop downloads the target mod first and then the dependencies, while this branch downloads dependencies first. IDK if it matters at all, but inadvertently changing feels like begging for problems.

anttimaki (approved):
I'm now happy with these changes, with exception of this comment not being addressed:

> Bonus: it seems at some point the order of downloads has changed. Develop downloads the target mod first and then the dependencies, while this branch downloads dependencies first. IDK if it matters at all, but inadvertently changing feels like begging for problems.

I think you should either change the order back; or get a confirmation from ebkr that changing the order doesn't matter.

It's probably easy to overlook the comments that are not assigned to any row, so I'll try to avoid them in future reviews even if the comment isn't really related to any specific row. That being said, read the non-line review comments anyway.

VilppeRiskidev:
Fixed the main mod vs. dependencies ordering thing 👍

src/pages/DownloadMonitor.vue
```diff
@@ -19,18 +19,58 @@
...
-                                    :value='100'
-                                    :className="['is-danger']"
-                                />
+
+                                <div v-if="downloadObject.failed">
+                                    <p>Download failed</p>
+                                    <Progress
+                                        :max='100'
+                                        :value='100'
+                                        :className="['is-success']"
```
anttimaki, line 28: Why is the failed state using `is-success`? Wasn't this fixed once already? Are there likely be other regressions with the latest changes? Why wasn't this detected when you tested the changes?

anttimaki (changes requested):

src/pages/DownloadMonitor.vue
VilppeRiskidev: 🤦‍♂️ Oops, my bad, should be all good now.

src/pages/DownloadMonitor.vue
```diff
@@ -19,18 +19,58 @@
...
-                                    :value='100'
-                                    :className="['is-danger']"
-                                />
+
+                                <div v-if="downloadObject.failed">
+                                    <p>Download failed</p>
+                                    <Progress
+                                        :max='100'
+                                        :value='100'
+                                        :className="['is-error']"
```
anttimaki, line 28: The diff says this used to be `is-danger`. As far as I can tell, the component doesn't support `is-error` class at all.

anttimaki (changes requested):
In addition my testing revealed some buggy behaviour that is hard to reproduce. We can discuss these in the daily. Until I understand what happened in those cases, I don't feel this is ready for release yet.

src/pages/DownloadMonitor.vue
VilppeRiskidev: I could've sworn I copied it back from the diff, but I guess not. 💩💩💩

anttimaki (approved):
Alright, looks finally good to me. Couldn't reproduce the problems from yesterday anymore so lets hope they were once in a blue moon sort of thing.

PR #1627, opened 2025-01-27
Remove invalid manifest locations from list before attempting to search for valid installation directories

ebkr:


src/r2mm/manager/win32/GameDirectoryResolver.ts
```diff
@@ -120,6 +120,15 @@ export default class GameDirectoryResolverImpl extends GameDirectoryResolverProv
                 null
             )
         }
+
+        // Filter to only locations that exist on the drive
+        for (const location of locations) {
+            if (!await fs.exists(location)) {
+                var index = locations.indexOf(location);
+                locations.splice(index, 1);
```
anttimaki, line 128: Seems to me removing items from the array that's being iterated over this way causes the iteration to [skip items](https://www.typescriptlang.org/play/?ts=5.7.3#code/G4QwTgBANg9gxiALgSxgOwM4QLwQNoAMANAIxEBMRAzEQCxECsRAbEQOxEAcRAnALoBuALAAoUJACmAD2QZEWXIQqNeg0aIBmMSAAo46OdHhJUaCDA1GEKAwEoIAb1EQI+zDCgSAdLADmOgAMAYQALCTgAa2Q0XwgAEgdYa1MAXwDbYREXZEsdAEJpWXkvaLgoAFcAEwkMHSSTdFt7JyyXCHEIaOqpHCsGzBK0boB5DTrjGzQM5zb6yYwvDAAHKGQ4CR0u6SIIEmnWlzcMD28-QJcAJQkAWxhgCUr4xInUnbnTLFkINBgAdwC3i87JkXClRGCREA). Perhaps the `fs.exists` check should be done below, just before `fs.readDir` instead?

anttimaki (changes requested):

anttimaki (approved):
LGTM

PR #1628, opened 2025-01-28
Add option for copying troubleshooting information to clipboard

anttimaki:
This allows users to share relevant information when asking for support e.g. on Discord.


VilppeRiskidev (approved):
Seems to work OK
Maybe it would be nice to have some sort of feedback when the contents are copied, but this does what it is supposed to.

anttimaki:
Added a second commit to obfuscate Windows usernames from paths as some consider those sensitive information.

The idea about feedback when the values are copied is very good but would be awkward to implement for current `SettingsRow` items.

VilppeRiskidev:
Yeah, seems to work well 👌

PR #1631, opened 2025-02-05
Update ModLoaderVariantRecord.ts

anttimaki:
Added BepInExPack_AmongUs to MODLOADER_PACKAGES


anttimaki:
Based on #1629 but fixes the root folder parameter to match the folder structure used in the package uploaded to Thunderstore. Done as a separate PR as I'm not 100% sure we want to do this instead of updating the package.

x753 (approved):

PR #1633, opened 2025-02-11
Add a getter isModListOutdated

VilppeRiskidev:
Reference: https://github.com/thunderstore-io/thunderstore-mod-manager/pull/310

### DIFF
--- a/src/store/modules/TsModsModule.ts
+++ b/src/store/modules/TsModsModule.ts
@@ -2,6 +2,7 @@ import { ActionTree, GetterTree, MutationTree } from 'vuex';
 
 import { State as RootState } from '../index';
 import ManifestV2 from '../../model/ManifestV2';
+import R2Error from "../../model/errors/R2Error";
 import ThunderstoreMod from '../../model/ThunderstoreMod';
 import VersionNumber from '../../model/VersionNumber';
 import CdnProvider from '../../providers/generic/connection/CdnProvider';
@@ -9,7 +10,7 @@ import * as PackageDb from '../../r2mm/manager/PackageDexieStore';
 import { isEmptyArray, isStringArray } from '../../utils/ArrayUtils';
 import { retry } from '../../utils/Common';
 import { Deprecations } from '../../utils/Deprecations';
-import { fetchAndProcessBlobFile, getAxiosWithTimeouts } from '../../utils/HttpUtils';
+import { fetchAndProcessBlobFile, getAxiosWithTimeouts, isNetworkError } from '../../utils/HttpUtils';
 
 export interface CachedMod {
     tsMod: ThunderstoreMod | undefined;
@@ -44,6 +45,7 @@ function isPackageListChunk(value: unknown): value is PackageListChunk {
 }
 
 const EXCLUSIONS = 'https://raw.githubusercontent.com/ebkr/r2modmanPlus/master/modExclusions.md';
+const PARTIAL_UPDATE_ERROR = 'Failed to fully refresh the online mod list. Some mod versions might be unavailable.';
 
 /**
  * For dealing with mods listed in communities, i.e. available through
@@ -121,6 +123,21 @@ export const TsModsModule = {
         isModListOutdated(state) {
             return state.modsLastUpdated instanceof Date
                 && (Date.now() - state.modsLastUpdated.getTime()) > (1000 * 60 * 60);
+        },
+
+        /*** A more concise version of the error message */
+        conciseThunderstoreModListUpdateErrorMessage(state): string|undefined {
+            if (!state.thunderstoreModListUpdateError) {
+                return undefined;
+            }
+
+            let conciseError = "Failed to load mod list";
+            if (isNetworkError(state.thunderstoreModListUpdateError)) {
+                conciseError = "Failed to fully refresh the online mod list due to network error"
+            } else if (state.thunderstoreModListUpdateError.name === PARTIAL_UPDATE_ERROR) {
+                conciseError = "Failed to fully refresh the online mod list";
+            }
+            return conciseError;
         },
 
         /*** Return ThunderstoreMod representation of a ManifestV2 */
@@ -257,7 +274,7 @@ export const TsModsModule = {
         },
 
         async fetchAndCachePackageListChunks(
-            {dispatch},
+            {commit, dispatch},
             {packageListIndex, progressCallback}: {packageListIndex: PackageListIndex, progressCallback?: ProgressCallback},
         ): Promise<boolean> {
             const chunkCount = packageListIndex.content.length;
@@ -282,6 +299,13 @@ export const TsModsModule = {
             // hash is updated in the API.
             if (successes === chunkCount) {
                 await dispatch('cacheIndexHash', packageListIndex.hash);
+            } else {
+                commit('setThunderstoreModListUpdateError',
+                    new R2Error(
+                        PARTIAL_UPDATE_ERROR,
+                        `Only ${successes} out of ${chunkCount} parts of the list were updated successfully`,
+                    )
+                );
             }
 
             return successes === chunkCount;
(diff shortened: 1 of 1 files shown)


src/store/modules/TsModsModule.ts
```diff
@@ -118,6 +118,12 @@ export const TsModsModule = {
             return getters.cachedMod(mod).isLatest;
         },
 
+        /*** Was the last successful mod list update more than an hour ago? */
+        isModListOutdated(_state) {
```
anttimaki, line 122: The `_` prefix is there to indicate the positional argument is required although it's not used by the function. Some linters would comlain about it without the prefix, but that's not true in this repo. Anyways, since the argument is used, it shouldn't be prefixed.

anttimaki (approved):

src/store/modules/TsModsModule.ts
```diff
@@ -45,6 +45,7 @@ function isPackageListChunk(value: unknown): value is PackageListChunk {
 }
 
 const EXCLUSIONS = 'https://raw.githubusercontent.com/ebkr/r2modmanPlus/master/modExclusions.md';
+const PARTIAL_UPDATE_ERROR = 'Failed to fetch and cache some of the packages.';
```
anttimaki, line 48: I think this message could be made more informative for the users. Instead of "fetch and cache" we could use the "refresh online mod list" used elsewhere in the UI, and we could explain how this affects the users. Something like:
"Failed to fully refresh the online mod list. Some mod versions might be unavailable."

It's not perfect since it doesn't tell the user whether it was one package out of one, two, or twenty that failed. But it'll probably do for now.

src/store/modules/TsModsModule.ts
```diff
@@ -118,6 +119,17 @@ export const TsModsModule = {
...
         },
 
+        /*** Was the last successful mod list update more than an hour ago? */
+        isModListOutdated(state) {
+            return state.modsLastUpdated instanceof Date
+                && (Date.now() - state.modsLastUpdated.getTime()) > (1000 * 60 * 60);
+        },
+
+        /*** Is the error from only some of the package chunks failing to be fetched/cached? */
+        isStatusPartiallyFailed(state) {
```
anttimaki, line 129: Is the error message suggestion in my previous comment too long to be shown on the status indicator? Having a better message in the error object doesn't help if the UI component just replaces it with the original PARTIAL_UPDATE_ERROR again.

Naming is this module gets quickly very confusing when similar and/or ambiguous terms are used to refer different things. I did a round of renames recently, thus some of the names are rather long. "status" is a term that can be easily misinterpreted or otherwise confusing. I would recommend something like `isThunderstoreModListUpdateErrorPartial`. 

Alternatively, to keep the UI texts consistent, we could do the shortening in this module by replacing this with a `conciseThunderstoreModListUpdateErrorMessage` getter. It could do what the UI component now does, i.e. check for Network Errors and partial failure and return a string that can be shown in UI as is.

src/store/modules/TsModsModule.ts
```diff
@@ -118,6 +119,17 @@ export const TsModsModule = {
...
+            return state.thunderstoreModListUpdateError?.message === PARTIAL_UPDATE_ERROR;
```
anttimaki, line 130: You can't use `?` like that in this project, as shown by the CI pipeline failure.

src/store/modules/TsModsModule.ts
```diff
@@ -277,6 +289,8 @@ export const TsModsModule = {
             // hash is updated in the API.
             if (successes === chunkCount) {
                 await dispatch('cacheIndexHash', packageListIndex.hash);
+            } else {
+                throw new Error('Failed to fetch and cache some of the packages.');
```
anttimaki, line 293: - If I'm reading this right. throwing here turns the partial failure to complete failure since `syncPackageList` skips the rest of the update. Instead, I would try to throw (or just commit) the error message at the end of `syncPackageList`'s try-block if all chunks weren't successful
- Not using `PARTIAL_UPDATE_ERROR` makes me seriously question how much effort you put into this PR

anttimaki (changes requested):
I think this PR was not ready for review.

src/store/modules/TsModsModule.ts
```diff
@@ -118,6 +120,27 @@ export const TsModsModule = {
...
+
+        /*** A more concise version of the error message */
+        conciseThunderstoreModListUpdateErrorMessage(state): string|undefined {
+            if (!state.thunderstoreModListUpdateError) {
+                return undefined;
+            }
+
+            let conciseError = "Failed to load mod list";
+            if (isNetworkError(state.thunderstoreModListUpdateError)) {
+                conciseError = "Failed to download mod list due to network error"
```
anttimaki, line 137: Could be "Failed to fully refresh the online mod list due to network error" to be consistent (if screen estate allows).

src/store/modules/TsModsModule.ts
```diff
@@ -277,6 +300,13 @@ export const TsModsModule = {
             // hash is updated in the API.
             if (successes === chunkCount) {
                 await dispatch('cacheIndexHash', packageListIndex.hash);
+            } else {
+                commit('setThunderstoreModListUpdateError',
+                    new R2Error(
+                        PARTIAL_UPDATE_ERROR,
+                        `Only ${chunkCount - successes} out of ${chunkCount} chunks were updated successfully.`,
```
anttimaki, line 307: - Bug? Shouldn't this be `Only ${successes} out of ${chunkCount}...`
- Nitpick but "chunks" might not mean so much to the users. "Parts" would be slightly better? Some other wording could also work.
- If this ends in `.`, there will be two of them in the `ImportProfileModal`. Fix either here or there, but make sure all places where this is shown show 0..1 full stops.
- `ImportProfileModal` and `SettingsView` currently show this error message. Is that intentional? Looks to me `PARTIAL_UPDATE_ERROR` is currently shown only if user opens the error modal in `ModListUpdateBanner`, and that might the more user-friendly part of the message. This too can be changed (if it's changed) either here in the error definition, or on the sites where the error is used.

anttimaki (changes requested):

src/store/modules/TsModsModule.ts
VilppeRiskidev: To comment on the last point: It was/is intentional, but can be changed if you think so.
I think it is nicer this way, because those two places it is prefixed with `Error updating the mod list: `. 
In my opinion `Error updating the mod list: Failed to fully refresh the online mod list. Some mod versions might be unavailable.` would be a bit redundant, but there could be better wordings for this 🤔 I think it is kind of nice to show the numbers in most cases, but not necessarily on the StatusIndicator if that makes sense (if I was a user and it showed that it downloaded 99 out of 100 parts, I probably would just continue and be aware that something might be a bit broken and 1 success out of 100 would make me expect that it most likely is going to be broken.) That's my $0.02, feel free to ask for more changes.

src/store/modules/TsModsModule.ts
VilppeRiskidev: Now that you mentioned this, it reminded me that I was also contemplating if the StatusIndicator's text should be a warning or an error (yellow vs. red text), but that's also definitely out of the scope of this PR. :D

anttimaki: It's a partial failure so I'd say it's an error. Especially since some users seem to be confused about the status unless we're very in-your-face about it.

src/store/modules/TsModsModule.ts
anttimaki: Alright, let's go with as-is.

And I think you put more effort in writing the sentences about partial failure than some users will put into the whole thing. Seems to me that for them it's a binary thing, either it failed or it didn't, and the idea of "it didn't go quite as expected but you can try to continue" just doesn't register.

anttimaki (approved):
LGTM

PR #1646, opened 2025-02-26
Display information about mod list updating being prevented

VilppeRiskidev:
Display information about mod list updating being prevented while there are downloads in progress

### DIFF
--- a/src/components/profiles-modals/ImportProfileModal.vue
+++ b/src/components/profiles-modals/ImportProfileModal.vue
@@ -321,26 +321,35 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
                 </p>
 
                 <p class="margin-top">
-                    Updating the mod list from Thunderstore might solve this issue.
+                    Refreshing the mod list from Thunderstore might solve this issue.
 
-                    <span v-if="$store.state.tsMods.modsLastUpdated">
-                        The mod list was last updated on {{ valueToReadableDate($store.state.tsMods.modsLastUpdated) }}.
-                    </span>
+                    <div v-if="$store.getters['download/activeDownloadCount'] > 0">
+                        <span>
+                            However, the mod list can't be refreshed while the are mod downloads in progress.
+                            Please wait for the downloads to finish before continuing.
+                        </span>
+                    </div>
 
-                    <br />
+                    <div v-else>
+                        <span v-if="$store.state.tsMods.modsLastUpdated">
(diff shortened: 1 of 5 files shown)


src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -326,24 +326,33 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
-                        <a @click="$store.dispatch('tsMods/syncPackageList')">Retry</a>?
-                    </span>
-                    <span v-else>
-                        Would you like to
-                        <a @click="$store.dispatch('tsMods/syncPackageList')">update now</a>?
-                    </span>
+                    <div v-if="$store.getters['download/activeDownloadCount'] > 0">
+                        <span>
+                            However, the mod list can't be updated while the are mod downloads in progress.
+                            Please wait for the downloads to finish before continuing.
```
VilppeRiskidev, line 332: This part will probably be something like "The mod list will try to refresh after the downloads are finished.", or be completely refactored in the future.

anttimaki: I'd think it will make sense to handle the update before this box is rendered, perhaps in a separate step before the preview. Then in this step we'll either say that some mods can't be imported, or list all the mods normally if they're all known by the manager.

src/components/settings-components/SettingsView.vue
```diff
@@ -314,6 +314,9 @@ import CdnProvider from '../../providers/generic/connection/CdnProvider';
                 'Refresh online mod list',
                 'Check for any new mod releases.',
                 async () => {
+                        if (this.$store.getters['download/activeDownloadCount'] > 0) {
```
anttimaki, line 317: Here too it might make sense to prefer to show the current status of update if mod downloads and mod list refresh are going on at the same time. Note that while the refresh can't be started while mods are downloading, user can still start downloads while the refresh is in progress, so both can be happening in the same time.

VilppeRiskidev:
Updated the wording here and also in https://github.com/thunderstore-io/thunderstore-mod-manager/pull/316

src/components/settings-components/SettingsView.vue
VilppeRiskidev: Oops, should be fixed now 👌

anttimaki (approved):
LGTM

PR #1647, opened 2025-02-27
Fix profile import silently leaving some mods uninstalled

anttimaki:
To reduce time spent on the splash screen, recent changes skip refreshing the online mod list if older version is found in IndexedDB cache. This means the user might have outdated mod list when importing a profile, and the manager can't import mods it doesn't know about. To mitigate the issue, a warning was added to profile preview step with an option for the user to update the mod list.

However, the warning uses the mod list in Vuex to check if a mod is available, which is a mistake since that only checks if the mod itself is known, and not whether the mod version in the profile is known. This caused recently updated mods to not always trigger the warning. The download part uses IndexedDB as its source of truth, and silently ignored the unknown mods/versions, since user has explicitly decided to import a partial profile.

To fix the issue, both parts now use IndexedDB as the source of truth.

The implementation in ImportProfileModal now somewhat akwardly handles the mod information as ExportMods (read from mods.yml and required to know which mods are disabled), ThunderstoreMods (required to use existing UI components), and ThunderstoreCombos (required to download a specific version of the mod). This could potentially be simplified but is outside the scope of this PR.

ExportMod support was dropped from TsModsModule functions to remove the footgun that caused this whole problem.

### DIFF
--- a/src/components/profiles-modals/ImportProfileModal.vue
+++ b/src/components/profiles-modals/ImportProfileModal.vue
@@ -1,11 +1,12 @@
 <script lang="ts">
 import { mixins } from "vue-class-component";
-import { Component } from 'vue-property-decorator';
+import { Component, Watch } from 'vue-property-decorator';
 
 import ManagerInformation from "../../_managerinf/ManagerInformation";
 import R2Error from "../../model/errors/R2Error";
 import ExportFormat from "../../model/exports/ExportFormat";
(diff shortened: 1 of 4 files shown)


anttimaki:
@VilppeRiskidev I'm confident enough with this that I'm going to merge it unreviewed, but do a review anyway. At least it matters for the task where the online mod list refresh is automated.

ishikawayuzu:
very good

PR #1649, opened 2025-02-28
Only filter out versions when not dealing with modpacks

VilppeRiskidev:


src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -145,10 +145,12 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
 
         await this.buildDependencySet(combo.getVersion(), dependencies, versionMode);
         this.sortDependencyOrder(dependencies);
-        // #270: Remove already-installed dependencies to prevent updating.
-        return dependencies.filter((dep) => {
-            return !(modList.some(installed => installed.getName() === dep.getMod().getFullName()));
-        });
+
+        return isModpack ? dependencies :
+            // #270: Remove already-installed dependencies to prevent updating.
```
anttimaki, line 150: It might be clearer if we dropped this comment altogether and replaced the docstring on L140 with something like:

> When installing a modpack ensure compatibility by installing the exact
> versions of dependencies even if it means upgrading/downgrading already
> installed mods. When installing a regular mod, install the latest
> versions of dependencies but don't change already installed mods.

anttimaki (approved):
LGTM and seems to fix the issue reported by user in Discord. Consider updating the comment though to avoid having misleading comments in the future.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
VilppeRiskidev: Comment: Updated the comment based on the comment.

PR #1655, opened 2025-03-06
Profile import auto refresh when some or all packages are not found

VilppeRiskidev:


src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -123,7 +125,7 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
     }
 
     // Check that selected profile zip is valid and proceed.
-    async validateProfileFile(files: string[] | null) {
+    async validateProfileFile(files: string[] | null, isRetry: boolean = false) {
```
anttimaki, line 128: `isRetry` is never used.

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -144,6 +146,25 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
         }
 
         this.profileImportFilePath = files[0];
+
+        if (this.profileMods.known.length === 0 || this.profileMods.unknown.length > 0) {
```
anttimaki, line 150: Nitpick: I think we're only interested in the number of unknown mods here? Checking number of known mods won't probably hurt, but at the same time refreshing the mod list can only solve issues with unknown mods.

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -144,6 +146,25 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
+            // Sometimes the reason some packages are unknown is that the mod is out of date,
```
anttimaki, line 151: Minor: should "mod is out of date" say "mod list is out of date"?

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -144,6 +146,25 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
+            // so let's try refreshing it
+            try {
+                this.activeStep = 'REFRESH_MOD_LIST';
+                await this.$store.dispatch('tsMods/syncPackageList');
+                this.profileMods = await ProfileUtils.exportModsToCombos(
```
anttimaki, line 156: Question: Is this needed? `syncPackageList` should update the mod list in Vuex, and that's watched by `updateProfileModsOnOnlineModListRefresh`, which does the same thing as this line? Seems to me we do this thing twice now?

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -297,18 +318,39 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
+        </template>
+    </ModalCard>
+
     <ModalCard v-else-if="activeStep === 'REVIEW_IMPORT'" key="REVIEW_IMPORT" :is-active="isOpen" @close-modal="closeModal">
         <template v-slot:header>
             <h2 class="modal-title">Packages to be installed</h2>
         </template>
         <template v-slot:body>
-            <OnlineModList :paged-mod-list="knownProfileMods" :read-only="true" />
+            <OnlineModList
```
anttimaki, line 340: A bit unrelated to the changes, but I would probably change the order so that warning box is shown on top and the `OnlineModList` below it. The rationale is that for large profiles user might not scroll the list to the bottom and don't then see the warning.

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -297,18 +318,39 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
+                v-if="profileMods.known.length > 0"
```
anttimaki, line 341: Nitpick: I think this should check `knownProfileMods.length` as that's what gets passed to the component.

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -297,18 +318,39 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
+            </p>
             <div v-if="profileMods.known.length === 0 || profileMods.unknown.length > 0" class="notification is-warning margin-top">
                 <p v-if="profileMods.known.length === 0">
-                    None of the packages in the profile were found on Thunderstore:
+                    These packages were not found:
                 </p>
                 <p v-else>
-                    Some of the packages in the profile were not found on Thunderstore:
+                    These packages in the profile were found on Thunderstore:
                 </p>
```
anttimaki, line 354: I would perhaps reconsider how these three paragraphs work together. First saying no packages were found and then listing all the packages seems redundant. Saying "These packages in the profile were found on Thunderstore" and then listing `unknownProfileModNames` seems like a bug?

Should this perhaps just say  "These packages in the profile were not found on Thunderstore" (notice the "not") inside the yellow box and then list the unknown mods for all cases?

anttimaki (changes requested):

src/components/profiles-modals/ImportProfileModal.vue
VilppeRiskidev: Good point, not needed

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -297,67 +314,46 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
     </ModalCard>
 
+    <ModalCard v-else-if="activeStep === 'REFRESH_MOD_LIST'" key="REFRESH_MOD_LIST" :is-active="isOpen" :can-close="false">
+        <template v-slot:header>
+            <h2 class="modal-title">Refreshing mod list</h2>
+        </template>
+        <template v-slot:footer>
+            <div>
+                <p>
+                    At least some of the packages were not found. We're trying to fix this by refreshing the online mod list:</br>
```
anttimaki, line 324: Minor: I wonder if it's clear to the users what "not found" means in this context? Would something like this be clearer?

> Some of the packages in the profile are not recognized by the mod manager. Refreshing the online mod list might fix the problem.

src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -297,67 +314,46 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
...
-                        </span>
-                        <span v-else>
-                            Would you like to
-                            <a @click="$store.dispatch('tsMods/syncPackageList')">refresh now</a>?
-                        </span>
-                    </div>
-                </p>
             </div>
+
+            <p v-if="knownProfileMods.length > 0" class="margin-bottom">These packages will be installed:</p>
```
anttimaki, line 345: Minor: this seems redundant if the yellow box is not shown, as the text is quite similar to the header right above it. I'd show this only when the yellow box is shown.

anttimaki (approved):
Approved with minor suggestions.

src/components/profiles-modals/ImportProfileModal.vue
VilppeRiskidev: Yeah, much better, I'll do this 👍

VilppeRiskidev:
All of the comments have been taken into account and this PR is ready to be merged.

PR #1657, opened 2025-03-08
Added dispatch to update API cache in background if cache was already present

ebkr:


anttimaki (approved):
Based on very quick testing this seems to work. Personally I would perhaps add a comment or something that the bit odd placing and missing `await` are intentional as this is supposed to happen on the background.

PR #1658, opened 2025-03-10
Add 3 new games: Dusk, Monster Train 2 and My Dream Setup

VilppeRiskidev:


VilppeRiskidev:
Also note that Monster Train 2 is currently only available as a demo, so that needs to be updated once the game releases.

anttimaki (approved):

PR #1663, opened 2025-03-11
Reload profile mod list when uninstalling a mod as an installation step

VilppeRiskidev:


src/components/mixins/DownloadMixin.vue
```diff
@@ -85,7 +85,8 @@ export default class DownloadMixin extends Vue {
         assignId: number
     ): Promise<void> {
         await ProfileModList.requestLock(async () => {
-            const modList: ManifestV2[] = await installModsToProfile(downloadedMods, profile, undefined,(status, modName, installProgress) => {
+            const reloadModList = () => this.$store.dispatch('profile/tryLoadModListFromDisk');
```
VilppeRiskidev, line 88: This could use `updateModList` instead of `tryLoadModListFromDisk`, but that would then require passing the mod list from `ProfileUtils`' `installModsToProfile`.
I thought this way would be better as the thing was to make the list match what's on the disk to begin with.

src/utils/ProfileUtils.ts
```diff
@@ -92,6 +93,9 @@ export async function installModsToProfile(
 
             // Uninstall possible different version of the mod before installing the target version.
             throwForR2Error(await ProfileInstallerProvider.instance.uninstallMod(manifestMod, profile));
+            throwForR2Error(await ProfileModList.removeMod(manifestMod, profile));
```
VilppeRiskidev, line 96: I think this _could_ be called somewhere else, either in DownloadMixin/ImportProfileModal or deeper inside the `ProfileInstallerProvider.instance.uninstallMod`, but this seemed like the place that most directly represents the actual order of operations

anttimaki (commented):
This seems to do fix the issue, but there are two reasons I don't like the implementation.

The main problem is that it will perform the disk operation for every iteration of the loop, i.e. possibly hundreds of times. The docstring of `installModsToProfile` states that it was created specifically to avoid such inefficiency.

The second and lesser problem is that it once more sneakily injects the Vuex store to the helper library, strengthening the coupling between the two. This time it can be avoided though.

I've created an alternative implementation that addresses the two issues: #1669 
Please review it for logic problems and test it like you really want to find problems with it. Note that this affects things more that just the MelonLoader case mentioned in the ticket, so try to think this from a bigger perspective.

In case you're wondering, my PR doesn't touch `ImportProfileModal` as it doesn't need to reload the mod list from the disk to update UI, as the relevant UI is not rendered at that point.

anttimaki:
Closing in favour of #1669.

PR #1669, opened 2025-03-14
Sync partial profile changes to disk in case mod (un)install fails

anttimaki:
This attempts to fix a situation where uninstalling an older version of a mod is successful but installing the replacement version fails. Before the changes the UI kept showing the uninstalled mod still in the profile. It will have effects beyond that, e.g. if installation of a third mod fails, the first two will be shown on the profile correctly.

Ideally the installation would be an atomic operation but creating a rollback functionality is beyond the current scope.


VilppeRiskidev (approved):
Ahh, yes, very nice. Similar to how I _tried_ to do it (before doing it the way I did), but better.

PR #1672, opened 2025-03-18
Update mod list refresh status in SplashMixin

anttimaki:
This is mostly for TSMM so it can show something's happening in the new status indicator.


VilppeRiskidev (commented):
Makes sense 👍

PR #1673, opened 2025-03-18
Change how profile import wizard waits for online mod list refresh

anttimaki:
Previously, if the mod list refresh was already in progress when user entered the related profile import step, the refresh triggered by the modal immediately returned and the import process progressed to preview although the mod list refresh was still in progress.

We now poll the status in a while-loop and hope any bugs don't cause infinite loops.

Also includes slight touches to UI texts and styles.


VilppeRiskidev (commented):
Nice! 👍

PR #1674, opened 2025-03-18
Show if active mod downloads are blocking a profile import

anttimaki:
The part of the profile import that force refreshes online mod list didn't take into account that active mod downloads blocks the refresh. The import modal will now wait for downloads to finish before triggering the online mod list refresh.


VilppeRiskidev (commented):
Nice! 👍

PR #1675, opened 2025-03-18
Fix download progress getting stuck

VilppeRiskidev:


anttimaki (changes requested):
- Now that the fix was changed a bit, the commit message is a bit confusing and should be reworded to match the actual changes
- I'm not 100% certain this is the same bug I've encountered since while testing the issue described in the commit message on `develop` branch, the behaviour is a bit different: the download percentage jumps back and forth but eventually reaches 100% and the install progress increases normally, whereas for me the download got stuck midway and the install progress stayed at 0%.
  - That being said, the changes here do fix at least one issue and should thus be merged. With some luck, the fix also resolves the other issue, if there indeed was another similar issue at play.

VilppeRiskidev:
Updated the commit message (a bit shorter now, but descriptive enough IMO).

anttimaki (approved):

PR #1678, opened 2025-03-21
Add installation tracking to mod installation when it's initiated by a link

VilppeRiskidev:
- I also made it correctly update the status (to `failed: true`) when the installation fails

### DIFF
--- a/src/components/views/DownloadModModal.vue
+++ b/src/components/views/DownloadModModal.vue
@@ -103,24 +103,35 @@ import ProfileModList from '../../r2mm/mods/ProfileModList';
                             }
                         );
                     } catch (e) {
+                        store.commit('download/updateDownload', { assignId, failed: true });
                         return reject(e);
                     }
                     await ProfileModList.requestLock(async () => {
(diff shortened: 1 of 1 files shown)


src/components/views/DownloadModModal.vue
```diff
@@ -103,24 +103,36 @@ import ProfileModList from '../../r2mm/mods/ProfileModList';
...
                         return reject(e);
                     }
                     await ProfileModList.requestLock(async () => {
+                        let currentDownloadIndex = 0;
                         for (const combo of downloadedMods) {
                             try {
                                 await DownloadModModal.installModAfterDownload(profile, combo.getMod(), combo.getVersion());
                             } catch (e) {
+                                store.commit('download/updateDownload', { assignId, failed: true });
+                                currentDownloadIndex++;
```
anttimaki, line 116: I think this does nothing as we `return` on the next line? A bit confusing for the reader though, so consider removing if it's useless.

anttimaki (approved):
LGTM and passes manual testing,

PR #1680, opened 2025-03-23
Mod preview panel

ebkr:


src/components/modals/SortModal.vue
```diff
@@ -0,0 +1,75 @@
...
+import SortingStyle from '../../model/enums/SortingStyle';
+import useStore from '../../store';
+import { computed, ref } from 'vue';
+
+const store = useStore();
+
+const directions = Object.keys(SortDirection);
+const behaviours = Object.keys(SortingStyle);
+
+const selectedBehaviour = computed(() => store.state.modFilters.sortBehaviour);
```
anttimaki, line 13: I noticed the [computed accepts a setter](https://vuejs.org/guide/essentials/computed.html#writable-computed), e.g. something like this:
```typescript
const selectedBehaviour = computed({
    get() {
        return store.state.modFilters.sortBehaviour;
    },
    set(value: string) {
        // Note that setSortBehaviour currently expects object key and we pass the value directly.
        store.commit("modFilters/setSortBehaviour", value);
    }
});
```

Markup could then use v-model:
```html
<select class="select select--content-spacing" v-model="selectedBehaviour">
    <option v-for="(value, key) in SortingStyle" :value="value" :key="`sort-behaviour--${key}`">
        {{ value }}
    </option>
</select>
```

This would remove the need to swap between the key and the value.

src/components/v2/MarkdownRender.vue
```diff
@@ -0,0 +1,99 @@
+<script lang="ts" setup>
+import { computed } from 'vue';
+import markdownit from 'markdown-it';
```
anttimaki, line 3: I'm not 100% certain how experimental this is, but the new website uses an undocumented endpoint, e.g. `https://thunderstore.io/api/cyberstorm/package/ymc_mhz/morehead/latest/readme/` (also available for changelog). The major thing is it returns HTML instead of MD, so we could avoid requiring these dependencies, and IIRC get some security against XSS at the same time. Additionally it would make it easier to have the content look the same on the manager and the website, so the author doesn't need to specifically target one or another.

Also Oksamies might be able to provide some CSS that would make it even easier to get the content look the same, but I'm not 100% sure how feasible that is.

src/components/v2/MarkdownRender.vue
```diff
@@ -0,0 +1,99 @@
...
+import { markdownItTable } from 'markdown-it-table';
+import LinkProvider from 'src/providers/components/LinkProvider';
```
anttimaki, line 5: Non-relative import. There's more of these in other files.

src/components/v2/MarkdownRender.vue
```diff
@@ -0,0 +1,99 @@
...
+    }
+    return e;
+}
+</script>
+
+<template>
+    <p class="markdown-body" v-html="markdownToRender" @click="captureClick"></p>
+</template>
+
+<style lang="scss">
```
anttimaki, line 46: Is the `scoped` attribute intentionally omitted here? `table th` looks like a rather generic selector. (I don't quite remember how the unscoped styles bleed around in Vue.)

src/components/v2/OnlinePreviewPanel.vue
```diff
@@ -0,0 +1,195 @@
...
+        if (props.mod === modToLoad) {
+            changelog.value = res;
+        }
+    });
+}
+
+function fetchAll(modToLoad: ThunderstoreMod) {
+    setActiveTab("README");
+    loadingPanel.value = true;
+    buildDependencies(modToLoad)
```
anttimaki, line 57: - Could consider error handlers in case of the Network errors
- Any specific reason for chaining the loading of the tab contents instead of triggering them all at once?
- If we want to keep them chained, should loading README be the first in the chain as it's always shown first?
- The way loadingPanel is used in the markup, all of the tabs show "fetching" until all of the tabs have finished loading (and get stuck to that if one of the other tabs fail to load)

src/components/v2/OnlinePreviewPanel.vue
```diff
@@ -0,0 +1,195 @@
...
+
+<template>
+    <div class="c-preview-panel">
+        <div class="c-preview-panel__header">
+            <h1 class="title">{{ mod.getName() }}</h1>
+            <h2 class="subtitle">
+                By {{ mod.getOwner() }}
+            </h2>
+            <p class='card-timestamp'><strong>Downloads:</strong> {{mod.getDownloadCount()}}</p>
+            <p class='card-timestamp'><strong>Likes:</strong> {{mod.getRating()}}</p>
```
anttimaki, line 118: Is the donation option intentionally omitted in this component? It's still available via the heart icon on the OnlineRowCard, but less prominent than it was in the old collapsible section.

src/components/v2/OnlinePreviewPanel.vue
```diff
@@ -0,0 +1,195 @@
...
+                        <p>Fetching {{ activeTab }} for {{ props.mod.getFullName() }}</p>
+                    </div>
+                </div>
+            </template>
+            <template v-else-if="activeTab === 'Dependencies'">
+                <template v-if="dependencies.length > 0">
+                    <OnlineModList :paged-mod-list="dependencies" :read-only="true" />
+                </template>
+                <template v-else>
+                    <p>No dependencies</p>
```
anttimaki, line 151: This could use the "notification"/"container" wrappers to visually match "No changelog" case.

src/components/views/OnlineModListWithPanel.vue
```diff
@@ -0,0 +1,107 @@
...
+    @Prop()
+    pagedModList!: ThunderstoreMod[];
+
+    @Prop()
+    selectedMod: ThunderstoreMod | null = null;
+
+    @Prop({default: false})
+    readOnly!: boolean;
+
+    private funkyMode: boolean = false;
```
anttimaki, line 71: - This doesn't seem to be used?
- Thus `created` is unnecessary
- And since OnlineRowCards can't be expanded, `${settings.getContext().global.expandedCards}` could be removed from the key attribute
- Thus `settings` getter is unnecessary

src/components/views/OnlineModView.vue
```diff
@@ -1,71 +1,82 @@
...
-                <div class="no-padding-left card-header-title">
-                    <div class="input-group input-group--flex margin-right">
-                        <label for="online-search">Search</label>
-                        <DeferredInput
-                            v-model="thunderstoreSearchFilter"
-                            id="online-search"
-                            class="input"
-                            type="text"
-                            placeholder="Search for a mod"
-                            autocomplete="off"
```
anttimaki, line 14: This shouldn't be removed, it's needed in TSMM.

src/components/views/OnlineModView.vue
```diff
@@ -1,71 +1,82 @@
...
+                                    id="thunderstore-search-filter"
+                                    class="input"
+                                    type="text"
+                                    placeholder="Search"
+                                />
+                            </div>
+                            <div class="input-group">
+                                <div class="input-group input-group--flex">
+                                    <label for="thunderstore-category-filter">&nbsp;</label>
+                                    <button
```
anttimaki, line 21: I like how using the modal button here instead of the selects clean up the UI 👍Outside of the scope here but we should consider something similar to installed mod view?

src/components/views/OnlineModView.vue
```diff
@@ -100,6 +115,7 @@ export default class OnlineModView extends Vue {
     sortingDirectionModel = SortingDirection.STANDARD;
     sortingStyleModel = SortingStyle.DEFAULT;
```
anttimaki, line 116: These could be cleaned up now.

src/components/views/OnlineModView.vue
```diff
@@ -226,8 +234,59 @@ export default class OnlineModView extends Vue {
         });
     }
 
+    openPreviewForMod(mod: ThunderstoreMod) {
```
anttimaki, line 237: Nitpick: really a "toggle" rather than "open" function.

src/store/modules/ModFilterModule.ts
```diff
@@ -60,6 +68,15 @@ export default {
             state.allowNsfw = value;
         },
 
+        setSortBehaviour(state: State, sortBehaviour: keyof typeof SortingStyle) {
+            state.sortBehaviour = SortingStyle[sortBehaviour];
+        },
+
+        setSortDirection(state: State, direction: keyof typeof SortDirection) {
+            console.log(direction)
```
anttimaki, line 76: Leftover

anttimaki:
Some color-related readability issues especially on dark mode:
image

Tables, at least inside a collapsible element, don't render contents:
image

Both examples are from https://new.thunderstore.io/c/repo/p/YMC_MHZ/MoreHead/

anttimaki:
Not sure what part of the changes, but TSMM now has two scroll bars, making things awkward. Also don't know if it's easier to address this on TSMM side or R2MM.

image

src/components/v2/OnlinePreviewPanel.vue
ebkr: > Could consider error handlers in case of the Network errors

Valid. I can keep track and display relevant errors where appropriate in the form of the notification boxes if that approach sounds good?

> Any specific reason for chaining the loading of the tab contents instead of triggering them all at once?

It's just a lower number of requests in one hit. Node really doesn't like a large number of concurrent requests and these could pile up when combined with background downloads. I'm happy to change this though as it's only one extra call.

> If we want to keep them chained, should loading README be the first in the chain as it's always shown first?

Good spot, will change.

src/components/v2/OnlinePreviewPanel.vue
ebkr: Looks like it got removed when I was playing around with the website link. Will re-add.

I do think it would make sense to move it in the future. It should probably belong in the installed list rather than the online section as I don't know why people would want to donate before having had downloaded the mod 🤔

src/components/views/OnlineModView.vue
ebkr: We should at some point. It would be nice to add the panel to the installed list too and would make sense to do it for a release following this one.

I'm not doing it currently as I think we can do this incrementally and the online section is the bigger benefit IMO. We can also gauge reactions more this way without a large scale overhaul.

For the installed one we'd instead need to read from the cache, and some may not even have a README if they were locally imported. It's a bit of a logic change and I think we can address it separately.

ebkr (commented):
Some thoughts on the approach. 

It's also worth noting that I don't expect TSMM to immediately pick this up and is something that should be refined because we need to account for ad space and minimum sizes too. Of which if we decide that TSMM shouldn't immediately pick this up, we can provide the old mod list instead.

src/App.vue
```diff
@@ -1,6 +1,8 @@
 <template>
     <div>
-        <router-view v-if="visible"/>
+        <div class="router-view">
```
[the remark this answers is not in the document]
ebkr: Nothing anymore - removed 🔥

src/components/v2/OnlinePreviewPanel.vue
anttimaki: > I can keep track and display relevant errors where appropriate in the form of the notification boxes if that approach sounds good?

Sounds good to me.

src/components/modals/SortModal.vue
ebkr: This doesn't actually appear to work correctly 🤔 I've had a play around with it and I can't really figure out why v-model isn't liked by the select element.

You need to select the option again, after the state change, for it to appear as though it were actually selected.

(Ignore the text)

https://github.com/user-attachments/assets/80c49f59-f93a-4f71-9b38-27821a94277d

anttimaki: @ebkr Did you take this comment into account:
```typescript
// Note that setSortBehaviour currently expects object key and we pass the value directly.
```

Changing the mutator to use the value seems to work:

```typescript
setSortBehaviour(state: State, sortBehaviour: string) {
    state.sortBehaviour = sortBehaviour;
},
```

That does erode the type safety a bit (at least theoretically, in practice our current Vuex version doesn't check the types anyway). It could be fixed with using a proper enum, something like:

```typescript
enum SortingStyle {
    DEFAULT = 'Default',
    LAST_UPDATED = 'Last updated',
    ALPHABETICAL = 'Alphabetical',
    DOWNLOADS = 'Download count',
    RATING = 'Rating',
};

interface State {
    sortBehaviour: SortingStyle;
}


state: (): State => ({
    sortBehaviour: SortingStyle.DEFAULT
}),

mutations: {
    setSortBehaviour(state: State, sortBehaviour: SortingStyle) {
        state.sortBehaviour = sortBehaviour;
    },
```

package.json
```diff
@@ -45,9 +45,13 @@
         "fflate": "^0.8.2",
         "floating-vue": "^1.0.0-beta.19",
         "fs-extra": "^8.1.0",
+        "github-markdown-css": "^5.7.0",
         "glob-parent": "^6.0.2",
         "highlight.js": "^10.4.1",
         "lodash.debounce": "^4.0.8",
+        "markdown-it": "^14.1.0",
```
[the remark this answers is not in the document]
ebkr: Ah no, good spot. Will remove.

anttimaki (approved):
:shipit:

PR #1682, opened 2025-03-25
Add a button to remove an inactive download from the list in DownloadMonitor

VilppeRiskidev:
- Added a corresponding vuex mutation
- Separated indexOfDownloadProgress function
- Removed the "There was a mismatch between download update's assign ID" check as that is now expected behavciour when a download is removed from the list.

### DIFF
--- a/src/store/modules/DownloadModule.ts
+++ b/src/store/modules/DownloadModule.ts
@@ -93,24 +93,19 @@ export const DownloadModule = {
     },
 
     mutations: {
+        removeDownload(state: State, download: UpdateObject) {
+            const index = getIndexOfDownloadProgress(state.allDownloads, download.assignId);
+            if (index > -1) {
+                state.allDownloads.splice(index, 1);
+            }
+        },
         updateDownload(state: State, update: UpdateObject) {
-            const newDownloads = [...state.allDownloads];
-            const index = newDownloads.findIndex((old: DownloadProgress) => {
-                return old.assignId === update.assignId;
-            });
-
-            if (index === -1) {
-                // The DownloadProgress by the ID from the update wasn't found at all.
-                console.warn(`Couldn\'t find DownloadProgress object with assignId ${update.assignId}.`);
-                return;
+            const index: number = getIndexOfDownloadProgress(state.allDownloads, update.assignId);
+            if (index > -1) {
+                const newDownloads = [...state.allDownloads];
+                newDownloads[index] = {...newDownloads[index], ...update};
+                state.allDownloads = newDownloads;
             }
-
-            if (index !== update.assignId) {
-                console.log(`There was a mismatch between download update\'s assign ID (${update.assignId}) and the index it was found at (${index}).`);
-            }
-
-            newDownloads[index] = {...newDownloads[index], ...update};
-            state.allDownloads = newDownloads;
         },
         // Use actions.toggleIngoreCache to store the setting persistently.
         setIgnoreCacheVuexOnly(state: State, ignoreCache: boolean) {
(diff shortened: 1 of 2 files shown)


VilppeRiskidev:
I think there's a bug where this causes the "There was a mismatch between download update's assign ID"... I'll fix that.

src/store/modules/DownloadModule.ts
```diff
@@ -93,24 +93,19 @@ export const DownloadModule = {
...
-            const newDownloads = [...state.allDownloads];
-            const index = newDownloads.findIndex((old: DownloadProgress) => {
-                return old.assignId === update.assignId;
-            });
-
-            if (index === -1) {
-                // The DownloadProgress by the ID from the update wasn't found at all.
-                console.warn(`Couldn\'t find DownloadProgress object with assignId ${update.assignId}.`);
-                return;
+        remove(state: State, assignId: number) {
```
anttimaki, line 96: Minor: we have `addDownload` and `updateDownload` so this should probably be `removeDownload`.

Nitpick: to match `updateDownload` this could perhaps take `UpdateObject` as the argument instead of the `assignId`.

src/store/modules/DownloadModule.ts
```diff
@@ -93,24 +93,19 @@ export const DownloadModule = {
...
+            const index: number = indexOfDownloadProgress(state.allDownloads, assignId);
```
anttimaki, line 97: Nitpick: personally I would omit the `: number` here as the return type of `indexOfDownloadProgress` is strongly typed and the type of `index` in unlikely to cause confusion in the context of this mutator. Granted, it doesn't really hurt either but for me it seems like a distraction when reading this code.

src/store/modules/DownloadModule.ts
```diff
@@ -121,3 +116,16 @@ export const DownloadModule = {
         }
     },
 }
+
+function indexOfDownloadProgress(allDownloads: DownloadProgress[], assignId: number): number {
```
anttimaki, line 120: Minor: I would name this `getIndexOfDownloadProgress`.

src/store/modules/DownloadModule.ts
```diff
@@ -121,3 +116,16 @@ export const DownloadModule = {
...
+    const index = [...allDownloads].findIndex((downloadProgress: DownloadProgress) => {
```
anttimaki, line 121: Minor: this too has unnecessarily written out types, as well as unnecessarily using braces with an arrow function.

```typescript
const index = [...allDownloads].findIndex((downloadProgress) => downloadProgress.assignId === assignId);
```

src/store/modules/DownloadModule.ts
```diff
@@ -121,3 +116,16 @@ export const DownloadModule = {
...
+        return downloadProgress.assignId === assignId;
+    });
+
+    if (index === -1) {
+        // The DownloadProgress by the ID wasn't found at all.
```
anttimaki, line 126: Nitpick: this comment doesn't really provide value.

src/store/modules/DownloadModule.ts
```diff
@@ -121,3 +116,16 @@ export const DownloadModule = {
...
+        console.warn(`Couldn\'t find DownloadProgress object with assignId ${assignId}.`);
```
anttimaki, line 127: Nitpick: no need to escape the apostrophe when using backticks to define a string, just makes the reading harder.

anttimaki (approved):
Nice, code-wise this looks good to me. Couldn't resist leaving some nitpicks related to code style, but they're not blockers.

I would perhaps consider the UI again though. The remove button is positioned kind of awkwardly. Vertically it doesn't seem to be aligned in any sensible way (I know it's positioned at the top of its container but that's not what the user sees due to other whitespace around it). Horizontally there's less margin between the button and progress bar than the button has on its other sides.

Take into account that in the future the failed downloads would have two buttons here, "remove" and "retry". So something smaller like icon-sized button, without the borders and padding of the button, might be easier to deal with. I can't recall if there's support for such buttons in the code base currently.

And to clarify the UI side isn't necessarily a blocker either.

anttimaki:
@VilppeRiskidev What's the status of this PR? There's a force push but doesn't seem the review comments are addressed one way or another.

VilppeRiskidev:
The comments have been addressed.
Action button's styles look good to my eye now 🤷‍♂️ .

anttimaki (approved):
The icon is a bit large but otherwise this looks nice now.

PR #1683, opened 2025-03-25
Add a button which clears all inactive downloads form the list in DownloadMonitor

VilppeRiskidev:


VilppeRiskidev:
I'm still considering if I should:
- Separate activeDownloads into a function to reduce repeating code
- Align the "Clear inactive" -button to right

src/pages/DownloadMonitor.vue
```diff
@@ -13,6 +13,14 @@
             </div>
         </template>
         <template v-else>
+            <div class="inherit-background-colour sticky-top sticky-top--search non-selectable border-at-bottom py-2">
+                <button
+                    class="button ghost"
+                    @click="$store.commit('download/removeAllInactive')"
+                >
+                    <i class="fas fa-times mr-2" />Clear inactive
```
anttimaki, line 21: Would "Clear finished" be more understandable for the users?

anttimaki (approved):
Nice.

> Separate activeDownloads into a function to reduce repeating code

I don't see why not. Alternatively `removeAllInactive` could be an action which have access to getters. But keeping this a mutation might be more consistent.

> Align the "Clear inactive" -button to right

I was wondering the exact same thing before I saw this comment. Didn't test it so IDK if I'd actually prefer that.

anttimaki (changes requested):
Now the button looks bad as it touches the edge of the app i.e. has no right margin at all (at least in r2mm).

Also clicking the button turns the icon/text dark and it then stays that way (focus state or something), this is easily visible when clearing the finished downloads while there are other downloads still active. IDK if it's issue specific to this button here. If it's an issue with other buttons too, then it should be fixed in a separate PR probably.

VilppeRiskidev:
> Now the button looks bad as it touches the edge of the app i.e. has no right margin at all (at least in r2mm).
> 
> Also clicking the button turns the icon/text dark and it then stays that way (focus state or something), this is easily visible when clearing the finished downloads while there are other downloads still active. IDK if it's issue specific to this button here. If it's an issue with other buttons too, then it should be fixed in a separate PR probably.

Seems like bulma uses `:focus` and not `:focus-visible`, which makes it visible when navigating with a mouse (👎 ). I can maybe try overriding it...

VilppeRiskidev:
- Fixed the alignment issue which prompted me to do this: https://github.com/thunderstore-io/thunderstore-mod-manager/pull/332 
- Fixed focus style issue here: https://github.com/ebkr/r2modmanPlus/pull/1684

anttimaki (approved):
👍🏻

anttimaki:
Force-pushed to resolve merge conflicts.

PR #1684, opened 2025-03-27
Fix focus styles for buttons

VilppeRiskidev:
A bit hard to test without the `clear-inactive-downloads-all` branch, but doesn't technically depend on that.

### DIFF
--- a/src/css/custom.scss
+++ b/src/css/custom.scss
@@ -436,6 +436,19 @@ code {
     outline-color: yellow !important;
 }
 
+.button:focus-visible {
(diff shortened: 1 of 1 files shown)


VilppeRiskidev:
The issue was that bulma messes with the focus styles and makes the experience bad for mouse users.
(the focus styles apply regardless of if a mouse or a keyboard is used to navigate, thus making the focus style stay on after a button is clicked, which looks bad).

anttimaki (approved):
LGTM

PR #1685, opened 2025-03-27
Add Schedule I

Oksamies:


src/r2mm/installing/default_installation_rules/InstallationRuleApplicator.ts
```diff
@@ -166,6 +166,7 @@ export default class InstallationRuleApplicator {
             buildBepInExRules("DUSK"),
             buildBepInExRules("MyDreamSetup"),
             buildBepInExRules("MonsterTrain2"),
+            buildBepInExRules("ScheduleI"),
```
anttimaki, line 169: This is not needed, and if it was, BepInExRules probably wouldn't work with MelonLoader game.

anttimaki (changes requested):

anttimaki (approved):

PR #1687, opened 2025-03-28
Add generic support for launching games with only a launch arg pointing to profile folder

anttimaki:


ebkr (approved):
I'm happy with the approach outlined here. It might be nice if we can move the launch arg to be more generic as opposed to the one-size-fits-all solution though.

Granted we likely don't want to bloat GameManager any more than necessary, is there somewhere that could be a better fit? I'm thinking we don't want to necessarily enforce that a game takes `mods-path` as the argument because we don't really have much control over what the application accepts.

For the purpose of Ancient Dungeon and Starship, if this works, then for now this seems fine.

anttimaki:
> I'm happy with the approach outlined here. It might be nice if we can move the launch arg to be more generic as opposed to the one-size-fits-all solution though.
> 
> Granted we likely don't want to bloat GameManager any more than necessary, is there somewhere that could be a better fit? I'm thinking we don't want to necessarily enforce that a game takes `mods-path` as the argument because we don't really have much control over what the application accepts.
> 
> For the purpose of Ancient Dungeon and Starship, if this works, then for now this seems fine.

@ebkr I added the [fifth commit](https://github.com/ebkr/r2modmanPlus/pull/1687/commits/fa4dadbeff463d25a9eabe0636bf4382826aa9d6) that introduces a new rule generator that accepts hardcoded launch arguments. By default it provides no launch args, and that's what is now used with PackageLoader.NONE as the default. It can also be used as a game-specific override, but in this commit I used a subclass of it for Ancient Dungeon.

If you're still happy with the changes, I'll probably separate the "Add Starship" part from this PR and merge.

ebkr:
That looks good to me 👍

anttimaki:
Force-pushed to remove Starship from this PR as I'll have to check if the community is still interested.

Merging this as is. I'll get back to this if I can come up with a use case where throwing exceptions seems like a better idea than letting the game launch with empty launch arguments.

PR #1691, opened 2025-04-02
Add a checkbox confirmation to ImportProfileModal when partially importing a profile

VilppeRiskidev:


src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -34,6 +34,7 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
     private profileImportFilePath: string | null = null;
     private profileImportContent: ExportFormat | null = null;
     private profileMods: {known: ThunderstoreCombo[], unknown: string[]} = {known: [], unknown: []};
+    private partialImportCheckbox: boolean = false;
```
anttimaki, line 37: Minor: `allowPartialImport` or `isPartialImportAllowed` would IMO be a better name

anttimaki (changes requested):
Overall the changes look great but (at least on r2mm):

- The label covers the checkbox
- Would be nice if the checkbox would be vertically centered, now it seems to align with the baseline of the text
- The tick in the checkbox seems a bit small. It's fine if it's that's the default Bulma style, or if it's line with other checkboxes (I couldn't remember where one would be used so didn't check).

image

VilppeRiskidev:
> Overall the changes look great but (at least on r2mm):
> 
> * The label covers the checkbox
> * Would be nice if the checkbox would be vertically centered, now it seems to align with the baseline of the text
> * The tick in the checkbox seems a bit small. It's fine if it's that's the default Bulma style, or if it's line with other checkboxes (I couldn't remember where one would be used so didn't check).

image
Looks like this on my machine 🤔

VilppeRiskidev:
Oh, it only loads those styles after past the profile selection. Coming back to impoting then has that style. It's got something to do with the dark/light mode setting I think... I'll try to fix this.

ebkr (commented):
Are the checkboxes usually vertically aligned? (See online > filter categories)

If not I don’t think we should make that change

anttimaki (approved):
👍🏻

anttimaki:
> Are the checkboxes usually vertically aligned? (See online > filter categories)
> 
> If not I don’t think we should make that change

Yeah let's keep them consistent.

PR #1692, opened 2025-04-03
Add a retry button to failed downloads in DownloadMonitor

VilppeRiskidev:


src/pages/DownloadMonitor.vue
```diff
@@ -83,6 +83,14 @@
 
                             </div>
                         </div>
+                        <button
+                            v-if="downloadObject.failed"
+                            class="button download-item-action-button"
+                            v-tooltip.left="'Retry'"
+                            @click="$store.dispatch('download/retryDownload', downloadObject)"
+                        >
+                            <i class="fas fa-redo" />
```
anttimaki, line 92: IMO the retry icon is now comically large when compared to rest of the UI (even if it's the same size in pixels as the remove icon, it's visually larger).

src/store/modules/DownloadModule.ts
```diff
@@ -57,6 +59,13 @@ export const DownloadModule = {
             settings.setIgnoreCache(!settings.getContext().global.ignoreCache);
             commit('setIgnoreCacheVuexOnly', settings.getContext().global.ignoreCache);
         },
+        async retryDownload({commit, rootGetters}, download: DownloadProgress) {
+            commit('removeDownload', download);
+
+            download.initialMods.forEach(combo => {
```
anttimaki, line 65: Clever to call the static method here.

However, having to call it separately for each combo is IMO a problem. When using "update all" for a profile with hundred mods pending update, this creates hundred new rows into the DownloadMonitor's list. While this works it makes the UI quite unresponsive, the downloads takes ages as there's so many simultaneous ones, and the installation takes ages as each download requests and waits for file lock to update the mods.yml.

Although I've no clear path to it yet, I think we can approach this problem smarter if we do some other refactoring first. We can discuss them in voice chat. So I'd suggest we leave this task on hold for now.

VilppeRiskidev:
Do the refactorings first, then redo this retry thingy.

PR #1697, opened 2025-04-07
Remove the promise and instead just throw the errors in installModAfterDownload

VilppeRiskidev:


src/components/views/DownloadModModal.vue
```diff
@@ -184,47 +184,35 @@ import ProfileModList from '../../r2mm/mods/ProfileModList';
             }, 1);
         }
 
-        static async installModAfterDownload(profile: Profile, mod: ThunderstoreMod, version: ThunderstoreVersion): Promise<R2Error | void> {
-            return new Promise(async (resolve, reject) => {
+        static async installModAfterDownload(profile: Profile, mod: ThunderstoreMod, version: ThunderstoreVersion): Promise<void> {
```
anttimaki, line 187: I was going to write that we need to change the call site to handle errors no longer being returned but seems they haven't been handler previously? Now they are though. I _guess_ that's a good thing?

src/components/views/DownloadModModal.vue
```diff
@@ -184,47 +184,35 @@ import ProfileModList from '../../r2mm/mods/ProfileModList';
...
+            const profileModList = await ProfileModList.getModList(profile.asImmutableProfile());
+            if (profileModList instanceof R2Error) {
```
anttimaki, line 189: Could use `throwForR2Error` here?

VilppeRiskidev: Didn't do it at first, because I couldn't figure out how to make typescript happy with that, but now I did. :D

anttimaki: Ah, right. If you would've mentioned it messes with type inference I would've probably told you to not bother with it 😅

anttimaki (approved):

PR #1698, opened 2025-04-07
Use ThunderstoreCombo and ImmutableProfile types in installModAfterDownload

VilppeRiskidev:
in DownloadModModal's installModAfterDownload() and ManifestV2


anttimaki (commented):
Alright but what about changing the Profile argument to ImmutableProfile?

anttimaki (approved):

PR #1699, opened 2025-04-07
Remove an unnecessary check for if the mod is BepInEx

VilppeRiskidev:


anttimaki (changes requested):
The check is actually if the mod is NOT BepInEx. And the idea had been that BepInEx shouldn't be uninstalled before installing another version because that could break things. But since it apparently hasn't caused problems for all BepInEx games that use a different BepInEx package than the hard coded one, it's safe to assume it's an unnecessary precaution. What that in turn means, is that we want to uninstall old version of any packages before installing a new one, so the uninstall command shouldn't be removed, only the wrapping if-block.

It would probably be good if the commit message would mention something along those lines instead of just stating that the check was unnecessary. Just in case some archeologist digs this up.

PR #1700, opened 2025-04-07
Perform some cleanup for installModAfterDownload

VilppeRiskidev:
- Merge two conditionals
- Use return early pattern


src/model/ManifestV2.ts
```diff
@@ -284,4 +284,8 @@ export default class ManifestV2 {
     public getDependencyString(): string {
         return `${this.getName()}-${this.getVersionNumber().toString()}`;
     }
+
+    public getAuthorModNameString(): string {
```
anttimaki, line 288: I think `.getName()` might already be returning the same thing?

anttimaki (approved):

PR #1704, opened 2025-04-09
Export mod loader mappings from TS module

MythicManiac:
Export all mod loader mappings from the TS module they're declared in. This is done in order to allow the ecosystem-schema import script to render them into yaml/json


anttimaki (approved):
👍🏻

PR #1705, opened 2025-04-10
Remove Promises and other unnecessary waiting

VilppeRiskidev:


src/components/views/DownloadModModal.vue
```diff
@@ -78,63 +78,55 @@ import ProfileModList from '../../r2mm/mods/ProfileModList';
...
-                    'download/addDownload',
-                    [`${combo.getMod().getName()} (${combo.getVersion().getVersionNumber().toString()})`]
+            let downloadedMods: ThunderstoreCombo[] = [];
+            try {
+                downloadedMods = await ThunderstoreDownloaderProvider.instance.download(
+                    profile,
+                    combo,
+                    store.state.download.ignoreCache,
+                    (downloadProgress: number, modName: string, status: number, err: R2Error | null) => {
+                        try {
```
anttimaki, line 93: Minor: as the error is just re-thrown, using the try-catch block here is just noise.

src/components/views/DownloadModModal.vue
```diff
@@ -152,7 +144,7 @@ import ProfileModList from '../../r2mm/mods/ProfileModList';
             tsCombo.setMod(tsMod);
             tsCombo.setVersion(tsVersion);
 
-            setTimeout(async () => {
+            await Promise.resolve().then(async () => {
```
anttimaki, line 147: Nitpick: would be a bit cleaner to not do this here just to revert it in the next PR.

src/pages/Manager.vue
```diff
@@ -637,18 +637,20 @@ import ModalCard from '../components/ModalCard.vue';
...
-                            await this.$store.dispatch('profile/updateModList', modList);
-                        } else {
-                            this.$store.commit('error/handleError', modList);
-                        }
-                    })
-                    .catch(
-                        (err: R2Error) => this.$store.commit('error/handleError', err)
-                    );
+
+				try {
```
anttimaki, line 641: Would this be an improvement?

```
				try {
                	await DownloadModModal.downloadSpecific(this.profile.asImmutableProfile(), combo, this.$store);
				    const modList = await ProfileModList.getModList(this.profile.asImmutableProfile());
                    if (modList instanceof R2Error) {
                        throw modList;
                    }
                    await this.$store.dispatch('profile/updateModList', modList);
				} catch (err) {
					this.$store.commit('error/handleError', R2Error.fromThrownValue(err));
				}
```

`throwForR2Error` seems to be imported but not used.

anttimaki (approved):
LGTM. I think here it would be fine to do a cleanup commit/PR at the end of the stack to address the minor issues that pile up during the PRs stack.

src/components/views/DownloadModModal.vue
VilppeRiskidev: I'll fix this in an upcoming cleanup PR 👍

src/pages/Manager.vue
VilppeRiskidev: I'll fix this in an upcoming cleanup PR 👍

PR #1708, opened 2025-04-10
Move installSpecific away from UI component and rename the dispatches

VilppeRiskidev:


src/utils/DownloadUtils.ts
```diff
@@ -17,3 +17,5 @@ export function addSolutionsToError(err: R2Error): void {
         err.solution = 'Using "Change data folder" option in the settings to select a shorter path might solve the issue';
     }
 }
+
+// TODO: move generateProgressPercentage to DownloadUtils
```
anttimaki, line 21: Nitpick: IIRC ebkr isn't a fan of this sort of TODO comments (one's you don't plan to fix immediately) as they tend to be forgotten. We should probably create tasks for them instead to keep the codebase clean and to actually remember to do them.

PR #1710, opened 2025-04-11
Move downloadAndInstallSpecific away from the UI component

VilppeRiskidev:


src/store/modules/DownloadModule.ts
```diff
@@ -126,6 +126,28 @@ export const DownloadModule = {
             });
         },
 
+        async downloadAndInstallSpecific({state, commit, dispatch}, params: {
```
anttimaki, line 129: Add a task for thinking about the names of the methods in this module. This is perhaps best done after we've finished with other refactoring so we have a clearer picture what methods there will be.

For example, does the "Specific" suffix in this context still make sense? Or is the difference between "addDownload" and "downloadAndInstall" clear enough?

anttimaki (approved):

PR #1713, opened 2025-04-12
Switched validation to more linux compatible method that enables selection of ELF 32 Bit Binary files

AmyNicole8715:
1. Modified the file selection dialog on Linux to use "All Files (*)" filter, which allows seeing all executable files regardless of extension, including ELF binaries.
2. Enhanced the validation logic to accept both named Steam binaries and detect executable files with "steam" in their name, allowing the application to properly recognize ELF executables.
3. Added a utility method isFileExecutable() that verifies if a file exists and is a regular file.


AmyNicole8715:
Hang on I found a new error that happens due to the proton issue

AmyNicole8715:
> Hang on I found a new error that happens due to the proton issue

Ok, users who use this will need to remove the .forceproton, but thats still not 100% solved I still seem to be struggling to get it to start the modded part of the operation, it just doesn't throw errors now

AmyNicole8715:
Alright, so far it seems it should work but I cannot be certain that it is due to when my build of the fix loads it does not give me the linux instructions anymore for setting the launch parameters in Valheim, that said I hate programming so I am opting to start the game regardless will report back if I can manage to get "start modded" to properly launch it

PR #1718, opened 2025-04-16
Make throwForR2Error return the value if it'sn't an Error

VilppeRiskidev:


anttimaki (commented):
I take it there was no problems detected when testing this?

Also, could this PR contain the changes for the few places in DownloadModule that the R2Error change was intended for in the first place? Unless you've planned to include them in some other PR. I mean the places where we have pattern like:

```typescript
// From
const maybeError = await someAction();
throwForR2Error(maybeError);

// To
const notAnError = throwForR2Error(await someAction());
```

VilppeRiskidev:
> I take it there was no problems detected when testing this?
> 
> Also, could this PR contain the changes for the few places in DownloadModule that the R2Error change was intended for in the first place? Unless you've planned to include them in some other PR. I mean the places where we have pattern like:
> 
> ```ts
> // From
> const maybeError = await someAction();
> throwForR2Error(maybeError);
> 
> // To
> const notAnError = throwForR2Error(await someAction());
> ```

Good point, that's now included. I also removed the now-unnecessary `as ManifestV2[]`s

anttimaki (approved):
👍🏻

PR #1722, opened 2025-04-16
Rename DownloadModule functions

VilppeRiskidev:
* addDownload to addDownloadProgressTracking
* installMods to installMultiple
* installMod to install
* downloadAndInstallSpecific to downloadAndInstall

### DIFF
--- a/src/components/mixins/UtilityMixin.vue
+++ b/src/components/mixins/UtilityMixin.vue
@@ -39,7 +39,7 @@ export default class UtilityMixin extends Vue {
                 const profile = this.$store.getters['profile/activeProfile'].asImmutableProfile();
                 const combo = throwForR2Error(await ThunderstoreCombo.fromProtocol(protocolUrl, game));
 
-                await this.$store.dispatch('download/downloadAndInstallSpecific', {profile, combo});
+                await this.$store.dispatch('download/downloadAndInstall', {profile, combo});
 
                 const modList = throwForR2Error(await ProfileModList.getModList(profile));
                 await this.$store.dispatch('profile/updateModList', modList);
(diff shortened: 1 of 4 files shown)


VilppeRiskidev:
These are debatable, I'm not the best at naming things. 🤷

anttimaki (commented):
IDK, I think the changes are ok but on the other hand not a huge improvement. The changes can be merged or the PR closed.

I think the above is because we should've waited a bit longer to see how the other changes affect the stuff the download module actually does. I spent quite some time thinking about this while doing this review, but not far enough to be able to write a proper spec. Here's roughly the most relevant thoughts I had:

1. We got one path that goes `UtilityMixin` -> `downloadAndInstall` -> `installMultiple` -> `install`. AFAICT none of the Vuex actions are called from anywhere else
2. We got second path that goes `DownloadMixin.downloadCompletedCallback` -> `DownloadMixin.installModsAndResolveConflicts` -> `ProfileUtils.installModsToProfile`
3. I.e. we have installation logic in duplicated in two places
    - Caveat: We need to check if the installations are actually identical. I'd expect them to have minor differences. But I'd wager those differences might be merely happenstance and the installations should be identical. But this warrants more research before starting implementation
4. It would make sense that `DownloadModule` would handle only downloads, and either call `ProfileUtils` directly, or perhaps preferably so that the UI components would call the methods separately, like:
```typescript
await this.$store.dispatch('download/download', combos);
await ProfileUtils.installModsToProfile(combos, profile, [], callback);
```

Now, IDK how feasible all this would be currently (i.e. before further refactoring) or at all. Or how much effort it would require, i.e. if it makes sense given the possibility that some of the code might be soon-ish made obsolete by TCLI. But it's something that you could look into.

PR #1723, opened 2025-04-16
Add DownloadStatusEnum and use it when tracking download/install status

VilppeRiskidev:
* Instead of a boolean type field "failed", set status to DownloadStatusEnum.FAILED
* Update status accordingly when download is initiated, when installation starts and when installation finishes.

### DIFF
--- a/src/components/mixins/DownloadMixin.vue
+++ b/src/components/mixins/DownloadMixin.vue
@@ -2,6 +2,7 @@
 import Vue from 'vue';
 import Component from 'vue-class-component';
 
+import { DownloadStatusEnum } from '../../model/enums/DownloadStatusEnum';
 import R2Error, { throwForR2Error } from "../../model/errors/R2Error";
 import Game from "../../model/game/Game";
 import Profile, { ImmutableProfile } from "../../model/Profile";
@@ -41,9 +42,11 @@ export default class DownloadMixin extends Vue {
 
     async downloadCompletedCallback(downloadedMods: ThunderstoreCombo[], downloadId: number): Promise<void> {
         try {
+            this.$store.commit('download/updateDownload', {downloadId, status: DownloadStatusEnum.INSTALLING});
             await this.installModsAndResolveConflicts(downloadedMods, this.profile.asImmutableProfile(), downloadId);
+            this.$store.commit('download/updateDownload', {downloadId, status: DownloadStatusEnum.DONE});
         } catch (e) {
-            this.$store.commit('download/updateDownload', {downloadId, failed: true});
+            this.$store.commit('download/updateDownload', {downloadId, status: DownloadStatusEnum.FAILED});
             this.$store.commit('error/handleError', R2Error.fromThrownValue(e));
         }
     }
--- a/src/components/mixins/DownloadMixin.vue
+++ b/src/components/mixins/DownloadMixin.vue
@@ -2,7 +2,6 @@
 import Vue from 'vue';
 import Component from 'vue-class-component';
 
-import { DownloadStatusEnum } from '../../model/enums/DownloadStatusEnum';
 import R2Error, { throwForR2Error } from "../../model/errors/R2Error";
 import Game from "../../model/game/Game";
 import Profile, { ImmutableProfile } from "../../model/Profile";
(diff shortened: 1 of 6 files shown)


src/components/mixins/DownloadMixin.vue
```diff
@@ -41,9 +42,11 @@ export default class DownloadMixin extends Vue {
 
     async downloadCompletedCallback(downloadedMods: ThunderstoreCombo[], downloadId: number): Promise<void> {
         try {
+            this.$store.commit('download/updateDownload', {downloadId, status: DownloadStatusEnum.INSTALLING});
```
anttimaki, line 45: Thoughts on having custom mutators for these, e.g. `this.$store.commit('download/setInstalling', downloadId);`, and others for done and failed? Would be a bit shorter and wouldn't have to import the enum in all UI components.

src/model/enums/DownloadStatusEnum.ts
```diff
@@ -0,0 +1,8 @@
+export enum DownloadStatusEnum {
+    DOWNLOADING = 0,
+    INSTALLING = 1,
+    DONE = 2,
+    FAILED = 3
+}
+
+export default DownloadStatusEnum;
```
anttimaki, line 8: I consider default exports a bad practice, so I wouldn't use one here as there's no need.

src/store/modules/DownloadModule.ts
```diff
@@ -236,7 +239,7 @@ function getIndexOfDownloadProgress(allDownloads: DownloadProgress[], downloadId
 
 function getOnlyActiveDownloads(downloads: DownloadProgress[]): DownloadProgress[] {
```
anttimaki, line 240: Why not use DONE status here? Also would this be easier to read:
```typescript
function getOnlyActiveDownloads(downloads: DownloadProgress[]): DownloadProgress[] {
    const active = [DownloadStatusEnum.DOWNLOADING, DownloadStatusEnum.INSTALLING];
    return downloads.filter(dl => active.includes(dl.status));
}
```

anttimaki (changes requested):
No hard blockers per se but maybe implement at least some of the comments?

src/components/mixins/DownloadMixin.vue
VilppeRiskidev: Yess, I'll do that

src/store/modules/DownloadModule.ts
```diff
@@ -211,6 +214,30 @@ export const DownloadModule = {
                 state.allDownloads = newDownloads;
             }
         },
+        setDone(state: State, downloadId: number) {
+            const index: number = getIndexOfDownloadProgress(state.allDownloads, downloadId);
```
anttimaki, line 218: Could there be another helper function so this would become something like
`state.allDownloads = updateDownloadStatus(state.allDownloads, downloadId, DownloadStatus.DONE);`
and then use it in the other commits too to avoid the repetition?

src/components/views/UpdateAllInstalledModsModal.vue
```diff
@@ -55,7 +57,7 @@ export default class UpdateAllInstalledModsModal extends mixins(DownloadMixin)
             try {
                 if (status === StatusEnum.FAILURE) {
                     this.setIsModProgressModalOpen(false);
-                    this.$store.commit('download/updateDownload', {downloadId, failed: true});
+                    this.$store.commit('download/updateDownload', {downloadId, status: DownloadStatusEnum.FAILED});
```
anttimaki, line 60: Use the `setFailed` instead?

anttimaki (changes requested):

src/components/views/UpdateAllInstalledModsModal.vue
```diff
@@ -29,6 +30,7 @@ import ModalCard from "../ModalCard.vue";
 import DownloadMixin from "../mixins/DownloadMixin.vue";
 import * as DownloadUtils from "../../utils/DownloadUtils";
 import DownloadModVersionSelectModal from "../views/DownloadModVersionSelectModal.vue";
+import { DownloadStatusEnum } from "../../model/enums/DownloadStatusEnum";
```
anttimaki, line 33: Unnecessary import 🙈

anttimaki (approved):

src/components/views/UpdateAllInstalledModsModal.vue
VilppeRiskidev: Removed ✅

anttimaki (approved):

PR #1724, opened 2025-04-16
Thunderstore ecosystem integration

anttimaki:
Use Thunderstore ecosystem as the source of truth for game definitions.

**PR contents:**

- First commit: ethangreen-dev's earlier work rebased on top of the current develop branch
- Third commit: add tooling for updating the local ecosystem files
- Fourth commit: switch to using "live" data instead of the hardcoded files included in the first commit
- Rest of the commits change the code to use the information provided by the ecosystem, add/fix methods for mapping the ecosystem data to values used by the mod manager, and clean up implementation where it's made possible by the improved typing and more complete ecosystem data

**Mod manager TODO:**

- ~Required: Fix unit tests so the CI pipelines pass~ #1744
- ~Required: Testing. So far testing has mostly consisted of seeing that r2mm dev build process passes. No testing at all has been done with TSMM.~
- Optional: update `Adding a game.md`
- ~Optional: Rethink mapping functions. Some can be removed by making the mod manager to use the enum values used in the ecosystem data, etc. If mapping functions are still needed, they could be centralized in a single location instead of being scattered around various files~
  - Mostly done in #1751
  - PackageLoader enum in #1805
- ~Optional: There's still clean up to do and possibly typing to improve, but I didn't want to bloat this PR any further~ #1744 #1745 #1746 

**Ecosystem TODO:**

- ~Required: Ecosystem data contains outdated iconUrl for R.E.P.O.~
- ~Required: update schema to contain PackageLoader.None and latest games~
- ~Optional: Provide information for dedicated servers~
- Optional: Provide game images as URLs rather than filenames
- Optional: Consider providing package loader's RecommendedVersion information
- Optional: Provide package installer information


src/r2mm/installing/profile_installers/ModLoaderVariantRecord.ts
```diff
@@ -1,256 +1,59 @@
...
+    BONELAB: [
+        new ModLoaderPackageMapping(
+            'LavaGang-MelonLoader',
+            '',
+            PackageLoader.MELON_LOADER,
+        ),
+    ],
+}
 
-export const ALL_MODLOADER_MAPPINGS = [
```
anttimaki, line 80: Is it safe to drop these exports already?

scripts/sync.ts
```diff
@@ -1,19 +1,48 @@
 import fs from "fs";
 
+import {FetchingJSONSchemaStore, InputData, JSONSchemaInput, quicktype} from "quicktype-core";
+
+const ECOSYSTEM_DATA_URL = "https://thunderstore.io/api/experimental/schema/dev/latest/";
+const ECOSYSTEM_JSON_SCHEMA_URL = "https://thunderstore.io/api/experimental/schema/ecosystem-json-schema/latest/";
+const ECOSYSTEM_DATA_PATH = "./src/assets/data/ecosystem.json";
+const ECOSYSTEM_JSON_SCHEMA_PATH = "./src/assets/data/ecosystemJsonSchema.json";
+const ECOSYSTEM_DATA_TYPES_PATH = "./src/assets/data/ecosystem.d.ts";
```
anttimaki, line 9: I'm considering changing this from ecosystem.d.ts` to something like `ecosystemTypes.ts` so Enums can be imported instead of having to be redefined in the mod manager.

anttimaki: This is now done in https://github.com/ebkr/r2modmanPlus/pull/1751

scripts/sync.ts
```diff
@@ -1,19 +1,48 @@
...
 /**
  * This script synchronizes the in-repo ecosystem schema JSON to the latest
- * version.
+ * version and generates matching TypeScript types.
  */
-
 async function updateSchema() {
-    console.log("Updating games.json...")
-    const response = await fetch("https://thunderstore.io/api/experimental/schema/dev/latest/");
+    console.log("Updating ecosystem.json...");
```
ebkr, line 16: Nit: Do we want to keep console.log calls? Similar for the other logs

src/assets/data/ecosystem.d.ts
```diff
@@ -0,0 +1,401 @@
...
+
+export interface ThunderstoreEcosystem {
+    communities:       { [key: string]: ThunderstoreEcosyste };
+    games:             { [key: string]: Game };
+    modloaderPackages: ModloaderPackage[];
+    packageInstallers: { [key: string]: PackageInstaller };
+    schemaVersion:     string;
+}
+
+export interface ThunderstoreEcosyste {
```
ebkr, line 14: Are we concerned about the generated interface name?

src/model/game/StorePlatform.ts
```diff
@@ -7,3 +9,20 @@ export enum StorePlatform {
     XBOX_GAME_PASS = "Xbox Game Pass",
     OTHER = "Other",
 }
+
+export function getStorePlatformFromName(name: string): StorePlatform {
```
ebkr, line 13: Would we prefer to do something a little more generic?

```ts
function getEnumKeyFromEnumValue<T extends Record<string, string>>(enumObject: T, value: string): keyof T {
  const result = Object.keys(enumObject).find(key => enumObject[key] === value);
  if (!result) {
    throw new Error();
  }
  return result;
}
```

We could then do this:
```ts
const key = getEnumKeyFromEnumValue(SortDirection, "Reverse");
```

This could be a util and then means we don't need to reimplement it for every record type.

src/model/schema/ThunderstoreSchema.ts
```diff
@@ -0,0 +1,49 @@
...
+import R2Error from "../errors/R2Error";
+
+
+export class EcosystemSchema {
+    private static _isValidated: boolean = false;
+
+    /**
+     * Get a validated instance of the ecosystem schema.
+     */
+    private static get get(): ThunderstoreEcosystem {
```
ebkr: Not a fan of this name, what is it getting?
If we rename this to: `thunderstoreEcosystem` so it becomes `this.thunderstoreEcosystem`

src/model/schema/ThunderstoreSchema.ts
```diff
@@ -0,0 +1,49 @@
...
+
+        this._isValidated = true;
+        return ecosystem as ThunderstoreEcosystem;
+    }
+    /**
+     * Get a list of r2modman entries i.e. games supported by the mod manager.
+     */
+    static get supportedGames() {
+        return Object.values(this.get.games).filter(
+            (x): x is typeof x & {r2modman: NonNullable<typeof x["r2modman"]>} => x.r2modman != null
```
ebkr: Can we make this easier to read?

scripts/sync.ts
anttimaki: I kind of like them as progress indicators when running the script.

src/assets/data/ecosystem.d.ts
anttimaki: Would be nice to have a better name for this, but IDK how to change it with quicktype, and I think the right place for the change would be the ecosystem schema repo anyway.

If we actually start using this type on r2mm, we can mitigate the issue by renaming when it's [re-exported here](https://github.com/ebkr/r2modmanPlus/pull/1751/files#diff-42d08a9764b57b0ca2942ed552da05dd1fb87aad9070ed990e7ddf295e550b48R11).

src/model/game/StorePlatform.ts
anttimaki: Looks like a nice solution but I think it becomes obsolete when #1751 kicks in.

src/model/schema/ThunderstoreSchema.ts
anttimaki: The naming is leftover from when the method was public, so it read `ThunderstoreSchema.get`. I'm fine renaming it but I'll probably do it in a separate PR.

src/model/schema/ThunderstoreSchema.ts
anttimaki: Yes.

src/model/schema/ThunderstoreSchema.ts
anttimaki: Had to rebase the whole stack anyway so ended up doing the changes in this PR instead. `get` was renamed to shorter `ecosystem` [here](https://github.com/ebkr/r2modmanPlus/pull/1724/files#diff-42d08a9764b57b0ca2942ed552da05dd1fb87aad9070ed990e7ddf295e550b48R16) so usage now reads like `this.ecosystem.games`.

src/model/schema/ThunderstoreSchema.ts
anttimaki: Based on changes in a latter commit, swapped the order of `filter` and `flatMap` and [made the typing IMO easier to read](https://github.com/ebkr/r2modmanPlus/pull/1724/files#diff-42d08a9764b57b0ca2942ed552da05dd1fb87aad9070ed990e7ddf295e550b48R39-R45). I could have imported the R2Modman type from ecosystem type definitions, but I think using `NonNullable` makes the intent clearer, even if it's a bit more verbose.

src/model/schema/ThunderstoreSchema.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /**
+     * Get a list of r2modman entries i.e. games supported by the mod manager.
+     */
+    static get supportedGames() {
+        return Object.values(this.ecosystem.games).flatMap(
+            (game) => game.r2modman
+        ).filter(
+            (r2modman): r2modman is NonNullable<typeof r2modman> => r2modman != null
```
ebkr, line 43: Why is the type declaration like this rather than:
```ts
.filter(value => value != null);
```
Why do we need an assuming cast for NonNullable when we do a null assertion as part of the filter anyway?

anttimaki: The type predicate is needed to narrow the type because TypeScript's type inference does not handle it automatically. I'm not entirely sure WHY, but I assume it's because any function passed to `.filter()` needs to return a boolean-ish value, so we need to tell TS how to infer the type based on that boolean value. Using the function you suggested wouldn't change the type, so the returned value would still be `(R2Modman | null)[]`.

Apparently the following could be used to have proper type inferring, but IMO it's a bit hacky and makes the intent less clear (it relies on `.flatMap()` filtering out the empty arrays.

```typescript
return Object.values(this.ecosystem.games).flatMap(
    (game) => game.r2modman ? game.r2modman : []
);
```

PR #1730, opened 2025-04-22
Add "id" prop to ModalCard and define ids where ModalCard is used

VilppeRiskidev:


src/components/profiles-modals/ImportProfileModal.vue
```diff
@@ -274,7 +274,7 @@ export default class ImportProfileModal extends mixins(ProfilesMixin) {
 </script>
 
 <template>
-    <ModalCard v-if="activeStep === 'FILE_CODE_SELECTION'" key="FILE_CODE_SELECTION" :is-active="isOpen" @close-modal="closeModal">
+    <ModalCard id="import-profile-from-file-or-code-modal" v-if="activeStep === 'FILE_CODE_SELECTION'" key="FILE_CODE_SELECTION" :is-active="isOpen" @close-modal="closeModal">
```
anttimaki, line 277: Nitpick: wouldn't have minded if all the modals in this file would've used the same `import-profile` or `profile-import` prefix consistently.

anttimaki (approved):
Another nitpick: the commit message could also tell WHY the change was done, as now it's not all that obvious.

PR #1735, opened 2025-04-24
WIP: Generic downloader

VilppeRiskidev:


src/components/mixins/UtilityMixin.vue
```diff
@@ -39,7 +44,32 @@ export default class UtilityMixin extends Vue {
                 const profile = this.$store.getters['profile/activeProfile'].asImmutableProfile();
                 const combo = throwForR2Error(await ThunderstoreCombo.fromProtocol(protocolUrl, game));
 
-                await this.$store.dispatch('download/downloadAndInstallSpecific', {profile, combo});
+                const downloadId = await this.$store.dispatch(
```
anttimaki, line 47: UtilityMixin is still a Vue component so I consider this a step backwards in our efforts to remove the business logic from the UI components. Keeping the Vuex between the UI and logic would make sense to me.

src/providers/ror2/downloading/ThunderstoreDownloaderProvider.ts
```diff
@@ -45,7 +45,15 @@ export default abstract class ThunderstoreDownloaderProvider {
                                callback: (progress: number, modName: string, status: number, err: R2Error | null) => void,
                                completedCallback: (modList: ThunderstoreCombo[]) => void): void;
 
-    /**
+
+    public abstract downloadCombos(
```
anttimaki, line 49: If `combos` passed in and the return value out of the function are actually the same array, then passing it around is IMO detrimental for clarity.

```typescript
// Makes me think :(
const combosToDownload = getDependencies(combo);
const combosToInstall = await download(combosToDownload);
await install(combosToInstall );

// Smooth brain approved ^_^
const combos = getDependencies(combo);
await download(combos);
await install(combos);
```

src/utils/DownloadUtils.ts
```diff
@@ -21,4 +28,66 @@ export function addSolutionsToError(err: R2Error): void {
 export function generateProgressPercentage(progress: number, currentIndex: number, total: number): number {
     const completedProgress = (currentIndex / total) * 100;
     return completedProgress + (progress * 1/total);
-}
\ No newline at end of file
+}
+
```
anttimaki, line 32: Maybe ponder a bit if placing dependency resolution stuff in DownloadUtils makes sense. Pedantically it's a separate issue from downloading, but sometimes there might be practical limitations that makes being pedantic a bad thing.

src/utils/DownloadUtils.ts
```diff
@@ -21,4 +28,66 @@ export function addSolutionsToError(err: R2Error): void {
...
+
+export async function getFullDependencyList(
+    profile: ImmutableProfile,
+    combos: ThunderstoreCombo[],
+): Promise<ThunderstoreCombo[]> {
+    const modList = await ProfileModList.getModList(profile);
```
anttimaki, line 38: I don't think this has anything to do with the actual dependency resolution and shouldn't be here.

src/utils/DownloadUtils.ts
```diff
@@ -21,4 +28,66 @@ export function addSolutionsToError(err: R2Error): void {
...
+    if (modList instanceof R2Error) {
+        throw modList;
+    }
+
+    const allModsToDownload: ThunderstoreCombo[] = [...combos];
+
+    for (const combo of combos) {
+        const isModpack = combo.getMod().getCategories().find(value => value === "Modpacks") !== undefined;
+        const versionMode = isModpack
```
anttimaki, line 47: Are you certain we can determine the builder mode internally in this method without external arguments and still make this a one-size-fits-all solution? Consider profile imports vs. update all vs. single mod install.

If it can be done (and might make sense even if not), then maybe `DependencySetBuilderMode` should reside in the same file instead of being imported from the download provider.

src/utils/DownloadUtils.ts
```diff
@@ -21,4 +28,66 @@ export function addSolutionsToError(err: R2Error): void {
...
+}
+
+
+
+async function buildDependencySet(
+    mod: ThunderstoreVersion,
+    builder: ThunderstoreCombo[],
+    mode: DependencySetBuilderMode): Promise<void>
+{
+    const game = GameManager.activeGame;
```
anttimaki, line 76: Would be a nice bonus if the game would be passed as an argument rather than depending on the GameManager, but that's not necessarily a matter for this PR.

anttimaki (commented):
1. Disclaimer: the comments are based purely on reading the code, I did no testing at all
2. Scope-wise having the PR just separate the dependency resolution from the download provider would be nice
3. _"None of those implementations are really specific to overwolf vs. r2mm, so I don't see why not just use move those to a shared location like DownloadUtils?"_ For dependency resolution I agree. For download stuff it's not that clear cut IMO. Consider that the idea of a "provider" is to be a swappable standalone plugin, and therefore the code repetition is intentional in this case. There might be parts that could reside outside the providers, e.g. if they're very UI related, but I think each of those cases should be considered separately.

VilppeRiskidev:
Still a couple of things to clean up and check (names, return types, error handling) and then testing. I've not encountered anything broken yet so at least these work in the most normal case:
- downloading a **specific** version of a single mod (downloads **latest** versions of dependencies)
- downloading a modpack (downloads correct **specific** versions of dependencies, replacing a possible previous install)
- importing a profile (downloads the correct **specific** versions of the mods)
- updating all mods (updates everything correctly)
- download tracker correclty shows only the main initial mod(s), not all of the dependencies, (as the title)
- I've tested some error cases which all seem to work normally, but I still need to do more.

src/components/mixins/UtilityMixin.vue
```diff
@@ -37,9 +37,10 @@ export default class UtilityMixin extends Vue {
             try {
                 const game = this.$store.state.activeGame;
                 const profile = this.$store.getters['profile/activeProfile'].asImmutableProfile();
-                const combo = throwForR2Error(await ThunderstoreCombo.fromProtocol(protocolUrl, game));
+                const combos = [throwForR2Error(await ThunderstoreCombo.fromProtocol(protocolUrl, game))];
+                const downloadId = await this.$store.dispatch('download/addDownload', combos);
 
-                await this.$store.dispatch('download/downloadAndInstallSpecific', {profile, combo});
+                this.$store.dispatch('download/downloadAndInstallCombos', {combos, profile, downloadId});
```
anttimaki, line 43: Why remove the `await`? Wouldn't this suggest that `modList` below is loaded before the installation might have finished?

src/components/views/DownloadModModal.vue
```diff
@@ -72,37 +70,13 @@ import ThunderstoreDownloaderProvider from '../../providers/ror2/downloading/Thu
         async downloadHandler(tsMod: ThunderstoreMod, tsVersion: ThunderstoreVersion) {
```
anttimaki, line 70: Is there any reason for this to be defined in `DownloadModModal` anymore, instead of placing it directly in `DownloadModVersionSelectModal`. Is there any reason for `DownloadModVersionSelectModal` and `UpdateAllInstalledModsModal` to be subcomponents of this component, for that matter?

Just pointing this out, it can and probably should be handled in a separate PR.

src/components/views/DownloadModModal.vue
```diff
@@ -72,37 +70,13 @@ import ThunderstoreDownloaderProvider from '../../providers/ror2/downloading/Thu
...
-                } catch (e) {
-                    this.setIsModProgressModalOpen(false);
-                    this.$store.commit('download/updateDownload', { downloadId, failed: true });
-                    this.$store.commit('error/handleError', R2Error.fromThrownValue(e));
-                    return;
-                }
-                await this.downloadCompletedCallback(downloadedMods, downloadId);
-                this.setIsModProgressModalOpen(false);
-            }, 1);
+            this.$store.dispatch('download/downloadAndInstallCombos', {combos, profile: this.profile.asImmutableProfile(), downloadId})
```
anttimaki, line 79: Await?

src/components/views/UpdateAllInstalledModsModal.vue
```diff
@@ -27,49 +27,21 @@ import { Component } from "vue-property-decorator";
...
-                    this.$store.commit('download/updateDownload', {downloadId, modName, downloadProgress});
-                }
-            } catch (e) {
-                this.$store.commit('error/handleError', R2Error.fromThrownValue(e));
-            }
-        }, async (downloadedMods) => {
-            await this.downloadCompletedCallback(downloadedMods, downloadId);
-            this.setIsModProgressModalOpen(false);
-        });
+        this.$store.dispatch('download/downloadAndInstallCombos', {combos, profile: this.profile.asImmutableProfile(), downloadId})
```
anttimaki, line 44: Await?

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -1,117 +1,36 @@
...
-            throw modList;
-        }
-
-        const dependencies = await this.getDependenciesWithCorrectVersions(combo, modList);
-        const allModsToDownload = [combo, ...dependencies];
-
-        let modInProgressName = combo.getMod().getName();
+        let modInProgressName = combos[0].getMod().getName();
         let downloadCount = 0;
-        // Mark the mod 80% processed when the download completes, save the remaining 20% for extracting.
```
anttimaki, line 104: Was this removed intentionally? I kinda like it as it explains the reason for magic number `0.8` below.

src/store/modules/DownloadModule.ts
```diff
@@ -50,11 +52,11 @@ export const DownloadModule = {
     }),
 
     actions: <ActionTree<State, RootState>>{
-        addDownload({state}, initialMods: string[]): UUID {
+        addDownload({state}, combos: ThunderstoreCombo[]): UUID {
```
anttimaki, line 55: After the latest refactors it seems that having the Vue components call this action is unnecessary, as they only passes the received download ID back to `downloadAndInstallCombos`. Vue could simply call `downloadAndInstallCombos` which instead would call this action internally. We could even consider prefixing the name of this action with `_` to indicate it's not intended to be called directly.

src/store/modules/DownloadModule.ts
```diff
@@ -125,28 +127,67 @@ export const DownloadModule = {
...
+                    (downloadProgress, modName, status, err) => {
+                        dispatch('downloadProgressCallback', { downloadId: params.downloadId, downloadProgress, modName, status, err });
                     }
                 );
-                await dispatch('installMods', {downloadedMods, downloadId, profile: params.profile});
             } catch (e) {
-                commit('updateDownload', { downloadId, failed: true });
-                throw e;
+                commit('updateDownload', { downloadId: params.downloadId, failed: true });
+                commit('error/handleError', R2Error.fromThrownValue(e), { root: true });
```
anttimaki, line 144: Why handle the error hear instead of throwing it? Now it looks like `downloadAndInstallCombos` doesn't know if this call fails and will continue to the install step even if the mods weren't downloaded.

src/store/modules/DownloadModule.ts
```diff
@@ -125,28 +127,67 @@ export const DownloadModule = {
...
             }
         },
 
+        downloadAndInstallCombos({commit, dispatch}, params: {
+            combos: ThunderstoreCombo[],
+            profile: ImmutableProfile,
+            downloadId: UUID
+        }) {
+            setTimeout(async () => {
```
anttimaki, line 153: Nitpick: doing `const {combos, profile, downloadId} = params;` would somewhat avoid repetition.

Also I think I've asked this before in some context but as the location of the code has changed again: is the setTimeout still necessary?

src/store/modules/DownloadModule.ts
```diff
@@ -125,28 +127,67 @@ export const DownloadModule = {
...
+                try {
+                    await getFullDependencyList(params.combos);
+                    await dispatch('download', {combos: params.combos, downloadId: params.downloadId})
+                    await dispatch('installModsAndResolveConflicts', {combos: params.combos, profile: params.profile, downloadId: params.downloadId});
+                    commit('setIsModProgressModalOpen', false);
```
anttimaki, line 158: Nitpick: this could be done in `finally` block.

src/store/modules/DownloadModule.ts
```diff
@@ -125,28 +127,67 @@ export const DownloadModule = {
...
+                    return;
```
anttimaki, line 159: Unnecessary `return` (in `catch` too).

src/utils/DependencyUtils.ts
```diff
@@ -11,3 +21,49 @@ export const splitToNameAndVersion = (dependencyString: string): [string, string
     const fullName = parts.join('-');
     return [fullName, version];
 };
+
+export async function getFullDependencyList(combos: ThunderstoreCombo[]) {
```
anttimaki, line 25: IDK how much effort it would require to get all necessary things mocked, but having unit tests for this method would be great. It should cover at least the use cases of installing a mod, installing a mod pack, updating all for profile containing both mods and modpacks, and importing a profile containing mods and mod packs.

The test coverage overall is poor, but in a refactor like this, having those tests in place would be great.

anttimaki (commented):
- The download provider seems to clean up nicely
- Some of my earlier comments haven't been addressed (or communicated as wontfixes)
- I still didn't do any manual testing, LMK when this is ready for that

src/utils/DownloadUtils.ts
anttimaki: I meant this shouldn't be done as part of the dependency resolution. Now it seems it was dropped altogether. I'm not sure why it's here, but I would wary of just dropping it.

src/utils/DownloadUtils.ts
anttimaki: Do you think this comment is resolved?

src/utils/DownloadUtils.ts
anttimaki: Any thoughts on this?

src/utils/DownloadUtils.ts
VilppeRiskidev: I made an archaelogical trip to the code and added this filtering stuff  back in. It is now in `getFullDependencyList` (in `DependencyUtils`).
I think there might be some sort of weird edge-case behaviour if you install multiple mods and modpacks with same mods (but different versions) as their dependencies, but I also think that was already the case before.

VilppeRiskidev: the modlist is only used to filter out already installed mods, so I think getting the mod list in `getFullDependencyList` is ok. The list could also be passed as a parameter if you think that'd be better.

src/utils/DownloadUtils.ts
VilppeRiskidev: The second part is resolved.
For the first part: At least to my testing, the builder works correctly, but unit testing could prove me wrong.

The whole enum could even be omitted, **IF** it turns out the mode only needs to be used once (or twice) and therefore is just same as checking `isModPack`.

src/utils/DownloadUtils.ts
VilppeRiskidev: I changed it to be passed as an argument, not too big of a change.
_Whats a few more to already a huge pile of changes xD._

src/components/mixins/UtilityMixin.vue
VilppeRiskidev: I Added await to the usages.

src/components/views/UpdateAllInstalledModsModal.vue
VilppeRiskidev: I Added await to the usages.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
VilppeRiskidev: I agree, I added this back 👍

src/store/modules/DownloadModule.ts
VilppeRiskidev: I changed it so that addDownload is only called internally and prefixed it with an underscore.
I also added the underscore to other similar functions.

src/store/modules/DownloadModule.ts
VilppeRiskidev: I changed it back to `throw`.

src/store/modules/DownloadModule.ts
VilppeRiskidev: I removed the `setTimeout` and used that params deconstructing in a few places.

src/utils/DependencyUtils.ts
VilppeRiskidev: I'll look into this, I'll create another PR with the tests and chain it to this one.

src/store/modules/DownloadModule.ts
```diff
@@ -70,99 +71,86 @@ export const DownloadModule = {
...
-                        installProgress: DownloadUtils.generateProgressPercentage(100, currentDownloadIndex, params.downloadedMods.length)
-                    });
-                    currentDownloadIndex++;
-                }
+            const { combos, game, profile } = params;
 
-                const modList = throwForR2Error(await ProfileModList.getModList(params.profile));
-                throwForR2Error(await ConflictManagementProvider.instance.resolveConflicts(modList, params.profile));
-            });
+            const downloadId = await dispatch('_addDownload', combos);
```
anttimaki, line 81: Minor: perhaps this too should be in the `try` block so it gets error handled like the rest? That would mean the catch-block needs to check if `downloadId` is set, but that sound like a fair tradeoff.

anttimaki (changes requested):
See (updated) previous comments about changes to `getFullDependencyList`.

src/utils/DownloadUtils.ts
anttimaki: pr-review

VilppeRiskidev:
Added another commit that fixed the filtering (I noticed it being broken by developing the unit tests)

src/store/modules/DownloadModule.ts
```diff
@@ -1,22 +1,24 @@
...
 import ConflictManagementProvider from "../../providers/generic/installing/ConflictManagementProvider";
-import ProfileInstallerProvider from "../../providers/ror2/installing/ProfileInstallerProvider";
 import ThunderstoreDownloaderProvider from "../../providers/ror2/downloading/ThunderstoreDownloaderProvider";
 import ManagerSettings from "../../r2mm/manager/ManagerSettings";
 import ProfileModList from "../../r2mm/mods/ProfileModList";
 import { State as RootState } from "../../store";
 import * as DownloadUtils from "../../utils/DownloadUtils";
+import { getFullDependencyList } from "../../utils/DependencyUtils";
+import { installModsToProfile } from "../../utils/ProfileUtils";
+import ManifestV2 from "src/model/ManifestV2";
```
anttimaki, line 17: Use relative import.

src/store/modules/DownloadModule.ts
```diff
@@ -70,99 +72,90 @@ export const DownloadModule = {
...
-                }
+            const { combos, game, profile } = params;
+            let downloadId: UUID | undefined;
 
-                const modList = throwForR2Error(await ProfileModList.getModList(params.profile));
-                throwForR2Error(await ConflictManagementProvider.instance.resolveConflicts(modList, params.profile));
-            });
+            try {
+                downloadId = await dispatch('_addDownload', combos);
+                const modList = await throwForR2Error(ProfileModList.getModList(profile));
```
anttimaki, line 85: - I'd change `modList` to `installedMods` for clarity
- Put the `await` inside the `throwForError`, the promise returned by `getModList` will not ever be thrown
  - This also removes the need to do `as ManifestV2[]` on the next line

src/utils/DependencyUtils.ts
```diff
@@ -11,3 +22,54 @@ export const splitToNameAndVersion = (dependencyString: string): [string, string
...
+    mod: ThunderstoreVersion,
+    game: Game,
+    builder: ThunderstoreCombo[],
+    mode: DependencySetBuilderMode,
+    installedMods: ManifestV2[]): Promise<void>
+{
+    const useLatestVersion = Boolean(mode);
+    let foundDependencies = await getCombosByDependencyStrings(game, mod.getDependencies(), useLatestVersion);
+
+    // Filter out already added AND installed mods AFTER reading packages from the DB to
```
anttimaki, line 59: > fixed the filtering (I noticed it being broken by developing the unit tests)

Has this been broken by a change you or I have done recently? If so, can you show when/where that has happened?

If it's something that has "always" been "broken", I'd beware of changing it lightly even if the new behaviour would be more correct than the old one. Rationale is that someone is always depending on an "undocumented feature" and fixing it will break things for them. In this scenario I think doing the change warrants a separate discussion and this PR shouldn't be blocked by that discussion or its outcome.

anttimaki (changes requested):

src/utils/DependencyUtils.ts
VilppeRiskidev: I somewhat reverted this change, now it works as it did before.

When it goes through the list, it does first come first served for the dependency versions, meaning:
- If a mod pack precedes a mod in a list, it will use the specific mod pack's dependency version.
- If a mod precedes a mod pack, it will use the latest version for the dependency.

VilppeRiskidev:
I decided to prefix the enum's name with ModPack (`ModPackDependencySetBuilderMode`) and add a comment to the enum. It's a little bit different than what we discussed. Also I didn't make it a required parameter, as I think it is clear enough what it does for if/when it is needed. default behaviour is _default_ behaviour, so there should be a _default_ value, right? IDK, LMK if theres some changes needed to the enum.

src/components/views/UpdateAllInstalledModsModal.vue
```diff
@@ -28,49 +28,21 @@ import { Component } from "vue-property-decorator";
...
-                    this.$store.commit('download/updateDownload', {downloadId, modName, downloadProgress});
-                }
-            } catch (e) {
-                this.$store.commit('error/handleError', R2Error.fromThrownValue(e));
-            }
-        }, async (downloadedMods) => {
-            await this.downloadCompletedCallback(downloadedMods, downloadId);
-            this.setIsModProgressModalOpen(false);
-        });
+        await this.$store.dispatch('download/downloadAndInstallCombos', {combos, profile: this.profile.asImmutableProfile(), game: this.activeGame, modPackDependencySetBuilderMode: ModPackDependencySetBuilderMode.USE_LATEST_VERSION})
```
anttimaki, line 45: This line is too long, and it's missing a semicolon.

src/utils/DependencyUtils.ts
```diff
@@ -1,3 +1,21 @@
...
+import { getCombosByDependencyStrings } from '../r2mm/manager/PackageDexieStore';
+
+/**
+ * ModPackDependencySetBuilderMode controls how dependencies are resolved for mod packs:
+ * - USE_EXACT_VERSION: Use exactly the version specified in the mod pack. (This is the default behaviour).
+ * - USE_LATEST_VERSION: Use the latest version instead
+ *
+ * For regular mods (non-modpacks), the latest version of dependencies is always used.
+ */
+export enum ModPackDependencySetBuilderMode {
```
anttimaki, line 14: Would `ModpackDependencyStrategy` be a better name? "Builder" is an internal term of the helper function used by `getFullDependencyList`. It means or tells me nothing when used on `UpdateAllInstalledModsModal` call site. It's allso a bit shorter.

src/utils/DependencyUtils.ts
```diff
@@ -11,3 +29,62 @@ export const splitToNameAndVersion = (dependencyString: string): [string, string
     const fullName = parts.join('-');
     return [fullName, version];
 };
+
+export async function getFullDependencyList(
+    combos: ThunderstoreCombo[],
+    game: Game,
+    installedMods: ManifestV2[],
+    modPackDependencySetBuilderMode: ModPackDependencySetBuilderMode = ModPackDependencySetBuilderMode.USE_EXACT_VERSION
```
anttimaki, line 37: > Also I didn't make it a required parameter, as I think it is clear enough what it does for if/when it is needed. default behaviour is default behaviour, so there should be a default value, right?

Hopefully I'm not putting words into ebkr's mouth, but the argument he has made was along the lines that explicitly having to set the argument on the call site makes that part of code easier to understand. I.e.

```typescript
getFullDependencyList(combos, game, installedMods);
```

does not tell me it's possible to affect the dependency resolution without checking the implementation of the function (which may be cumbersome to do if reading the code outside an IDE). Also, if a few lines later I see

```typescript
getFullDependencyList(combos, game, installedMods, ModPackDependencySetBuilderMode.USE_LATEST_VERSION);
```

I can't tell if this does the same thing as the first function call, or how the results of the first function call might differ from the results of the second one.

I agree that these points are valid, but I see it as a tradeoff and think one should consider the readability of the code as whole. E.g. defining (multiple) default parameters for all calls just because one use case uses different values is a lot of boilerplate and enum imports. I.e. default values aren't inherently bad, as long as they're not abused.

In this particular case I don't have strong opinion one way or another.

src/utils/DependencyUtils.ts
```diff
@@ -11,3 +29,62 @@ export const splitToNameAndVersion = (dependencyString: string): [string, string
...
+) {
+    const originalCombos = [...combos];
+    for (const combo of originalCombos) {
+        const isModpack = combo.getMod().getCategories().some(value => value === "Modpacks");
+        const versionMode = isModpack ? modPackDependencySetBuilderMode : ModPackDependencySetBuilderMode.USE_LATEST_VERSION;
+
+        await buildDependencySet(combo.getVersion(), game, combos, versionMode);
+
+        // Only filter if this combo is not a modpack and there are some installed mods
```
anttimaki, line 46: Yes, the next line tells me as much. What this comment should tell me is WHY we want this behaviour, especially WHY mods and modpacks are treated differently.

src/utils/DependencyUtils.ts
```diff
@@ -11,3 +29,62 @@ export const splitToNameAndVersion = (dependencyString: string): [string, string
...
+        if (!isModpack && installedMods.length > 0) {
+            const filteredCombos = combos.filter((dep) => 
+                !installedMods.some(installed => installed.getName() === dep.getMod().getFullName())
+            );
+            combos.length = 0;
+            combos.push(...filteredCombos);
```
anttimaki, line 52: Wait, has this always been here? What's the idea of resetting `combos` to empty array on every iteration? We lose all the work done so far?

anttimaki (changes requested):
I'll defer reviewing the unit test PR  until the issues here are addressed.

src/utils/DependencyUtils.ts
```diff
@@ -11,3 +29,65 @@ export const splitToNameAndVersion = (dependencyString: string): [string, string
...
+    for (const combo of originalCombos) {
+        const isModpack = combo.getMod().getCategories().some(value => value === "Modpacks");
+        const dependencyVersionStrategy = isModpack ? modpackDependencyStrategy : ModpackDependencyStrategy.USE_LATEST_VERSION;
+
+        await buildDependencySet(combo.getVersion(), game, combos, dependencyVersionStrategy);
+
+        // By default this filters out mods that are already installed.
+        // However, in the case of modpacks we don't want to filter them out,
+        // because modpacks (by default*) want the dependencies at specific versions.
+        // *(unless the modpackDependencyStrategy is set to USE_LATEST_VERSION)
```
anttimaki, line 49: Uh, this last line confuses me. If things are as the comment suggests, should the if-block then check `dependencyVersionStrategy` rather than `!isModPack`?

src/utils/DependencyUtils.ts
anttimaki: This comment is unaddressed.

anttimaki (changes requested):

src/utils/DependencyUtils.ts
VilppeRiskidev: The comment should now be very descriptive

src/utils/DependencyUtils.ts
VilppeRiskidev: Oops...
So yeah after scratching my head for hours and hours, I finally figured out why the filtering didn't work correctly.
Now the filtering logic is fixed and thus the `.length = 0` wasn't needed anymore.

src/utils/DependencyUtils.ts
VilppeRiskidev: This filtering is now explained in the big beautiful comment ™️  above the function.

src/utils/DependencyUtils.ts
```diff
@@ -11,3 +29,71 @@ export const splitToNameAndVersion = (dependencyString: string): [string, string
...
+
+/**
+ * - When installing a regular mods, always install the latest versions of dependencies.
+ * - When installing a mod pack we usually want to ensure compatibility by installing the exact
+ *      versions of dependencies, even if it means upgrading/downgrading already installed mods.
+ *   - If, however, modpackDependencyStrategy is set to USE_LATEST_VERSION, we install the latest
+ *      versions of dependencies for mod packs too.
+ * - When installing latest versions of dependencies, already installed mods are filtered out.
+ * - The order of the input combos array matters, combos earlier on the array get their dependencies'
+ *      versions prioritised over later ones.
```
anttimaki, line 41: I think the second bullet point and its sub bullet point together are misleading. The first one says "even if it means upgrading/downgrading already installed mods", and the second one "if ... USE_LATEST_VERSION, we install the latest versions ... for mod packs too." However, if `USE_LATEST_VERSION` is used, already installed mods are ignored in dependency resolution. As much is stated in the third bullet point, but reader can't know which one takes precedence.

The last bullet point is also a bit confusing, as per the current usage when installing a single package there's only one item in `combos`, and when updating all, all dependencies resolve to the latest version anyway.

I started to write what I considered an improved version of the doc string, but that ended up giving me an idea how the problems on the code should be addressed (see the new doc string in the added commits) 🦆

src/utils/DependencyUtils.ts
```diff
@@ -11,3 +29,71 @@ export const splitToNameAndVersion = (dependencyString: string): [string, string
...
+ */
+export async function getFullDependencyList(
+    combos: ThunderstoreCombo[],
+    game: Game,
+    installedMods: ManifestV2[],
+    modpackDependencyStrategy: ModpackDependencyStrategy
+) {
+    const originalCombos = [...combos];
```
anttimaki, line 49: I think there's no need for this method to mutate the `combos` argument, it can as well return a new array which makes it easier to grasp what's going on in the call site. (And functions generally should have side effects only if there's a good reason for it).

src/utils/DependencyUtils.ts
```diff
@@ -11,3 +29,71 @@ export const splitToNameAndVersion = (dependencyString: string): [string, string
...
+    for (const combo of originalCombos) {
+        const isModpack = combo.getMod().getCategories().some(value => value === "Modpacks");
+        const dependencyVersionStrategy = isModpack ? modpackDependencyStrategy : ModpackDependencyStrategy.USE_LATEST_VERSION;
+
+        let dependencies: ThunderstoreCombo[] = [];
+        await buildDependencySet(combo.getVersion(), game, dependencies, dependencyVersionStrategy);
+
+        combos.push(...dependencies.filter(dependency => {
+            const isNotAlreadyFound = !combos.some(alreadyAdded => alreadyAdded.getMod().getFullName() === dependency.getMod().getFullName());
```
anttimaki, line 58: The original implementation of "install single package" downloader doesn't actually do this check against an array that contains the "main" package itself. So the subtle difference here is that in the original implementation, in a case of circular dependencies, the result might contain the same package twice. Since I consider this a very rare/speculative case, and the change basically a bug fix, I think the change should be included even if it doesn't match the original implementation.

src/utils/DependencyUtils.ts
```diff
@@ -11,3 +29,71 @@ export const splitToNameAndVersion = (dependencyString: string): [string, string
...
+            const isNotInstalled = !installedMods.some(installed => installed.getName() === dependency.getMod().getFullName());
```
anttimaki, line 59: The original implementation did this filtering only when installing a single package that is not a modpack. While `||dependencyVersionStrategy === ModpackDependencyStrategy.USE_EXACT_VERSION)` check meant this filtering was not done when installing a single modpack, it was still done when updating all, which differs from the original implementation.

This change was masked by the fact that if a installed package had updates, the latest version was already included in the `combos` argument and it didn't matter that the package-as-a-dependency got filtered out. However, the change created a footgun where changing how the `combos` argument was gathered might have this work in unexpected way. I guess this unit test code explains it better than the sentence above:

```typescript
    describe("Filtering already installed mods when updating all to latest version", () => {
        it("works properly only when all mods with updates are included in the argument", async () => {
            const dependencyV2 = addMockPackage({
                name: "dependency",
                versions: [
                    {version: "2.0.0"},
                    {version: "1.0.0"}
                ]
            });
            const dependantV2 = addMockPackage({
                name: "dependant",
                dependencies: ["author-mod-2.0.0"],
                versions: [
                    {version: "2.0.0", dependencies: ["author-dependency-2.0.0"]},
                    {version: "1.0.0", dependencies: ["author-dependency-1.0.0"]}
                ]
            });

            // Old version of dependency is installed.
            const installedDependency = new ManifestV2().fromThunderstoreCombo(dependencyV2);
            installedDependency.setVersionNumber(new VersionNumber("1.0.0"));
            const installedMods = [installedDependency];

            // Call with "install single mod" use case.
            let combos = [dependantV2];
            await getFullDependencyList(combos, game, installedMods, ModpackDependencyStrategy.USE_LATEST_VERSION);

            // The requested version of dependant should be downloaded.
            // The dependency should not be downloaded as older version is already installed.
            expectToContainOnly(combos, [
                dependantV2.getDependencyString()
            ]);

            // Old version of both mods are installed.
            const installedDependant = new ManifestV2().fromThunderstoreCombo(dependantV2);
            installedDependant.setVersionNumber(new VersionNumber("1.0.0"));
            installedMods.unshift(installedDependant);

            // When calling with "update all" use case, `combos` already contains
            // the latest versions of all installed mods.
            combos = [dependencyV2, dependantV2];
            await getFullDependencyList(combos, game, installedMods, ModpackDependencyStrategy.USE_LATEST_VERSION);

            // Seems to work as expected. Both mods are actually filtered out but it
            // doesn't matter as they are both included in `combos` argument to begin with.
            expectToContainOnly(combos, [
                dependencyV2.getDependencyString(),
                dependantV2.getDependencyString()
            ]);

            // Would someone happen to call "update all" case with partial `combos`...
            combos = [dependantV2];
            await getFullDependencyList(combos, game, installedMods, ModpackDependencyStrategy.USE_LATEST_VERSION);

            // Error: doesn't contain the dependency at all as installed mods are filtered out.
            // In other words, differentiating between "install a mod" and "update all mods" is
            // done with `combos` argument, which is a bit of a footgun.
            expectToContainOnly(combos, [
                dependencyV2.getDependencyString(),
                dependantV2.getDependencyString()
            ]);
        });
```

src/utils/DependencyUtils.ts
```diff
@@ -11,3 +29,71 @@ export const splitToNameAndVersion = (dependencyString: string): [string, string
...
+
+            return isNotAlreadyFound && (isNotInstalled || dependencyVersionStrategy === ModpackDependencyStrategy.USE_EXACT_VERSION);
```
anttimaki, line 61: This combination of things is a bit hard to grasp and e.g. maintain. That, and the subtle changes in the implementation is probably due to the previous implementation having two separate methods that a bit arbitrary did things differently. Combining them into one function can easily go awry. While spending way too much time thinking about this, I think I came up with a solution that is a bit easier to understand and maintain while being closer to the original functionality.

(Another solution would've been to split "install one" and "update all" to separate functions but then we'd be pretty close to square 1 again.)

src/utils/DependencyUtils.ts
```diff
@@ -11,3 +34,100 @@ export const splitToNameAndVersion = (dependencyString: string): [string, string
     const fullName = parts.join('-');
     return [fullName, version];
 };
+
+/**
+ * Dependency resolution varies based on whether packages are being
+ * installed or updated, and whether the package is a mod or a modpack.
+ *
+ * 1. When installing a mod
```
VilppeRiskidev, line 42: This referring to a singular mod and not multiple is a bit misleading. While the current usage in the app is only either a singular mod, or updating multiple, new features like retrying make installing multiple specific versions, with latest dependencies. (the logic itself seems to work correctly, but the docstring doesn't explain what happens with multiple)

src/utils/DependencyUtils.ts
```diff
@@ -1,3 +1,26 @@
+import Game from "../model/game/Game";
+import ManifestV2 from "../model/ManifestV2";
+import ThunderstoreCombo from "../model/ThunderstoreCombo";
+import ThunderstoreVersion from "../model/ThunderstoreVersion";
+import { getCombosByDependencyStrings } from '../r2mm/manager/PackageDexieStore';
+
+export enum InstallMode {
+    INSTALL_SPECIFIC = 0,
+    UPDATE_ALL = 1
```
VilppeRiskidev, line 9: This might or might not be slighlty misleading, as this only uses the latest version for the dependencies, but keeps the original version(s) of the input combo(s). Maybe adding a docstring comment explaining this would be good.
Or maybe the actual issue here is the "- `installMode` should be `INSTALL_SPECIFIC`". Saying it "should be" is a bit vague. As in "it should be, but feel free to use the other one" or "it should be, because the code doesn't currently use it in any other way than this". 🤔

src/utils/DependencyUtils.ts
anttimaki: My intention was to write the whole docstring from the perspective of "current usage in the app". I guess I could try to make that more explicit and add a note about other usage. Also I'll reconsider the wording to make it more exact than "shoulds".

VilppeRiskidev:
Ok, the changes look good now 👍

anttimaki (approved):

PR #1739, opened 2025-04-28
Dependency list unit tests

VilppeRiskidev:


test/jest/__mocks__/dexie.ts
```diff
@@ -0,0 +1,242 @@
...
+            versions: [{
+                full_name: "modWithDeepDependency",
+                name: "modWithDeepDependency",
+                version_number: "2.0.0",
+                dependencies: ["author-modWithDependency-1.0.0"],
+            },{
+                full_name: "modWithDeepDependency",
+                name: "modWithDeepDependency",
+                version_number: "1.0.0",
+                dependencies: ["author-modWithDependency-1.0.0"],
```
anttimaki, line 154: The test case where this is used would be slightly better if these two dependencies would depend on a different mod/version. So we could be certain that the dependency on the tail-end was picked from a correct "path".

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+        versionNumber: "1.0.0"
+    });
+
+    modWithDependency = createCombo({
+        name: "modWithDependency",
+        versionNumber: "1.0.0",
+        dependencies: ["author-mod-1.0.0"]
+    });
+
+    modWithDependency2 = createCombo({
```
anttimaki, line 68: Seems like a duplicate of `modWithTheSameDependency` and isn't  used?

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+    modWithLoopDependencyB = createCombo({
+        name: "modWithLoopDependencyB",
+        versionNumber: "1.0.0",
+        dependencies: ["author-modWithLoopDependencyA-1.0.0"]
+    });
+};
+
+describe("DependencyUtils.getFullDependencyList", () => {
+
+    beforeEach(() => {
```
anttimaki, line 114: Should this be `beforeAll`? I don't think `beforeSetup` currently does anything that needs to be redone/reset before each test case.

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+            expectToContainOnly(combos, [
+                "author-modWithDeepDependency-1.0.0",
+                "author-modWithDependency-2.0.0",
+                "author-mod-2.0.0"
+            ]);
+        });
+    });
+
+
+    describe("A mod with a loop dependency", () => {
```
anttimaki, line 187: - What happens if A1.0.0 Depends on B1.0.0, but B2.0.0 depends on A2.0.0?
- What happens in the currently tested scenario, or the one above, if A is a modpack?

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+                "author-modWithTheSameDependency-2.0.0",
+                "author-mod-2.0.0"
+            ]);
+        });
+    });
+
+    
+
+    describe("Single mod with a dependency to a mod that's already installed", () => {
+        it("should return the mod and its dependency's latest version", async () => {
```
anttimaki, line 219: - This description doesn't seem to match what's actually tested in `expectToContainOnly`
- How does a modpack behave in the same scenario i.e. should it have it's own test case?

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+
+            expectToContainOnly(combos, [
+                "author-modPack-1.0.0",
+                "author-mod-1.0.0"
+            ]);
+        });
+    });
+
+
+    describe("A modPack and a mod with the same dependency", () => {
```
anttimaki, line 246: Nitpick: this should be plain English so "modPack" looks weird. (Also on the next line.)

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+        it("should return the modPack's specific version, mod's specific version and the dependency's last version", async () => {
```
anttimaki, line 247: Should it though? I'm not sure we have a spec that comments one way or another if the latest or specific version should be returned in a case like this? I might also be outside the scope of this PR.

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+    });
+}); 
+
+function expectToContainOnly(combos: ThunderstoreCombo[], expected: string[]) {
+    const dependencyStrings = combos.map(combo => combo.getDependencyString());
+    expected.forEach((e) => {
+        expect(dependencyStrings).toContain(e);
+    })
+    expect(combos).toHaveLength(expected.length);
+}
```
anttimaki, line 280: Nitpick missing newline.

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
VilppeRiskidev: It is now `modWithAnotherDependency` (and has a different dependency), both are now used in dependencies in dexie. Not in tests though, but I think it's nice to keep the dexie mock's and test's mod list similar.
(IDK if there's some better way to do it than just repeating the same stuff)

test/jest/__mocks__/dexie.ts
VilppeRiskidev: now it uses `modWithAnotherDependency` for the `1.0.0`

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
VilppeRiskidev: These are now demonstrated in test cases:
- What happens if A1.0.0 Depends on B1.0.0, but B2.0.0 depends on A2.0.0?
  - It will always use the latest version for the dependency

- What happens in the currently tested scenario, or the one above, if A is a modpack?
  - If we start from the mod pack, it uses the **specific** version of the dependency (mod)
  - If we start from the mod, it uses the **latest** version of the dependency (mod pack)

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
VilppeRiskidev: Oops, now it is described correctly and I added the case for the mod pack (which for some reason is broken, but I'll fix it)

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
VilppeRiskidev: This is now updated, it should now at least work in the same way it worked before.

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
VilppeRiskidev: Fixed now.

VilppeRiskidev:
All of the comments should be addressed.

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
```diff
@@ -0,0 +1,368 @@
...
+                "author-mod-1.0.0"
+            ]);
+        });
+    });
+
+
+    describe("A mod pack first and a mod with the same dependency last", () => {
+        it("should return the mod pack's specific version, " +
+            "mod's specific version, " +
+            "and the dependency's last version", async () => {
```
anttimaki, line 191: Should this say "specific" instead of "last"?

---

 _🎵 To be fair 🎵_, reading what's actually going on in these tests is cumbersome, as the reader has to jump from the test to the beforeAll to see what the combos actually contain, and then to the Dexie file to see what packages are actually actually present. Ideally each test case would be self-contained i.e. the "setup phase" (currently `const combos = [something];`) would actually populate Dexie with the packages required for this test case specifically.
 
 I'm not exactly sure how much effort it would take to get the Dexie mock to support this while retaining sufficient similarity to how Dexie actually works. Also this would be a tradeoff between boilerplate code/DRY. We can discuss this in the weekly/daily meeting.

anttimaki (commented):
I was in the middle of this review when I found out that the actual implmentation had some relatively major changes so I'll have to check that one first. This is therefore only a partial review.

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
```diff
@@ -0,0 +1,597 @@
...
+import ThunderstoreCombo from '../../../../src/model/ThunderstoreCombo';
+import ThunderstoreMod from "../../../../src/model/ThunderstoreMod";
+import ThunderstoreVersion from "../../../../src/model/ThunderstoreVersion";
+import VersionNumber from "../../../../src/model/VersionNumber";
+import { getFullDependencyList, InstallMode } from "../../../../src/utils/DependencyUtils";
+
+// Mock IndexedDB
+jest.mock('dexie');
+
+function createCombo(params: {
```
anttimaki, line 18: Seems unused?

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
```diff
@@ -0,0 +1,597 @@
...
+                ]
+            });
+
+            const combos = [modPack];
+            expectToContainOnly(await getFullDependencyList(combos, game, [], installMode), expected);
+        });
+    });
+
+
+    describe("A mod pack with a dependency, with a newer version of the dependency already installed", () => {
```
anttimaki, line 188: For clarity, this test case could be moved under the test case on L560, so the similar tests are grouped together.

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
```diff
@@ -0,0 +1,597 @@
...
+
+            const installedMod = new ManifestV2().fromThunderstoreCombo(mod);
+            installedMod.setVersionNumber(new VersionNumber("2.0.0"));
+
+            const combos = [modPack];
+            expectToContainOnly(await getFullDependencyList(combos, game, [installedMod], installMode), expected);
+        });
+    });
+
+    describe("A mod pack first and a mod with the same dependency last", () => {
```
anttimaki, line 219: I like that we have tests to check also combinations of `INSTALL_SPECIFIC` and passing multiple combos work in a sane fashion. At the same time they test behaviour that is not used nor formally specified. In practice I think this means the tests are now working as the spec, which would normally be fine. I'm just thinking of a scenario where someone augments `getFullDependencyList` when they actually want to install multiple specific mods, and then the tests break. In such case, I would be hesitant to move forward with the changes. I was wondering if we should alleviate those fears with a comment or something?

Just thinking out loud, not asking for any changes at the moment.

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
```diff
@@ -0,0 +1,597 @@
...
+    });
+
+
+    describe("A mod with a loop dependency, but one of the dependencies is a mod pack, starting from the mod pack", () => {
+        it.each([
+            [InstallMode.INSTALL_SPECIFIC, ["author-modWithLoopDependencyButOneOfThemIsAModPackA-1.0.0", "author-modWithLoopDependencyButOneOfThemIsAModPackB-1.0.0"]],
+            [InstallMode.UPDATE_ALL, ["author-modWithLoopDependencyButOneOfThemIsAModPackA-1.0.0", "author-modWithLoopDependencyButOneOfThemIsAModPackB-2.0.0"]],
+        ])
+        ("", async (installMode, expected) => {
+            const modWithLoopDependencyButOneOfThemIsAModPackA = addMockPackage({
```
anttimaki, line 428: Nitpick: `modWithLoopDependency` and `modpackWithLoopDependency` would be clearer names here and in the next test.

anttimaki (changes requested):
My comments aren't blockers, but the fact that this doesn't pass the CI pipeline is. Can you figure out what that's about?

test/jest/__tests__/utils/utils.DependencyUtils.ts.spec.ts
VilppeRiskidev: I added a comment (L67->), which should clarify it at least a little bit.

anttimaki:
@VilppeRiskidev I marked my last round of comments as resolved. Withholding the approval while the tests in CI fail. Any progress on that front?

anttimaki (approved):

PR #1742, opened 2025-04-29
Add no-wrap to DownloadMonitor's item

VilppeRiskidev:
* Long titles made the buttons wrap to the next row, which looked bad.


anttimaki (approved):
LGTM

PR #1744, opened 2025-04-30
Refactor and fix unit tests for InstallRuleInstaller

anttimaki:


test/jest/__tests__/impl/install_logic/Installer.Tests.spec.ts
```diff
@@ -122,26 +92,81 @@ describe('Installer Tests', () => {
...
+            expect(await FsProvider.instance.exists(
+                profile.joinToProfilePath(pkg.getName(), 'folder', 'subfolder', 'loose.file')
+            )).toBeTruthy();
+
+            expect(await FsProvider.instance.exists(
+                profile.joinToProfilePath(pkg.getName(), 'manifest.json')
+            )).toBeTruthy();
+        });
+
+        test('No default location', async () => {
```
ebkr, line 131: Is this the behaviour we'd expect? What would we expect to happen in this case - Maybe an error to be thrown?

Happy if this is refactoring to existing behaviour, not a blocker for this PR and can revisit what we think should happen.

test/jest/__utils__/InstallLogicUtils.ts
```diff
@@ -78,6 +81,81 @@ export async function createPackageFilesIntoCache(pkg: ManifestV2, filePaths: st
...
+        );
+        expect(await FsProvider.instance.exists(targetPath)).toBeTruthy();
+    }
+}
+
+// For testing file installation with the contents of folder-structure-testing folder.
+export async function testUntrackedFileStructure(pkg: ManifestV2, paths: string[]) {
+    for (const packagePath of paths) {
+        const targetPath = Profile.getActiveProfile().joinToProfilePath(
+            "BepInEx",
```
ebkr: Can we guarantee these are always going to be BepInEx? Would need to double check but I believe these are inherited from the route in the actual application.

test/jest/__tests__/impl/install_logic/Installer.Tests.spec.ts
anttimaki: This is just adding game-agnostic tests for the tracking method that didn't have any before. The actual implementation of the tracking method hasn't changed. How this (and other) tracking methods should work in special cases or misconfigurations is an interesting question but I don't have answers to it.

test/jest/__utils__/InstallLogicUtils.ts
anttimaki: I guess there's no guarantee. This refactoring merely takes a bit of code that was copy-pasted to 16 game specific test files and DRYs it into a single helper function. For those 16 games it just happens that the untracked install path that is tested is `BepInEx\config`, so the hardcoded "BepInEx" is expected for them all.

I guess I could pass the "BepInEx" part in as an argument? Although IDK if we should reconsider a) what we are testing b) how to test it effectively. After the ecosystem repo is the source of truth for install rules, these test will basically fail if the config in ecosystem repo is changed. That could have some value in detecting unintended changes, but at the same time it sounds like the client is telling the source of truth that it's wrong. The tail is wagging the dog. Anyway these bigger changes are IMO outside of the scope of the PR, but the argument-based approach would be doable?

anttimaki: Changed the helpers for [untracked installations](https://github.com/ebkr/r2modmanPlus/pull/1744/commits/784763ca4a59ed767bc4999005e79e3e88976ab0#diff-c0932d0cd1147c2f25976294467ac59449ed43b15829930c0df85735fa3ed004) and [subdir installations](https://github.com/ebkr/r2modmanPlus/pull/1744/commits/310621f7da8e3e20ce2857b77f1dbd74e11547a6#diff-c0932d0cd1147c2f25976294467ac59449ed43b15829930c0df85735fa3ed004) to use the same parameter type as the state tracked installations, i.e. passing the "BepInEx" part as a parameter.

Doing the changes reinforced my opinion that testing game specific InstallRuleInstallers in mod manager is a bit silly when the installation rules themselves are no longer defined within the repo. Instead, we should have more of generic tests where we [define a custom InstallRule](https://github.com/ebkr/r2modmanPlus/pull/1744/commits/310621f7da8e3e20ce2857b77f1dbd74e11547a6#diff-c0932d0cd1147c2f25976294467ac59449ed43b15829930c0df85735fa3ed004) and check that the files get placed in correct paths. Although I guess the tests should be located in TCLI (or its plugins) once it takes over the responsibility of installations, so this all a bit moot.

ebkr: Yeah I think that the installation logic tests should move to TCLI once we start using that. There's not a huge reason we'd need to verify that behaviour at that point because all we do is proxy it. Integration tests maybe, but they'd be for real scenarios.

PR #1746, opened 2025-04-30
Merge InstallationRuleApplicator to InstallationRules

anttimaki:
After moving from hard coded values to using Thunderstore ecosystem data as the source of truth, the class has become a stub and can be merged with the other class that contains rest of the related code.


VilppeRiskidev (approved):
LGTM 👍

ebkr (approved):

PR #1747, opened 2025-04-30
Profile switcher

ebkr:
https://github.com/user-attachments/assets/d9e5bfcb-7cfd-45e1-82ed-2a137745e2dd

### DIFF
--- a/src/components/modals/ProfileManagementModal.vue
+++ b/src/components/modals/ProfileManagementModal.vue
@@ -5,6 +5,7 @@ import useStore from '../../store';
 import SettingsItem from '../settings-components/SettingsItem.vue';
 
 import VueRouter from 'vue-router';
+import R2Error from 'src/model/errors/R2Error';
 
 const store = useStore();
 let router!: VueRouter;
@@ -25,17 +26,25 @@ async function changeProfile() {
 }
 
 async function exportProfileAsFile() {
-    store.dispatch("profileExport/exportProfileAsFile");
+    try {
+        await store.dispatch("profileExport/exportProfileAsFile");
+    } catch (e) {
+        store.commit('error/handleError', R2Error.fromThrownValue(e));
+    }
 }
 
 async function exportProfileAsCode() {
-    store.dispatch("profileExport/exportProfileAsCode");
+    try {
+        await store.dispatch("profileExport/exportProfileAsCode");
+    } catch (e) {
+        store.commit('error/handleError', R2Error.fromThrownValue(e));
+    }
     store.commit("closeProfileManagementModal");
 }
 </script>
 
 <template>
-    <ModalCard :can-close="true" :is-active="isOpen" @close-modal="closeModal">
+    <ModalCard id="profile-management-modal" :can-close="true" :is-active="isOpen" @close-modal="closeModal">
         <template v-slot:header>
             <h2 class='modal-title non-selectable'>Profile</h2>
         </template>
--- a/src/components/modals/ProfileManagementModal.vue
+++ b/src/components/modals/ProfileManagementModal.vue
@@ -2,7 +2,7 @@
 import ModalCard from '../ModalCard.vue';
 import { computed, getCurrentInstance, onMounted } from 'vue';
 import useStore from '../../store';
-import SettingsItem from 'components/settings-components/SettingsItem.vue';
+import SettingsItem from '../settings-components/SettingsItem.vue';
 
 import VueRouter from 'vue-router';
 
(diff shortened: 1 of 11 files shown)


src/store/modules/ProfileExportModule.ts
```diff
@@ -0,0 +1,69 @@
...
+            }
+            const immutableProfile = rootGetters["profile/activeProfileOrThrow"].asImmutableProfile();
+            const exportErr = await ProfileModList.exportModListToFile(immutableProfile);
+            if (exportErr instanceof R2Error) {
+                throw exportErr;
+            }
+        },
+
+        async exportProfileAsCode({rootGetters, getters, commit}) {
+            console.log(rootGetters, getters);
```
anttimaki, line 45: Logging

src/store/modules/ProfileExportModule.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        async exportProfileAsFile({rootGetters}) {
+            if (!rootGetters["profile/modList"].length) {
+                throw new R2Error(
+                    'Profile is empty',
+                    'The profile must contain at least one mod to export it as a file.'
+                );
+            }
+            const immutableProfile = rootGetters["profile/activeProfileOrThrow"].asImmutableProfile();
+            const exportErr = await ProfileModList.exportModListToFile(immutableProfile);
```
anttimaki, line 38: FYI I've added [throwForR2Error](https://github.com/ebkr/r2modmanPlus/blob/develop/src/model/errors/R2Error.ts#L34) that can be used in situations like this.

src/store/modules/ProfileExportModule.ts
```diff
@@ -0,0 +1,69 @@
...
+                } else {
+                    InteractionProvider.instance.copyToClipboard(code);
+                    commit('setExportCode', code);
+                }
+            });
+            if (exportErr instanceof R2Error) {
+                throw exportErr;
+            }
+            commit("openProfileCodeExportModal", null, {root: true});
+            commit("openProfileCodeExportModal", null, {root: true});
```
anttimaki, line 65: Double

anttimaki (commented):
Left some minor comments but overall LGTM.

src/store/modules/ModalsModule.ts
```diff
@@ -1,5 +1,7 @@
 import ManifestV2 from "../../model/ManifestV2";
 import ThunderstoreMod from "../../model/ThunderstoreMod";
+import { ActionTree } from 'vuex';
+import { State as RootState } from 'src/store';
```
anttimaki, line 4: Absolute import path.

src/components/modals/ProfileManagementModal.vue
```diff
@@ -0,0 +1,77 @@
+<script lang="ts" setup>
+import ModalCard from '../ModalCard.vue';
+import { computed, getCurrentInstance, onMounted } from 'vue';
+import useStore from '../../store';
+import SettingsItem from '../settings-components/SettingsItem.vue';
+
+import VueRouter from 'vue-router';
+import R2Error from 'src/model/errors/R2Error';
```
anttimaki, line 8: Another absolute import path sneaked in.

src/components/modals/ProfileManagementModal.vue
```diff
@@ -0,0 +1,77 @@
...
+function closeModal() {
+    store.commit('closeProfileManagementModal');
+}
+
+async function changeProfile() {
+    store.commit("closeProfileManagementModal");
+    router.push({name: 'profiles'});
+}
+
+async function exportProfileAsFile() {
```
anttimaki, line 28: Is it intentional that this leaves the modal open on error and success, whereas `exportProfileAsCode` closes the modal on error and switches to another modal on success?

src/store/modules/ModalsModule.ts
```diff
@@ -128,6 +140,15 @@ export default {
...
             state.isDownloadModModalOpen = true;
-        }
+        },
+
+        openProfileManagementModal: function(state: State): void {
+            state.isProfileManagementModalOpen = true;
+        },
+
+        openProfileCodeExportModal: function(state: State): void {
+            console.log("Opened");
```
anttimaki, line 150: Logging

anttimaki (commented):
A couple of new small comments.

src/components/modals/ProfileManagementModal.vue
ebkr: It is - The code export opens a new modal whereas this one doesn't and instead waits for the file selection dialog.

IMO it feels more natural to keep it like this - any thoughts?

anttimaki (approved):

src/components/modals/ProfileManagementModal.vue
ebkr: It's mostly just because Windows doesn't necessarily bring the file window to the foreground so the user doesn't really have any kind of confirmation that something has happened unless they happen to see File Explorer flashing.

We can probably circumvent this with a separate confirmation modal, although it feels redundant.

I'm happy to do a separate PR to auto-close it to keep it consistent though.

PR #1751, opened 2025-05-07
Utilise enums provided by Thunderstore ecosystem schema

anttimaki:
This removes the need to redefine same enums in the mod manager and having to map between the values.

PackageLoaders were not included in this PR as it makes more sense to revisit those once PackageLoader.None is used instead of null values. #1687 provides mod manager side of the changes, but changes are needed to ecosystem repo too.


VilppeRiskidev (approved):
Very nice 👌 👍

anttimaki:
Closing in favour of #1821 .

PR #1753, opened 2025-05-10
Moved import local mod functionality into ProfileManagementModal

ebkr:
The bulk of the changes of this PR are porting the modal to the composition API. Should be okay to mostly ignore that file, and loop back to changes to that file as part of the second commit.

### DIFF
--- a/src/components/importing/LocalFileImportModal.vue
+++ b/src/components/importing/LocalFileImportModal.vue
@@ -1,6 +1,6 @@
 <template>
     <div>
-        <ModalCard id="import-mod-from-file-modal" :can-close="true" @close-modal="emitClose" :is-active="visible">
+        <ModalCard id="import-mod-from-file-modal" :can-close="true" @close-modal="emitClose" :is-active="isOpen">
             <template v-slot:header>
                 <h2 class='modal-title'>Import mod from file</h2>
             </template>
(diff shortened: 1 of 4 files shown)


src/components/importing/LocalFileImportModal.vue
```diff
@@ -100,242 +98,244 @@ import { ImmutableProfile } from '../../model/Profile';
...
-
-    private resultingManifest = new ManifestV2();
-
-    @Watch("visible")
-    private visiblityChanged() {
-        this.fileToImport = null;
-        this.waitingForSelection = false;
-        this.validationMessage = null;
-    }
+type LocalFileImportModalProps = {
```
anttimaki, line 106: Offtopic, but should we have some repo-wide conventions on how the composition-style components should be structured/ordered? Does Vue docs define any best practices for this? I mean like trying to have the components define things in order, e.g.

1. Props type and props
2. Computed funcs
3. Watch funcs
4. Module-level variables
5. and so forth
6. Template

This probably can't be enforced with linters, but e.g. having a template for new components could be nice? Thoughts?

anttimaki (approved):

src/components/importing/LocalFileImportModal.vue
ebkr: Honestly I don't actually mind. I think it's better to allow these kinds of things to be a bit loose to let people make things closer where relevant.

Basically I don't really mind if it ends up like this:
```
type SomeFunctionality = { ... }
type DifferentBehaviour = { ... }

const selectedFunctionality = ref<SomeFunctionality>...
const selectedBehaviour = ref<DifferentBehaviour>...
```
compared to:
```
type SomeFunctionality = { ... }
const selectedFunctionality = ref<SomeFunctionality>...

type DifferentBehaviour = { ... }
const selectedBehaviour = ref<DifferentBehaviour>...
```
Both have their benefits and I think it's good to let people decide on how they think it'll work best based on what they're building.

I do think the only thing we should really enforce is the following:
- Store declaration at the top
  - This will be identical per file, so it makes sense to essentially treat it as boilerplate, as you'd do with imports.
- Props type definition following
  - Always something that's actually useful to see because that decides what you pass through.

PR #1756, opened 2025-05-14
Store provider

ebkr:


src/components/importing/LocalFileImportModal.vue
```diff
@@ -99,9 +99,10 @@ import ProfileModList from '../../r2mm/mods/ProfileModList';
 import LocalModInstallerProvider from '../../providers/ror2/installing/LocalModInstallerProvider';
 import ModalCard from '../ModalCard.vue';
 import { ref, defineProps, computed } from 'vue';
-import useStore from '../../store';
+import { getStore } from '../../providers/generic/store/StoreProvider';
+import { State } from '../../store';
 
-const store = useStore();
+const store = getStore<State>();
```
anttimaki, line 105: Not a blocker given that the type safety in our Vuex version is non-existing in the old implementations as well, but at least on Cursor the `store` here know only the types for State items in the index file, i.e. it has no idea about the submodules. I.e:

```typescript
const isOpen = computed(() => store.state.modals.isLocalFileImportModalOpen);  // isOpen is ComputedRef<any>

const game = computed(() => store.state.activeGame); // game is ComputedRef<Game>
```

anttimaki (approved):

PR #1757, opened 2025-05-15
DownloadMixin to Composable

ebkr:


src/components/composables/DownloadComposable.ts
```diff
@@ -0,0 +1,59 @@
+import { getStore } from '../../providers/generic/store/StoreProvider';
+import { State } from '../../store';
+import ThunderstoreCombo from 'src/model/ThunderstoreCombo';
```
anttimaki, line 3: Absolute imports.

src/components/composables/DownloadComposable.ts
```diff
@@ -0,0 +1,59 @@
...
+import R2Error, { throwForR2Error } from 'src/model/errors/R2Error';
+import { ImmutableProfile } from 'src/model/Profile';
+import ProfileModList from 'src/r2mm/mods/ProfileModList';
+import { installModsToProfile } from 'src/utils/ProfileUtils';
+import ConflictManagementProvider from 'src/providers/generic/installing/ConflictManagementProvider';
+import { computed } from 'vue';
+
+export function useDownloadComposable() {
```
anttimaki, line 11: Not a blocker or necessarily even a change request, but I'm wondering what some best practices for composables would be going forward.

1. Does it make sense grouping different functions into one composable? For mixins (at least on Python) it's the pattern, but AFAIK e.g. for React's custom hooks they tend to be standalone functions.

2. Does it makes sense to call composables in the computed getters? How sure are you that reactivity works as expected when doing so?

3. Does it make sense to wrap oneliners like `isOpen` and `closeModal` into composable? Especially when they're then wrapped again in the computed getters? Seems like an unnecessary extra hoop to jump through?

I guess the above questions stem from using composable as a one-to-one replacement for the mixin, whereas my understanding is that they're sort of a different pattern, and intended to be used differently, even if they technically can be used in the same way.

anttimaki (changes requested):

src/components/composables/DownloadComposable.ts
ebkr: > Does it make sense grouping different functions into one composable?

Honestly the docs aren't too helpful about it and I guess it's up to the consumer. For now I'd say that we should leave it as is purely from a port perspective rather than a refactor.

https://vuejs.org/guide/reusability/composables#conventions-and-best-practices

It does suggest each composable being named `useXYZ` so from a namespace perspective and name convention I think we're following that well enough.

> Does it make sense to wrap oneliners like `isOpen` and `closeModal` into composable?

After the issue with regards to cyclic referencing, no. I'll change this.

> Does it makes sense to call composables in the computed getters?

Likewise for above, removing the one liners will remove the need for the odd computed getters, and we'll only use getters to expose the underlying composable methods.

anttimaki (approved):

PR #1762, opened 2025-05-18
Re-added removed settings to aid transition to new profile management modal

ebkr:
It'll be quite a workflow jump to remove all of these settings and so it'll likely receive less backlash should these be removed gradually instead. 

We don't currently have a nice way to indicate that something like this has happened.


ebkr:
@anttimaki No need to review, just FYI

PR #1772, opened 2025-05-20
Composition API: Profile Modals

ebkr:


src/components/profiles-modals/RenameProfileModal.vue
```diff
@@ -1,75 +1,55 @@
...
-        const { makeProfileNameSafe } = useProfilesComposable();
-        return makeProfileNameSafe;
-    }
+const {
+    doesProfileExist,
+    makeProfileNameSafe,
+} = useProfilesComposable();
 
-    @Watch('$store.state.profile.activeProfile')
-    activeProfileChanged(newProfile: Profile, oldProfile: Profile|null) {
```
ebkr, line 27: This doesn't actually seem to affect anything. Removing has had no apparent behaviour changes.

anttimaki: It's probably a leftover from when all the profile modals were in one component and some of the UI elements were shared (e.g. between creating and renaming a profile).

anttimaki (approved):
LGTM

PR #1783, opened 2025-05-22
Revamp the DependencyUtils tests, simplifying it

VilppeRiskidev:
Revamp the DependencyUtils tests, simplifying it to only mock the getCombosByDependencyStrings() function


VilppeRiskidev:
Better alternative has been created, this is obsolete.

PR #1786, opened 2025-05-27
Drop Dexie mocking from DependencyUtils tests

anttimaki:
This is intended as an alternative for #1783


VilppeRiskidev:
Rebased the whole stack, hopefully there's no regressions

VilppeRiskidev (approved):
Ok 👍 Elegant and readable code, which also seems to do what it's supposed to 🤯

PR #1793, opened 2025-06-03
Ported App.vue to Composition API

ebkr:


src/components/composables/UtilityComposable.ts
```diff
@@ -60,45 +59,50 @@ export default class UtilityMixin extends Vue {
      * closing the window while update is in progress. Ignore the first
      * failure to see how this affects the number of reported errors.
      */
-    private async backgroundRefreshThunderstoreModList() {
+    async function backgroundRefreshThunderstoreModList(router: VueRouter) {
```
anttimaki, line 63: It doesn't look like the router arg is ever passed to this function?

ebkr: It's used on line 69 to prevent doing background refreshing on pages that wouldn't support it

anttimaki: What I meant was that the argument is not passed in on L19 and will always be `undefined`.

anttimaki (approved):
LGTM

PR #1795, opened 2025-06-04
Undo accidental reverts

anttimaki:


src/components/v2/OnlinePreviewPanel.vue
ebkr, line 1: Think this just needs a rebase?

ebkr: These are all the changes otherwise
image

anttimaki: > Think this just needs a rebase?

Not sure what you mean? The dependency resolution was yanked out of download provider and now resides in DependencyUtils. It also deals with combos (since download deals with combos). So I think the changes are needed for the dependency listing to work.

PR #1798, opened 2025-06-04
Add a retry functionality to DownloadMonitor's failed downloads

VilppeRiskidev:
* New action was added to Vuex store module
* installMode is now stored within the DownloadProgress


VilppeRiskidev:
This, to me, is the simplest way to solve this.

I thought about figuring out the installMode by looking at what's being retried (multiple downloads means updateAll etc.)... 

Then there's the future possibility of having an action "center"/"log", from which, in theory, any (failed) action could be retried.
So, I think in that case "update all" should be a separate _action_ from downloads...
And the download retrying should (in any case) be "generic" in a sense that it doesn't depend on how the rest of the application works, i.e.  it should _just_ retry the _exact_ download, the versions/dependencies etc. should be handled by other parts.

That's my train of thoughts, feel free to challenge me.

anttimaki:
The current implementation opens up a scenario where the user after a failed download changes the profile, or even active game, and then retries the download. Even for a different game, the main mod will be installed, although the dependency resolution fails if the deps aren't shared between the communities.

The simplest solution would probably be to just reset the download Vuex store whenever user proceeds from the profile selection field. Unfortunately this would soft-break the status indicator in TSMM, and might also cause issues when progress callbacks are triggered.

I think a better solution would be to store the selected game and profile in the DownloadProgress object too, and filter the DownloadMonitor listing based on those. "Clear finished" button would have to respect game/profile filtering too.

Thoughts? @ebkr could chip in too.

src/pages/DownloadMonitor.vue
```diff
@@ -89,7 +89,15 @@
                             v-tooltip.left="'Remove'"
                             @click="$store.commit('download/removeDownload', downloadObject)"
                         >
-                            <i class="fas fa-times" />
+                            <i class="fas fa-times x-icon" />
+                        </button>
+                        <button
```
anttimaki, line 94: I would swap the order of these two buttons, so when the listing has both completed and failed downloads, the X buttons would all be on the right.

ebkr:
> I think a better solution would be to store the selected game and profile in the DownloadProgress object too, and filter the DownloadMonitor listing based on those. "Clear finished" button would have to respect game/profile filtering too.

I agree that this is a nice way to do it. This also means if someone re-enters the game/profile that they'd be able to resume things. Unrelated thought if we do this approach but do we want to write them to an indexeddb store to resume at a later point?

An alternative idea is that we don't filter these out. We can then either:
1. Upon resume, open a modal to inform the user that these won't be installed on the current profile
2. Display a message alongside the retry button to inform the user that these won't be installed on the current profile

I don't overly have a preference. It's unlikely that someone wants to track downloads across several profiles at a time whilst downloads are in progress, and even if they did, it's out of scope for the managers to support this behaviour because it feels extremely power-user esque. This means that I'm leaning more towards the first solution anyway.

src/store/modules/DownloadModule.ts
```diff
@@ -173,12 +191,27 @@ export const DownloadModule = {
     },
 
     getters: <GetterTree<State, RootState>>{
+        profileDownloads(state, _getters, rootState, rootGetters) {
+            if (!state.allDownloads || state.allDownloads.length === 0) {
```
anttimaki: Nitpick: I don't think `!state.allDownloads` can ever be anything but truthy. There's IMO no benefit for checking an empty array here either?

src/store/modules/DownloadModule.ts
```diff
@@ -173,12 +191,27 @@ export const DownloadModule = {
...
+                return [];
+            }
+            return state.allDownloads.filter((dl: DownloadProgress) => {
+                return dl.game.internalFolderName === rootState.activeGame.internalFolderName
```
anttimaki: Nitpick: I don't think checking the internalFolderName is necessary here, the profile path should contain it anyway and should suffice alone.

src/store/modules/DownloadModule.ts
```diff
@@ -173,12 +191,27 @@ export const DownloadModule = {
...
+                    && dl.profile.getProfilePath() === rootGetters['profile/activeProfile'].getProfilePath();
+            });
+        },
         activeDownloadCount(_state, getters) {
             return getters.activeDownloads.length;
         },
+        activeProfileDownloadCount(_state, getters) {
```
anttimaki: Nitpick: if we have `profileDownloads` should this be named `profileActiveDownloadCount` for consistency? (Also `profileActiveDownloads` below).

Sorting the getters alphabetically would mean all `profile...` getters would then be nicely grouped together.

src/store/modules/DownloadModule.ts
```diff
@@ -249,6 +282,9 @@ function getIndexOfDownloadProgress(allDownloads: DownloadProgress[], downloadId
 }
 
 function getOnlyActiveDownloads(downloads: DownloadProgress[]): DownloadProgress[] {
+    if (downloads.length === 0) {
```
anttimaki: Nitpick: I think this is premature optimization and just noise?

anttimaki (changes requested):
I'm encountering bug like behaviour. I'm not sure but it might be caused by [this line](https://github.com/ebkr/r2modmanPlus/pull/1798/files#diff-cf33b44e204593e04370ee7c21148017934df9aea20b086a0936aa6651679555R60). Reproduction:

1. Start downloading a mod, disable network in dev tools so the download errors out
2. Enable network connection
3. Go to download monitor and retry. Download restarts as expected
4. Move to another view, e.g. installed. The download progress modal pops up. This is unexpected. If the modal is to pop up, it should happen when retry is clicked. However, as the retry is (in this case) clicked in a view that already displays the progress, it would be better to not show the modal at all. Note that when retrying from the error modal, the progress modal should be shown.


Sidenote: it would be a good practice to rebase PRs like this to latest develop branch when pushing changes.

src/store/modules/DownloadModule.ts
VilppeRiskidev: Renamed and sorted. I put conciseDownloadStatus last, as I feel it's cleaner this way.

VilppeRiskidev:
Comments are addressed 👍

anttimaki (approved):
:+1:

PR #1802, opened 2025-06-05
Allow retrying failed downloads from ErrorModal

VilppeRiskidev:
* Added support for extra action button to ErrorModal and use it to retry the failed download

### DIFF
--- a/src/components/modals/ErrorModal.vue
+++ b/src/components/modals/ErrorModal.vue
@@ -28,6 +28,11 @@ function close() {
                     <h5 class="title is-5">Suggestion</h5>
                     <p>{{solution}}</p>
                 </div>
+                <div class="mt-3 text-right" v-if="error.action">
+                    <button class="button is-white" @click="() => { error.action.function(); close(); }">
+                        {{error.action.label}}
+                    </button>
+                </div>
             </div>
         </div>
         <button class="modal-close is-large" aria-label="close" @click="close"></button>
--- a/src/model/errors/R2Error.ts
+++ b/src/model/errors/R2Error.ts
@@ -1,9 +1,15 @@
+interface Action {
+    function: () => Promise<void>;
+    label: string;
+}
+
 export default class R2Error extends Error {
     public name: string;
     public message: string;
     public stack?: string | undefined;
     public solution: string;
     public errorReferenceString: string | undefined;
+    public action?: Action;
 
     public constructor(name: string, message: string, solution: string | null = null) {
         super(message);
@@ -25,6 +31,10 @@ export default class R2Error extends Error {
         } else {
             return new R2Error(name, `${error}`, solution);
         }
+    }
+
+    public setAction(action: Action) {
+        this.action = action;
     }
 }
 
--- a/src/store/modules/DownloadModule.ts
+++ b/src/store/modules/DownloadModule.ts
@@ -67,6 +67,18 @@ export const DownloadModule = {
             commit('removeDownload', params.download);
         },
 
+        retryDownloadById({state, commit, dispatch}, downloadId: UUID) {
+            const download = state.allDownloads.find((d: DownloadProgress) => d.downloadId === downloadId);
+            if (!download) {
+                commit('error/handleError',
+                    new R2Error('Download not found', 'Tried to retry a download, but it wasn\'t found in the list of downloads.'),
+                    { root: true }
+                );
+                return;
+            }
+            dispatch('retryDownload', { download });
+        },
+
         _addDownload({state}, params: {
             combos: ThunderstoreCombo[],
             installMode: InstallMode,
@@ -96,7 +108,7 @@ export const DownloadModule = {
             commit('setIgnoreCacheVuexOnly', settings.getContext().global.ignoreCache);
         },
 
-        async downloadAndInstallCombos({commit, dispatch}, params: {
+        async downloadAndInstallCombos({commit, dispatch, rootGetters}, params: {
             combos: ThunderstoreCombo[],
             game: Game,
             installMode: InstallMode,
@@ -118,10 +130,17 @@ export const DownloadModule = {
                 await dispatch('_installModsAndResolveConflicts', { combos: modsWithDependencies, profile, downloadId });
                 commit('setDone', downloadId);
             } catch (e) {
+                const r2Error = R2Error.fromThrownValue(e);
                 if (downloadId) {
                     commit('setFailed', downloadId);
+                    if (profile.getProfilePath() === rootGetters['profile/activeProfile'].getProfilePath()) {
+                        r2Error.setAction({
+                            label: 'Retry',
+                            function: () => dispatch('retryDownloadById', downloadId)
+                        });
+                    }
                 }
-                commit('error/handleError', R2Error.fromThrownValue(e), { root: true });
+                commit('error/handleError', r2Error, { root: true });
             } finally {
                 if (!hideModal) {
                     commit('setIsModProgressModalOpen', false);
@@ -204,9 +223,6 @@ export const DownloadModule = {
     getters: <GetterTree<State, RootState>>{
         activeDownloadCount(_state, getters) {
             return getters.activeDownloads.length;
-        },
-        activeProfileDownloadCount(_state, getters) {
-            return getters.activeProfileDownloads.length;
         },
         activeDownloads(state) {
             return getOnlyActiveDownloads(state.allDownloads);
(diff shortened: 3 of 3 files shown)


src/components/buttons/ErrorActionButton.vue
```diff
@@ -0,0 +1,28 @@
+<template>
```
anttimaki: To not undo ebkr's progress, any new components that we add should IMO use the composition API.

src/components/modals/ErrorModal.vue
```diff
@@ -28,6 +29,9 @@ function close() {
                     <h5 class="title is-5">Suggestion</h5>
                     <p>{{solution}}</p>
                 </div>
+                <div class="mt-3 text-right" v-if="extraActionButton">
+                    <component :is="extraActionButton.component" v-bind="extraActionButton.props" />
```
anttimaki: Based on my intuition I think it might be better to just pass around the button text and a `async () => Promise<void>` type function than a component. Not a strong opinion though.

src/store/modules/DownloadModule.ts
```diff
@@ -63,6 +64,13 @@ export const DownloadModule = {
             commit('removeDownload', download);
         },
 
+        retryDownloadById({state, dispatch}, downloadId: UUID) {
+            const download = state.allDownloads.find((d: DownloadProgress) => d.downloadId === downloadId);
+            if (download) {
```
anttimaki: I'm not sure there's a scenario where the download object is not found, but currently this gives the user no indication that something went wrong and nothing's going to happen. Could we try to bake in error handling here?

src/store/modules/ErrorModule.ts
```diff
@@ -16,11 +18,17 @@ export default {
...
+        extraActionButton: null,
     }),
 
     mutations: {
         discardError: function(state: State): void {
             state.error = null;
+            state.extraActionButton = null;
+        },
+
+        setExtraActionButton: function(state: State, extraActionButton: Component): void {
```
anttimaki: I'm not a fan of having the action button a separate thing in the state as it seems unnecessarily complex. For example, it seems to me the button is never reset, so it gets shown again for the next error too? Would it be simpler to pass the button text and the action function within a new optional fields in the R2Error object? It gets passed to error modal anyway and it would ensure the error and the action would always be in sync.

src/components/buttons/ErrorActionButton.vue
VilppeRiskidev: Good point 👍 (although there's no new component anymore)

src/components/modals/ErrorModal.vue
VilppeRiskidev: Added the R2ErrorExtraAction interface.
I thought that the text and the function are tightly coupled (one shouldn't exist without the other).

src/components/modals/ErrorModal.vue
```diff
@@ -2,7 +2,7 @@
 import R2Error from '../../model/errors/R2Error';
 import { getStore } from '../../providers/generic/store/StoreProvider';
 import { State } from '../../store';
-import { computed, ComputedRef } from 'vue';
+import { computed, ComputedRef, Component } from 'vue';
```
anttimaki: `Component` is not used.

src/components/modals/ErrorModal.vue
```diff
@@ -28,6 +28,14 @@ function close() {
                     <h5 class="title is-5">Suggestion</h5>
                     <p>{{solution}}</p>
                 </div>
+                <div class="mt-3 text-right" v-if="error.extraAction">
+                    <button
```
anttimaki: Nitpick: trailing whitespace.

src/components/modals/ErrorModal.vue
```diff
@@ -28,6 +28,14 @@ function close() {
                     <h5 class="title is-5">Suggestion</h5>
                     <p>{{solution}}</p>
                 </div>
+                <div class="mt-3 text-right" v-if="error.extraAction">
+                    <button 
+                        class="button error-action-button is-danger"
```
anttimaki: Would `is-white` look better than `is-danger`? Then we wouldn't need the CSS rule for border either.

src/components/modals/ErrorModal.vue
```diff
@@ -28,6 +28,14 @@ function close() {
...
+                        @click="() => { error.extraAction.actionFunction(); $store.commit('error/discardError'); }"
```
anttimaki: Nitpick: could this just call `close()` instead of `$store.commit('error/discardError');`?

src/components/modals/ErrorModal.vue
```diff
@@ -38,4 +46,8 @@ function close() {
     p + div {
         margin-top: 1.5rem;
     }
+
```
anttimaki: Nitpick: trailing whitespace.

src/model/errors/R2Error.ts
```diff
@@ -1,9 +1,15 @@
+interface R2ErrorExtraAction {
```
anttimaki: Naming nitpick: the interface is used only internally in this module so the R2Error prefix is a bit redundant. "Extra" is also somewhat redundant IMO. Would "Action" be just as descriptive? Other option would be "ButtonProps" oslt?

src/model/errors/R2Error.ts
```diff
@@ -1,9 +1,15 @@
...
+    actionFunction: () => void;
```
anttimaki: Naming nitpick: if the interface is named "Action", then just "function" would suffice - on the call site is would read "action.function()". If "ButtonProps" is used for the interface, this could be "buttonProps.action()" or "buttonProps.function"().

Also could future proof this by using `Promise<void>` as the return value.

src/model/errors/R2Error.ts
```diff
@@ -1,9 +1,15 @@
...
+    actionText: string;
```
anttimaki: Naming nitpick: just "text" or "label" would suffice.

src/model/errors/R2Error.ts
```diff
@@ -1,9 +1,15 @@
...
+}
+
 export default class R2Error extends Error {
     public name: string;
     public message: string;
     public stack?: string | undefined;
     public solution: string;
     public errorReferenceString: string | undefined;
+    public extraAction?: R2ErrorExtraAction;
```
anttimaki: Naming nitpick: the attribute could as well be named just "action" or "button"

src/store/modules/DownloadModule.ts
```diff
@@ -1,6 +1,7 @@
 import { ActionTree, GetterTree } from "vuex";
 import UUID from 'uuid-js';
 
+import ErrorActionButton from '../../components/buttons/ErrorActionButton.vue';
```
anttimaki: Not used and the file doesn't even exist anymore?

src/store/modules/DownloadModule.ts
```diff
@@ -109,10 +122,18 @@ export const DownloadModule = {
                 await dispatch('_installModsAndResolveConflicts', { combos: modsWithDependencies, profile, downloadId });
                 commit('setDone', downloadId);
             } catch (e) {
+                const r2Error = R2Error.fromThrownValue(e);
                 if (downloadId) {
                     commit('setFailed', downloadId);
+                    if (game.internalFolderName ===  rootState.activeGame.internalFolderName
```
anttimaki: Nitpick: checking the profile path should be enough as it contains the game's name too.

src/store/modules/ErrorModule.ts
anttimaki: After the changes already done, are any changes actually needed in this file or should they all be reverted?

anttimaki (changes requested):
A couple cleanup comments and a bunch of nitpicks, but overall this works as expected. The downloader refactoring is bearing fruits.

VilppeRiskidev:
Fixed based on the comments.

src/store/modules/DownloadModule.ts
```diff
@@ -205,6 +224,9 @@ export const DownloadModule = {
         activeDownloadCount(_state, getters) {
             return getters.activeDownloads.length;
         },
+        activeProfileDownloadCount(_state, getters) {
```
anttimaki: Probably a rebasing error, this shouldn't be here as it's a duplicate of profileActiveDownloadCount.

VilppeRiskidev: Correct, oops 👍

PR #1803, opened 2025-06-05
Fix generateProgressPercentage by hard returning 100

VilppeRiskidev:
* Return 100 if the current item is the last one and the progress of it is 100


VilppeRiskidev:
So... This should fix it, I didn't figure out a good way to test this, as I didn't find a way to reproduce the bug... 🤔

anttimaki (approved):

PR #1805, opened 2025-06-06
Use PackageLoader enum provided by Thunderstore ecosystem

anttimaki:
Draft PR:

- Required changes are not implemented on ecosystem repo
  - Ecosystem files are edited manually to compensate, this separate commit should be dropped before merging
- Requires rebase once Manager.vue is ported to composition API

The code changes in the second commit are good for review though.


src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
```diff
@@ -379,8 +373,7 @@ export default class GenericProfileInstaller extends ProfileInstallerProvider {
             return false;  // Don't process mod loader with plugin installer.
         }
 
-        const installerId = GetInstallerIdForPlugin(GameManager.activeGame.packageLoader);
-        const installer = installerId ? PackageInstallers[installerId] : undefined;
+        const installer = getPluginInstaller(GameManager.activeGame.packageLoader);
```
ebkr, line 376: Do we actually need to do this here and below? Feels like we can just immediately return at the top if there's no installer provided?

The only kicker is that we have:
```ts
    private async uninstallModLoaderWithInstaller(mod: ManifestV2, profile: ImmutableProfile): Promise<boolean> {
        const modLoader = this.getModLoader(mod);
        const installerId = modLoader ? GetInstallerIdForLoader(modLoader.loaderType) : null;
        return this.uninstallWithInstaller(installerId, mod, profile);
    }
```
The call to `this.getModLoader` returns a `ModLoaderPackageMapping` type.
The call to `this.getPackageLoaderInstaller` returns a `PackageInstaller`

I haven't looked too deeply, but any idea why we don't get a type mismatch error? Is it expected to get both?

ebkr (approved):
Mostly looks good. One main comment on weird existing code flow. I don't think we need to address it in this PR but worth looking in to / raising.

Minor nit: Is there a way that we can have correct casing on the PackageLoader values? I'm thinking either `BepInEx` or `BEPINEX` is nicer than `Bepinex`. I'm assuming generated values and the answer is no.

src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
anttimaki: > Do we actually need to do this here and below? ...

Yes, I kept these changes intentionally out from this PR.

For `getPackageLoaderInstaller` my idea was to implement "NoOpInstaller" that is returned for PackageLoader.None. Basically for games using it, there shouldn't be package loader package that ever gets called anyway. For plugins, I though about making `getPluginInstaller` to return "InstallRuleInstaller", or something, that basically wraps the current "legacy" installation method to `PackageInstaller` class.

This way both of those "get installer" methods would always return a non-null value which would make the code flow in this file cleaner. How does that sound?

anttimaki: > The call to this.getModLoader returns a ModLoaderPackageMapping type.
> The call to this.getPackageLoaderInstaller returns a PackageInstaller
>
> I haven't looked too deeply, but any idea why we don't get a type mismatch error? Is it expected to get both?

Not sure I understand the question. Despite the odd naming, `getModLoader` basically checks if the `mod` argument is a package loader package, and returns the `ModLoaderPackageMapping`.  `getPackageLoaderInstaller` then uses data in that mapping to return the PackageInstaller instance. So the issue, which I think has been acknowledged earlier, that each package's manifest.json should basically define the installation method which should be used with it. That way we wouldn't need this whole complex mapping of things to things stuff.

anttimaki:
> Minor nit: Is there a way that we can have correct casing on the PackageLoader values? I'm thinking either BepInEx or BEPINEX is nicer than Bepinex. I'm assuming generated values and the answer is no.

Well yes, at this point when the feature isn't really in use, we can control that on ecosystem-schema repo side. I'd have to check with Mythic why he picked the lowercase strings. It might be just because that's what the original `GetInstallerIdForLoader` functions in r2mm used in the mapping function.

**Edit**: discussed this with Mythic. Basically he wants the ecosystem to use slugified (kebab-case) strings. He doesn't care too much what mod managers and other clients do with those. So currently the lib generating the types seems to use capitalize-word by default, so e.g. `return-of-modding` would become `ReturnOfModding`, but `bepinex` would become `Bepinex`. I'll see if the lib support adjusting the logic for enum key generation.

anttimaki:
Update to the above, QuickType doesn't really support adjusting how the output types are generated, but we can adjust the types with regexp before they're written to disk. Using UPPER_SNAKE_CASE makes sense here as it's easy to generate from kebab-case, and it's also what the code base used previously.

I'm not entirely sure what's the easiest way to introduce the changes but I'll likely end up combining this and the previous PR in the stack and doing the changes there.

anttimaki:
Closing in favour of #1821

PR #1806, opened 2025-06-06
Feature/vue27/card components

ebkr:


src/components/ExpandableCard.vue
```diff
@@ -39,52 +39,47 @@
...
-
-        @Prop()
-        id: string | undefined;
-
-        @Prop({default: false})
-        allowSorting: boolean | undefined;
+type ExpandableCardProps = {
+    image: string;
+    description: string;
+    id: string;
```
anttimaki, line 52: Should this be `string|undefined` as `undefined` is used as the default value?

src/components/ExpandableCard.vue
```diff
@@ -39,52 +39,47 @@
...
+    image: '',
+    description: '',
+    id: undefined,
+    allowSorting: false,
+    enabled: true,
+})
 
-        // Keep track of visibility
-        visible: boolean | undefined = false;
+const visible = ref<boolean | undefined>(false);
```
anttimaki: Should this be just `ref<boolean>` as `undefined` was dropped from the prop type?

src/components/ExpandableCard.vue
ebkr: In terms of direct porting, yes. It is always defined though so it seems like we should just keep it as a mandatory parameter.

src/components/ExpandableCard.vue
ebkr: It was originally like this due to the settings not necessarily having had migrated. I think it was extra safe-guarding. 

Since this now relies on the store, I'm happy to drop the undefined type.

PR #1810, opened 2025-06-10
Corrected IDs to remove type error

ebkr:


anttimaki (approved):
Approved but I do get type warnings from the OnlineRowCard component too.

PR #1821, opened 2025-06-18
Use enums provided Thunderstore ecosystem

anttimaki:
- This PR replaces #1751 and #1805
- Main difference is in the second commit, where regexp is used cast Enum keys to match the values used by r2mm
- Rest of the commits are the same, except a lot of enum renaming is no longer needed


ebkr (approved):
I've had a look through and it looks good. I'm not entirely sold on `enumKeysToUpperSnakeCase` however I can't think of a better alternative.

PR #1823, opened 2025-06-20
Quasar upgrade

ebkr:
Upgrades to the latest version of Quasar, which gives us the following benefits:

- Vue 3, which comes with:
  - Security fixes
  - Access to new Vue features such as Suspense and Teleport
  - Better performance / lower memory usage
  - Back onto a supported project as Vue 2 was EOL (Dec 31st, 2023).
- Vite builds
  - Performance is signficantly quicker (30-60s down to approx < 5s)
- IPC calls
  - This is a change that has been needed for an incredibly long time, but we've now been forced to implement IPC to call some of the internal node libraries. This is also a security change.
    - This is somewhat voided until we can figure out the CORS issue, however I don't think it's a huge concern for now.

There are an insane number of files changed. This PR is best reviewed on a per-commit basis, with comments addressing the relevant files on the main PR rather than individual commits, as there may be changes later that have already been reflected.

A large number of these files won't actually need much reviewing as they are things such as moved files (EG: Every game image has been moved), or imports being changed to use our new IPC based calls.

- [x] Rebase onto develop to keep game file location up to date


Lord-Kamina:
I think you deleted some files that shouldn't have been deleted in 06adc951effeca94a9a010f650b3558e8b1bfff5, like the babel and jest config files.

Also, you should probably update babel because you're using a version from roughly 6 years ago.

ebkr:
> I think you deleted some files that shouldn't have been deleted in [06adc95](https://github.com/ebkr/r2modmanPlus/commit/06adc951effeca94a9a010f650b3558e8b1bfff5), like the babel and jest config files.

This was intentional as we no longer need Jest or Babel. It now uses Vite and so we're using Rollup, and for tests we're able to switch to Vitest or just use `node:test`.

> Also, you should probably update babel because you're using a version from roughly 6 years ago.

If anything I just need to remove the dependency for it now. I'll do this as part of a review round.

Lord-Kamina:
> > I think you deleted some files that shouldn't have been deleted in [06adc95](https://github.com/ebkr/r2modmanPlus/commit/06adc951effeca94a9a010f650b3558e8b1bfff5), like the babel and jest config files.
> 
> This was intentional as we no longer need Jest or Babel. It now uses Vite and so we're using Rollup, and for tests we're able to switch to Vitest or just use `node:test`.
> 
> > Also, you should probably update babel because you're using a version from roughly 6 years ago.
> 
> If anything I just need to remove the dependency for it now. I'll do this as part of a review round.

Oh ok. Nevermind then

JamesKoenig:
> This was intentional as we no longer need Jest or Babel. It now uses Vite and so we're using Rollup, and for tests we're able to switch to Vitest or just use `node:test`.
> [...]
> If anything I just need to remove the dependency for it now. I'll do this as part of a review round.

Is the clearing of the configurations supposed to be sort of like a redtag?  To force the unit tests to fail until those dependency changes are addressed?

### Like you said, the dependencies are still in `package.json`:
https://github.com/ebkr/r2modmanPlus/blob/04a868a720f2124a75ef0897f6a422d7bd411b05/package.json#L74-L75
https://github.com/ebkr/r2modmanPlus/blob/04a868a720f2124a75ef0897f6a422d7bd411b05/package.json#L91-L92
https://github.com/ebkr/r2modmanPlus/blob/04a868a720f2124a75ef0897f6a422d7bd411b05/package.json#L100-L102
https://github.com/ebkr/r2modmanPlus/blob/04a868a720f2124a75ef0897f6a422d7bd411b05/package.json#L116

### And the `test:unit:ci` script still runs `jest --ci`:
https://github.com/ebkr/r2modmanPlus/blob/04a868a720f2124a75ef0897f6a422d7bd411b05/package.json#L22

### Which given that is called in the github workflow:
https://github.com/ebkr/r2modmanPlus/blob/04a868a720f2124a75ef0897f6a422d7bd411b05/.github/workflows/test.yml#L37-L39

This results in the two starting unit tests failing, while the latter two build tests succeed.

Along the same vein, would the unit tests like the ones in [`https://github.com/ebkr/r2modmanPlus/tree/develop/test/jest/__tests__`](https://github.com/ebkr/r2modmanPlus/tree/develop/test/jest/__tests__) need to be migrated to vitest?

ebkr:
> Is the clearing of the configurations supposed to be sort of like a redtag? To force the unit tests to fail until those dependency changes are addressed?

Honestly no, it's just that I'd like this reviewed before I make any further changes. Tests aren't essential to the Quasar upgrade changes and can be done separately.

They're left-over for now because I haven't gotten to the test changes yet.

src-electron/ipc/node-fs-impl.ts
```diff
@@ -0,0 +1,151 @@
...
+    return fs.promises.mkdir(mkdirPath, { recursive: true });
+}
+
+async function exists(path: string) {
+    return fs.promises.access(path, fs.constants.F_OK)
+        .then(() => true)
+        .catch(() => false);
+}
+
+function generateInlineStat<T extends fs.StatsBase<number>>(statLike: fs.Stats): T {
```
anttimaki: I'm not entirely sure what your intent here is, but none of the callsites actually pass a type for `T` so return type is effectively `fs.StatsBase<number>`, but the return value doesn't actually match the type since the methods are replaced with booleans. The error is then hidden with the `as T` assertion.

If we want to override the type, would something like this make the intent clearer without having to define types on the callsites:

```typescript
type InlineStat = Omit<fs.Stats, 'isDirectory' | 'isFile'> & {
    isDirectory: boolean;
    isFile: boolean;
};

function generateInlineStat(stats: fs.Stats): InlineStat {
    return {
        ...stats,
        isDirectory: stats.isDirectory(),
        isFile: stats.isFile(),
    };
}
```

Although the second question is: does it make sense to "inline" the stats? Wouldn't someone calling `node:fs:stat` over IPC assume the typing to match whatever Node has? Or is this a IPC serialization thing, in which case name like `SerializableStats` might be a better name? And finally, if it's about serialization, should we drop the other methods present in `fs.stats` from the returned object?

src-electron/ipc/zip-hook.ts
```diff
@@ -0,0 +1,79 @@
...
+    });
+
+    ipcMain.on('zip:create:new', (event) => {
+        const identifier = zipCreatorIdentifier++;
+        zipCreatorCache.set(identifier, new AdmZip());
+        event.returnValue = identifier;
+    });
+
+    ipcMain.on(`zip:create:addBuffer`, (event, identifier, fileName: string, data: Buffer) => {
+        const zip = zipCreatorCache.get(identifier)!;
```
anttimaki: Minor: I wouldn't be against checking for `undefined` here and the two other functions below, instead of using `!`.

src-electron/preload/node-fs.ts
```diff
@@ -0,0 +1,243 @@
+import { ipcRenderer } from 'electron/renderer';
```
anttimaki, line 1: Should we use IPC's [invoke+handle](https://www.electronjs.org/docs/latest/tutorial/ipc#pattern-2-renderer-to-main-two-way) approach rather than send+on(ce)? Looks like it's recommended for a two-way communication, like the stuff that's used e.g. on this file.

src/pages/Profiles.vue
```diff
@@ -163,4 +129,23 @@ async function backToGameSelection() {
     await ManagerSettings.resetDefaults();
     await router.push({name: "index"});
 }
+
+onMounted( async () => {
```
anttimaki, line 130: - How certain are you that it's safe to drop settings the default paths here?
- If the default paths are not set here, do we need to load the settings here at all?
- Overall I think changing to .then() made the code harder to reason about (might be just me), e.g. I'm not sure whether setSelectedProfile has finished before updateProfileList is called and if there's scenarios where it would matter

src/providers/generic/zip/AdmZipProvider.ts
```diff
@@ -1,37 +1,25 @@
...
+    // Type is Promise<Buffer | null>
+    async readFile(zip: string | Buffer, file: string): Promise<any> {
+        return window.zip.readFile(zip, file);
     }
 
-    async getEntries(zip: string | Buffer): Promise<ZipEntryInterface[]> {
-        const adm = new AdmZip(zip);
-        return (adm.getEntries() as unknown as ZipEntryInterface[]);
+    // Type is Promise<ZipEntryInterface[]>
+    async getEntries(zip: string): Promise<any> {
```
anttimaki: Seems a bit silly to use `as ZipEntryInterface[]` but then define the return type as `Promise<any>`?

src/providers/node/fs/NodeFsImplementation.ts
```diff
@@ -0,0 +1,97 @@
+import { NodeFsProvider } from './fs';
+import Lock from 'async-lock';
+
+const lock = new Lock();
+
+function acquireLockAndDo(path: string, action: (resolve: any, reject: any) => Promise<any>): Promise<any> {
```
anttimaki, line 6: - IIRC I've at some point somewhere fixed a bug which was caused by acquiring the lock for only one file for two-file operation (rename, copy...). `lock.acquire` accepts an array of paths, so it should probably be used here too
- Since lock.acquire already returns a promise it seems unnecessary to wrap it in a separate promise and pass resolve/reject around. At the time of this writing I haven't still managed to get the app running locally so I can't fiddle around to confirm
- My electron knowledge is quite limited - is it safe for us to assume that having this locking mechanism here is adequate given the multiple processes (main, renderer(s)) that will be running? Is the provider pattern enough to guarantee everyone always uses the same lock?

src/providers/generic/zip/AdmZipProvider.ts
```diff
@@ -1,37 +1,25 @@
...
 import ZipBuilder from './ZipBuilder';
 import ZipEntryInterface from './ZipEntryInterface';
 
 export default class AdmZipProvider extends ZipProvider {
 
     async extractAllTo(zip: string | Buffer, outputFolder: string): Promise<void> {
-        const adm = new AdmZip(zip);
-        outputFolder = outputFolder.replace(/\\/g, '/');
-        adm.extractAllTo(outputFolder, true);
+        return window.zip.extractAllTo(zip, outputFolder);
```
anttimaki, line 8: Not limited to this file or row, but is it just my IDE that has trouble figuring out the stuff augmented to `window`? It says `Property 'zip' does not exist on type 'Window & typeof globalThis'.` and then it gets typed as any and that propagates to all the methods in `zip` and wrecks type safety.

Can this be remedied with an IDE plugin or config, or via .d.ts file or something?

src/r2mm/mods/ProfileModList.ts
```diff
@@ -283,17 +284,29 @@ export default class ProfileModList {
...
-            path.resolve(profile.getProfilePath(), "BepInEx", "plugins", mod.getName(), "icon.png"),
+            path.join(profile.getProfilePath(), "BepInEx", "plugins", mod.getName(), "icon.png"),
             path.join(PathResolver.MOD_ROOT, "cache", mod.getName(), mod.getVersionNumber().toString(), "icon.png"),
         ]
 
-        for (const iconPath of paths) {
-            if (await FsProvider.instance.exists(iconPath)) {
-                mod.setIcon(iconPath);
-                return;
+        return new Promise(async (resolve, reject) => {
```
anttimaki: I feel the extra promise inside already async function just makes things unnecessarily difficult. It shouldn't be needed and the mixing of break and resolve opens up room for confusion.

src/r2mm/mods/ProfileModList.ts
```diff
@@ -283,17 +284,29 @@ export default class ProfileModList {
...
+            try {
+                for (const iconPath of paths) {
+                    const exists = await FsProvider.instance.exists(iconPath);
```
anttimaki: We're already inside a `try` so this `exists` is probably an unnecessary disk operation? Just try to read the file and let catch handle it not existing.

src/r2mm/mods/ProfileModList.ts
```diff
@@ -283,17 +284,29 @@ export default class ProfileModList {
...
+                    if (exists) {
+                        const content = await FsProvider.instance.base64FromZip(iconPath);
+                        mod.setIcon(`data:image/png;base64,${content}`);
+                        resolve(true);
+                        break;
+                    }
+                }
+                resolve(false);
```
anttimaki: Since the first resolve doesn't return from the function, this resolve also gets called every time.

src/r2mm/mods/ProfileModList.ts
```diff
@@ -283,17 +284,29 @@ export default class ProfileModList {
...
+            } catch (e) {
+                reject(e);
```
anttimaki: Since the for-loop is inside the try-block rather than the other way around,earlier path in `paths` throwing means the rest of the paths aren't even attempted.

src/r2mm/mods/ProfileModList.ts
```diff
@@ -283,17 +284,29 @@ export default class ProfileModList {
...
             }
-        }
-
-        mod.setIcon(FALLBACK_ICON);
+        }).then(resolved => {
+            if (!resolved) {
+                mod.setIcon("/unknown.png");
+            }
+        });
```
anttimaki: Rejections inside the promise aren't handled, leaving the icon potentially empty.

anttimaki: Combining the above comments, I'd keep this function closes to it's previous state. Unless I'm missing something, this should have the same intent but be clearer:

```typescript
    public static async setIconPath(mod: ManifestV2, profile: ImmutableProfile): Promise<void> {
        const paths = [
            path.join(profile.getProfilePath(), "BepInEx", "plugins", mod.getName(), "icon.png"),
            path.join(PathResolver.MOD_ROOT, "cache", mod.getName(), mod.getVersionNumber().toString(), "icon.png"),
        ]

        for (const iconPath of paths) {
            try {
                const content = await FsProvider.instance.base64FromZip(iconPath);
                mod.setIcon(`data:image/png;base64,${content}`);
                return;
            } catch (err: any) {
                continue;
            }
        }

        mod.setIcon("/unknown.png");
    }
```

Also worth noting that TSMM side has ModIconCache thingie which might or might not make sense on r2 too. I can't recall why it's implemented only on TSMM currently.

src/store/modules/DownloadModule.ts
```diff
@@ -273,6 +273,10 @@ export const DownloadModule = {
                 state.allDownloads = newDownloads;
             }
         },
+        addDownload(state: State, download: DownloadProgress) {
+            // @ts-ignore
```
anttimaki: What exactly are we ignoring here? Looks valid to me.

src/store/modules/ProfileModule.ts
```diff
@@ -467,6 +467,7 @@ export default {
         },
 
         async updateModList({commit}, modList: ManifestV2[]) {
+            commit('tsMods/prewarmCacheMod', modList, { root: true });
```
anttimaki, line 470: Nitpick: since updateModList is called in places where were already dealing with mods that are being used in the UI right away, this is not so much prewarming the cache as just updating it. Might consider renaming it at some point.

src/store/modules/ProfilesModule.ts
```diff
@@ -92,10 +95,11 @@ export const ProfilesModule = {
 
         async updateProfileList({commit, rootGetters}) {
             const profilesDirectory = Profile.getRootDir();
-
+            await FileUtils.ensureDirectory(profilesDirectory);
             let profilesDirectoryContents = await FsProvider.instance.readdir(profilesDirectory);
             let promises = profilesDirectoryContents.map(async function(file) {
-                return ((await FsProvider.instance.stat(path.join(profilesDirectory, file))).isDirectory() && file.toLowerCase() !== 'default' && file.toLowerCase() !== "_profile_update")
+                const fileStat = await FsProvider.instance.stat(path.join(profilesDirectory, file));
+                return (fileStat.isDirectory() && file.toLowerCase() !== 'default' && file.toLowerCase() !== "_profile_update")
```
anttimaki, line 102: Is this the sort of fileStat where `isDirectory` method was overwritten with the boolean value? Or do we have multiple FS implementations? Does store modules have access to node methods without having to go through IPC? Where do the boundaries between the renderer and main process actually stand? 😕

src/store/modules/SplashModule.ts
```diff
@@ -41,6 +63,9 @@ export const SplashModule = {
 
             return item;
         },
+        updateRequestItem({ commit }, body: UpdateRequestItemBody) {
```
anttimaki: Is this wrapper actually needed? Why not call the commit directly?

src/store/modules/SplashModule.ts
```diff
@@ -59,21 +84,23 @@ export const SplashModule = {
...
             } catch (e) {
                 console.error('SplashModule failed to fetch mod list index from API.', e);
                 return undefined;
             } finally {
-                const requestItem = await dispatch('getRequestItem', 'PackageListIndex') as RequestItem;
-                requestItem.setProgress(100);
+                await dispatch('updateRequestItem', {
+                    requestName: 'PackageListIndex',
+                    value: 100
+                } as UpdateRequestItemBody);
```
anttimaki, line 96: Is this an attempt to try to improve the type safety e.g. in case of using `value: "100"` or what?

src/store/modules/TsModsModule.ts
```diff
@@ -192,6 +176,28 @@ export const TsModsModule = {
         },
         updateDeprecated(state, allMods: ThunderstoreMod[]) {
             state.deprecated = Deprecations.getDeprecatedPackageMap(allMods);
+        },
+        prewarmCacheMod(state: State, mods: ThunderstoreMod[]) {
```
anttimaki, line 181: I made an earlier comment about this method "no longer" being about prewarming the cache, but I confused it with the old `prewarmCache` below. Still, this is more about updating the cache so another name might be better?

src/store/modules/TsModsModule.ts
```diff
@@ -192,6 +176,28 @@ export const TsModsModule = {
...
+            const localState = new Map<string, CachedMod>(state.cache.entries());
```
anttimaki, line 182: Just wondering if changing from mutating the cache Map to duplicating it might lead to performance issues on large profiles or race conditions if the commit is called in rapid succession 🤷🏻

src/store/modules/TsModsModule.ts
```diff
@@ -192,6 +176,28 @@ export const TsModsModule = {
...
+            mods.forEach(mod => {
+                const cacheKey = `${mod.getName()}-${mod.getVersionNumber()}`;
+
+                if (localState.get(cacheKey) === undefined) {
+                    const tsMod = state.mods.find((m) => m.getFullName() === mod.getName());
+
+                    // Updating Vuex state directly instead of mutations is a bad
```
anttimaki: After the changes this comment is incorrect in at least three different ways and should be removed.

anttimaki (commented):
- A collection of artisanal comments with varying importance
- I omitted nitpicks like missing semicolons and stuff like that, and ignored pretty much all CSS changes
- There were some TODO comments and console loggings that seemed they were intended to be cleaned up before a PR
- My main interest at this point is:
  1. Getting the project to run locally
  2. Will TSMM work? Do you have a plan for
  3. After the changes there seems to be a lot of warnings/errors on my IDE (Cursor) that bury the actual problems and weaken the type safety. Ideally those would be resolved one way or another and sooner rather than later

src-electron/ipc/node-fs-impl.ts
ebkr: Great catch, I wasn't aware of `Omit`. It fixes the typing errors I was getting previously (hence the `as unknown as T`).

Agreed on the `SerializableStats` naming too.


As for dropping the other methods, I don't think so. If we want to access anything added to the Node API then we can become aware of it this way, we'll just get no response and it means we have to add support for using it. I don't think it's particularly bad because we are still theoretically allowing the use of the Node API, it's just abstracted for security reasons.

src-electron/preload/node-fs.ts
ebkr: Ah it looks like a Promise version of send/on. Will refactor 👍

Lord-Kamina:
I don't know enough TS or JS to really make many meaningful comments here but just a minor BTW: In my local fork (of course not including this PR) I updated typescript to the latest 4.9.x and everything worked fine; in case doing so here might make any refactorings easier.

src/pages/Profiles.vue
ebkr: > How certain are you that it's safe to drop settings the default paths here?
 
1. Our call to `settings.load()` is useless, so we can remove this part.
  - We load it when getting the singleton anyway: Image 
2. I've manually tested all modals and they appear to behave as expected with the change. No difference.
3. The only use it has that I have been able to find is in `ProfileModule` where it's used as part of `generateTroubleshootingString` - which is only accessible from the Manager.vue screen

Here are the direct usages of the steam directory
Image

Likewise for the game directory usages
Image

> If the default paths are not set here, do we need to load the settings here at all?

I'm not sure. I decided to leave that one as the usages are a bit more unclear given it's store related.

Image

> Overall I think changing to .then() made the code harder to reason about (might be just me), e.g. I'm not sure whether setSelectedProfile has finished before updateProfileList is called and if there's scenarios where it would matter

A dispatch returns a Promise, so this would essentially be:

- load settings
- get last selected profile name
- set selected profile
- update profile list

Each one is sequential as every `.then` depends on the one prior to be resolved.

I don't overly have a preference between lambda-style vs await. I'm happy to change it as `await` has the same thread behaviour as the lambda style one.

src/providers/generic/zip/AdmZipProvider.ts
ebkr: This was a file commented out quite early on - It was likely a remnant from debugging. Will fix.

src/providers/node/fs/NodeFsImplementation.ts
ebkr: > IIRC I've at some point somewhere fixed a bug which was caused by acquiring the lock for only one file for two-file operation (rename, copy...). lock.acquire accepts an array of paths, so it should probably be used here too

Sweet, will change this to match.

> Since lock.acquire already returns a promise it seems unnecessary to wrap it in a separate promise and pass resolve/reject around. At the time of this writing I haven't still managed to get the app running locally so I can't fiddle around to confirm

I can have a play around with removing the wrapper. It felt like a nice way to defer control to the consumer but there doesn't really seem like it's actually being used.

> My electron knowledge is quite limited - is it safe for us to assume that having this locking mechanism here is adequate given the multiple processes (main, renderer(s)) that will be running?

All locks happen on the frontend currently, so we don't have a concern about the two manipulating files independently. We should probably move them to the backend as that means we can keep the renderer process running without halting however this is how it previously behaved. Since it's been done like-for-like here, it feels slightly out of scope for the upgrade. It also means we don't need to wildly rework TSMM's provider implementations for now.

> Is the provider pattern enough to guarantee everyone always uses the same lock?

Yeah, it's always the same instance that gets returned.

ebkr: So to add on to:
> Since lock.acquire already returns a promise it seems unnecessary to wrap it in a separate promise and pass resolve/reject around. At the time of this writing I haven't still managed to get the app running locally so I can't fiddle around to confirm

I've had a look into this and I can't quite work out why it's not working if we remove the wrapper behaviour. The promise doesn't resolve as intended if we just let it do it's own thing.

I'll leave it as-is for now, it would be good if someone could get around into looking as to why it breaks.

When changing the implementation to remove the wrapper, the manager starts up as a white screen and fails to progress into the standard index page.

src/providers/generic/zip/AdmZipProvider.ts
ebkr: We can likely remedy it with a `d.ts`

I haven't looked into it because I don't expect us to use it outside of a single file definition. I'm happy to look into it though.

src/r2mm/mods/ProfileModList.ts
ebkr: I'll refactor to make it clearer and avoid it as a code path but promises don't return the values from subsequent resolves.

src/store/modules/ProfilesModule.ts
ebkr: > Is this the sort of fileStat where isDirectory method was overwritten with the boolean value?

Yes, they get re-assigned to become their appropriate methods again, see NodeFsImplementation:
image

> Does store modules have access to node methods without having to go through IPC? 

No, it still calls into NodeFsImplementation.

Nothing on the frontend has access to Node libraries anymore. It results in app errors. The IPC change was mandatory.

> Where do the boundaries between the renderer and main process actually stand?

It's pretty much just that the main process has to do whatever native node libraries did now. It does mean that since all of our IO is now on the backend that we can theoretically get a more responsive frontend, but we still have to wait for the results from any IO operations so it's not that noticeable.

src/store/modules/SplashModule.ts
ebkr: It's called in one place on the frontend.

We could move it to a separate action, although I think that's an unrelated refactor.

image

src/store/modules/SplashModule.ts
ebkr: No, we aren't allowed to modify the RequestItem object in the same way now because VueX has added constraints in the later versions. We have to do these changes in a commit (this dispatch goes to a commit, related to the comment above - this one could commit directly though).

The actual type definitions are just because we don't have a use for anything other than number, and so make sense to have anyway.

src/pages/Profiles.vue
anttimaki: > Each one is sequential as every .then depends on the one prior to be resolved.

Do they though? Unless I'm mistaken, the way things are now chained, `updateProfileList` will be called immediately, even if `setSelectedProfile` hasn't finished execution, because `setSelectedProfile` isn't awaited and it's wrapping function doesn't return a promise (i.e. it's resolved immediately as undefined).

src/store/modules/SplashModule.ts
anttimaki: Umm, why not call the commit from `Splash.vue`? Has there been some change to this on Vuex upgrade?

src/store/modules/TsModsModule.ts
ebkr: This change was to mitigate a performance issue because it directly affects the `cachedMod` getter - Whenever the cache updates, the getter does a refresh too.

It's a valid concern about race conditions. I think we have the following options:
1. Wrap in a lock
  - IMO less ideal, but makes it easy to manage
2. Diff the cache and only re-set the changes - I'm unsure if calling `.set` re-triggers an update if the value is the same.
3. Keep the behaviour the same as before (always just call `.set`)
4. Something else?

src/store/modules/SplashModule.ts
ebkr: Huh that's fair, will change

src/pages/Profiles.vue
ebkr: It may just be a bad execution, although it should in theory because each .then uses the return result of the previous.

This has been changed now to be linearly executed though, so if it was performing incorrectly then it's now resolving as intended.

src/store/modules/ProfileModule.ts
ebkr: I'll leave these as they are for now since it feels out of scope to handle renaming all the usages of this - I do agree though.

src/pages/Profiles.vue
anttimaki: > because each .then uses the return result of the previous.

To clarify what I tried to say above. As far as I understand promises, this

```typescript
(profileName: string) => {
    if (store.state.tsMods.cache.size === 0) {
        setSelectedProfile(profileName, false);
    }
}
```

always returns an implicit `undefined`. Because it doesn't return a promise, the wrapping `.then()` immediately returns a resolved promise, and `.then(updateProfileList)` immediately starts execution (whether `setSelectedProfile` has finished or not).

If we'd want to avoid a possible race condition between `setSelectedProfile` and `updateProfileList`, the code above should be changed to

```typescript
(profileName: string) => {
    if (store.state.tsMods.cache.size === 0) {
        return setSelectedProfile(profileName, false);  // Now returns an unresolved promise
    }
}
```

This way the next then wouldn't start before `setSelectedProfile` has resolved.

anttimaki: And the stuff above is the kind of stuff that makes me prefer `await` approach 😆

src/store/modules/TsModsModule.ts
anttimaki: Well to think of it, the mutation is synchronous and the environment is single-threaded by nature, and only other place that changes `state.cache` is `clearModCache` mutation, so maybe we're safe in race condition regard?

src/providers/node/fs/NodeFsImplementation.ts
Lord-Kamina: Could you post a snippet of how you're going about it? Mixing using async and promises is fraught with gotchas and sometimes another set of eyes can help.

src-electron/electron-main.ts
```diff
@@ -40,16 +44,21 @@ function createWindow() {
...
-            webSecurity: false,
-            contextIsolation: false
-        },
-        icon: path.join(__dirname, 'icon.png'),
-        autoHideMenuBar: process.env.PROD
+            preload: path.resolve(
+                fileURLToPath(new URL('.', import.meta.url)),
+                path.join(process.env.QUASAR_ELECTRON_PRELOAD_FOLDER, 'electron-preload' + process.env.QUASAR_ELECTRON_PRELOAD_EXTENSION)
+            ),
+            // TODO - Remove and find an appropriate workaround for CORS blocking by Electron
```
anttimaki, line 54: How confident are you in merging this PR before this TODO is resolved? Electron docs specifically recommend against doing this. I get the feeling that if we must rely on this, we're probably doing something wrong. Could it at least be limited to dev builds?

src/r2mm/launching/instructions/instructions/loader/CustomInstructions.ts
```diff
@@ -1,31 +1,39 @@
...
 import { GameInstruction } from "../../GameInstructions";
 import { DynamicGameInstruction } from "../../DynamicGameInstruction";
+import path from '../../../../../providers/node/path/path';
 
 export class CustomInstructions extends GameInstructionGenerator {
-    private readonly instructions: GameInstruction;
+    private readonly instructions: () => GameInstruction;
 
-    constructor(instructions?: GameInstruction) {
+    constructor(instructions?: () => GameInstruction) {
```
anttimaki, line 11: Is the change to function form purely for optimization or does it serve some other purpose?

anttimaki (commented):
- Generally speaking it seems there's still some places were code is wrapped in new promises explicitly, and where then-chains are used when not necessary, It's probably apparent from earlier comments but I'm not personally a fan of that approach. Not a blocker for this PR but we should probably at least have a discussion about the subject on how the matter should be handled in code and code reviews going forward.
- When manually testing the dev build, page transitions feel a bit sluggish compared to pre-upgrade version. Like they take some 100-200ms longer. I might be hallucinating it. Or it could be vite doing some stuff JIT. Or the changes to the router? Have you noticed this?

src-electron/electron-main.ts
ebkr: Pre upgrade we had security disabled and so I'm not hugely concerned about this for now as part of the upgrade. Since it's an upgrade I don't think it matters too much as we're on-par with prior behaviour.

In theory we can flick the switch and things will work fine, we just need to resolve the CORS issue which we can do separately.

src/r2mm/launching/instructions/instructions/loader/CustomInstructions.ts
ebkr: This was required due to timings. The constructors were called before we could assign the path implementation and so we now lazy load them so that the path library can be bound before use and once actually needed.

ebkr:
>  When manually testing the dev build, page transitions feel a bit sluggish compared to pre-upgrade version. Like they take some 100-200ms longer. I might be hallucinating it. Or it could be vite doing some stuff JIT. Or the changes to the router? Have you noticed this?

No you're absolutely correct. I spent a while playing around with it and it appears to be a VueX issue which is what prompted me to bring up the discussion internally.

To recap:
A commit in the Vue 3 supported version of VueX takes approximately 150ms. If we move to something like globally exported reactives then we bring that number down to quite literally 5ms.

MobX has a greater overhead in terms of developer implementation time however the time to update is double that of globally exported reactives, but still more than acceptable at 10ms.

Lord-Kamina:
So... I took a stab at removing all those promises within promises within promises and here's what I came up with.
Mind you, I can't figure out a way to test it other than launching the app, because in early stages, even something that I would have thought wouldn't compile just ran fine (but refused to work as expected)

For this patch, there could of course be errors (I'd be surprised if there aren't really) but I tried to launch, create a new profile, download some mods and export the profile as a file and seemingly, all that worked.

At the very least, it might give you a base to get rid of all those extra `new Promise`

[0001-Remove-superfluous-promises.patch](https://github.com/user-attachments/files/21978759/0001-Remove-superfluous-promises.patch)

ebkr:
> So... I took a stab at removing all those promises within promises within promises and here's what I came up with. Mind you, I can't figure out a way to test it other than launching the app, because in early stages, even something that I would have thought wouldn't compile just ran fine (but refused to work as expected)
> 
> For this patch, there could of course be errors (I'd be surprised if there aren't really) but I tried to launch, create a new profile, download some mods and export the profile as a file and seemingly, all that worked.
> 
> At the very least, it might give you a base to get rid of all those extra `new Promise`
> 
> [0001-Remove-superfluous-promises.patch](https://github.com/user-attachments/files/21978759/0001-Remove-superfluous-promises.patch)

Instead of doing a patch, do the following:
1. Fork the repository
2. Clone your fork
3. Apply the patch to your fork
4. Commit and push
5. Create a PR from your forked repository into this repository

That way:
- You'll get to receive contributions identified by GitHub
- I can review it easier
- Any changes that are requested can be kept track of as part of the patch

PR #1829, opened 2025-06-23
Display correct information when there are no mods to update and the …

VilppeRiskidev:
…UpdateAllInstalledModsModal is opened.

* Previously when there either were no installed mods, or all mods were up to date, the modal produced an error.


anttimaki (approved):
LGTM

PR #1832, opened 2025-06-24
Add a background color to the buttons' sticky toolbar on DownloadMonitor

VilppeRiskidev:
* The background was transparent, thus showing the download list through itself when scrolling


anttimaki (approved):
Dunno where this task came from nor can I see it actually changing anything on the UI but seems harmless to me.

PR #1833, opened 2025-06-25
Auto-focus the name input when CreateProfileModal is opened

VilppeRiskidev:


anttimaki (approved):
@VilppeRiskidev I pushed some changes:

- Added manual typing for the existing computed `isOpen`
- (Switched from `watchEffect` to `watch`)
- Use the already computed `ipOpen` rather than referencing the Vuex store directly a second time (dunno if this matters in practice, but looks cleaner)
- Use conditional to avoid the `!` non-null assertion

PR #1840, opened 2025-06-28
Fix online mod list reactivity

anttimaki:
When watch() is passed a single computed or ref, Vue automatically unwraps the .value of the ref and triggers watch function when it changes. However, if a function that returns an array of refs are passed instead, the wrapping is not done automatically.


anttimaki:
FYI @ebkr

PR #1849, opened 2025-07-15
Moved usages of isProtonGame to safer Utils call

ebkr:


src/utils/LaunchUtils.ts
```diff
@@ -64,3 +66,9 @@ export const throwIfNoGameDir = async (game: Game): Promise<void> => {
         throw error;
     }
 };
+
+export async function isProtonRequired(activeGame: Game) {
+    return activeGame.activePlatform.storePlatform === Platform.STEAM
```
anttimaki: I assume this intentionally doesn't apply to STEAM_DIRECT games?

anttimaki (approved):
Approved with a question.

src/utils/LaunchUtils.ts
ebkr: Good spot, have added.

PR #1850, opened 2025-07-16
Proton settings selection

ebkr:
The previous expected behaviour was to include a `.forceproton` file in the game directory to enable launching the game with Proton. Great, except not entirely. It was obscure and if the manager detected it as a proton game and you wanted it to launch natively then you were out of luck.

This new behaviour gives the user more control and shows exactly which option we're doing when we resolve things ourselves.

This will only appear on non-Windows platforms, and only when the `Steam` platform is selected. Proton also doesn't apply to non-Steam games (to my knowledge) and so it doesn't make sense to include it for any other platform selection.

image

image

image

### DIFF
--- a/src/model/real_enums/launch/LaunchType.ts
+++ b/src/model/real_enums/launch/LaunchType.ts
@@ -18,5 +18,5 @@ export async function getLaunchType(game: Game) {
             return EnumResolver.from<LaunchType>(LaunchType, settings.getContext().gameSpecific.launchType);
         }
     }
-    return LaunchType.AUTO;
+    return EnumResolver.from<LaunchType>(LaunchType, LaunchType.AUTO);
 }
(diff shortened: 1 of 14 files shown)


src/model/real_enums/launch/LaunchType.ts
```diff
@@ -0,0 +1,22 @@
...
+    NATIVE = "Native",
+    PROTON = "Proton",
+}
+
+export async function getLaunchType(game: Game) {
+    if (game.activePlatform.storePlatform === Platform.STEAM) {
+        const settings = await ManagerSettings.getSingleton(game);
+        const savedLaunchType = settings.getContext().gameSpecific.launchType;
+        if (savedLaunchType !== undefined) {
+            return EnumResolver.from<LaunchType>(LaunchType, settings.getContext().gameSpecific.launchType);
```
anttimaki: Nitpick: could use `savedLaunchType` here.

src/pages/Manager.vue
```diff
@@ -128,6 +128,7 @@
         <DownloadProgressModal />
         <DownloadModVersionSelectModal />
         <UpdateAllInstalledModsModal />
+        <LaunchTypeModal/>
```
anttimaki: Would it be more hassle than it's worth to use `v-if` and render the component only on target OS?

anttimaki (approved):
- Not the most thorough review but lgtm
- Has absolute imports here and there though

PR #1854, opened 2025-07-23
Config Editor Rewrite

ebkr:
This is a rewrite of the config editor code to make it more maintained and extensible for future work.

### Key changes:

- Visual change for the "Entry" style editor (BepInEx configs).
- Easier to read
- Decoupled logic from raw/entry components
- The `Show more` action now affects text length as well as line count making it more practical.

### Future changes:

This allows us to add features such as:
- Color pickers
- Multi-select components
- Undo / reset actions

### Screenshots

image

image

### DIFF
--- a/src/components/config-components/ConfigEntryEditor.vue
+++ b/src/components/config-components/ConfigEntryEditor.vue
@@ -4,11 +4,11 @@
             <button class="button is-info margin-right margin-right--half-width" @click="save">Save</button>
             <button class="button is-danger" @click="cancel">Cancel</button>
         </div>
-        <div id="config-entry-main" v-if="configurationFile">
+        <div id="config-entry-main" v-if="configurationFileHolder.configurationFile">
             <div id="config-entry-sections">
                 <h3 class='subtitle is-3'>Sections</h3>
                 <ul>
-                    <template v-for="(section, sectionIndex) in configurationFile.sections" :key="`li-section-${sectionIndex}-${section.sectionName}`">
+                    <template v-for="(section, sectionIndex) in configurationFileHolder.configurationFile.sections" :key="`li-section-${sectionIndex}-${section.sectionName}`">
                         <li v-if="section.sectionName.length > 0">
                             <a :href="`#${sectionIndex}`">{{ section.sectionName }}</a>
                         </li>
@@ -17,7 +17,7 @@
                 <hr/>
             </div>
             <div id="config-entries">
-                <div class="outer-row margin-top margin-right" v-for="(section, sectionIndex) of configurationFile.sections">
+                <div class="outer-row margin-top margin-right" v-for="(section, sectionIndex) of configurationFileHolder.configurationFile.sections">
                     <div class="section-title" :id="sectionIndex.toString()">
                         <p class="title is-6" @click="() => toggleSectionVisibility(section)">{{ section.sectionName }}</p>
                         <p v-if="collapsedSections.includes(section)" class="smaller-font">({{ section.entries.length }} hidden)</p>
@@ -56,6 +56,13 @@
                                         </select>
                                     </div>
                                 </template>
+                                <template v-else-if="entry.displayType === 'multi-select'">
+                                    <MultiSelect
+                                        placeholder="Select an option"
+                                        :selected="entry.value.split(',')"
+                                        :options="getSelectOptions(entry)"
+                                        @selection-changed="(newSelection) => updateEntryMultiSelect(entry, newSelection)"/>
+                                </template>
                                 <template v-else>
                                     <div class="settings-input-container">
                                         <input type="text" class="input" v-model="entry.value"/>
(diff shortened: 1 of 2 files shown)


src/components/config-components/ConfigEntryEditor.vue
```diff
@@ -0,0 +1,165 @@
...
+}
+
+function getAppropriateCommentLines(entry: ConfigurationEntry) {
+    if (!entriesWithExpandedComments.value.includes(entry)) {
+        const commentLines = [...entry.commentLines]
+            .slice(0, 5)
+            .map(value => {
+                return {
+                    ...value
+                };
```
anttimaki, line 144: Nitpick: adding `displayValue: value.displayValue.substring(0, 200)` here would drop the need for the forEach loop.

src/utils/ConfigUtils.ts
```diff
@@ -0,0 +1,158 @@
...
+    displayType: ConfigurationEntryDisplayType;
+}
+
+export type CommentLine = {
+    isDescription: boolean;
+    displayValue: string;
+    rawValue: string;
+}
+
+export async function buildConfigurationFileFromPath(filePath: string): Promise<ConfigurationFile> {
```
anttimaki, line 32: Nitpick: lots of functions marked async in this file but it seems `saveConfigurationFile` is the only one that actually does something async.

src/utils/ConfigUtils.ts
```diff
@@ -0,0 +1,158 @@
...
+    if (comments.findIndex(comment => comment.includes('# Acceptable values:')) >= 0) {
+        return 'single-select';
+    }
+    return 'input';
+}
+
+async function buildComments(comments: string[]): Promise<CommentLine[]> {
+    return comments
+        .filter(value => value.trim().substring(1).length > 0)
+        .map((commentLine) => {
```
anttimaki: Could define the return type here (`(commentLine): CommentLine => {`), then we wouldn't need `as CommentLine` below and it would provide better type safety against missing properties.

test/jest/__tests__/utils/utils.ConfigUtils.ts.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+            cachedValue: "value",
+            commentLines: [],
+        };
+
+        expect(() => getSelectOptions(configurationEntry)).toThrowError(
+            new Error(`Invalid display type for select options. Got [input] for entry: InputEntry`)
+        );
+    });
+
+    test.each([
```
anttimaki: So, these tests work in "standalone r2mm", but if the project has tsmm in parent directory, they cause problems. I've encountered this before, and IIRC it's because there's both Jest and Mocha in the project(s) and their types get mixed somehow. The result is that the tests fail and also `yarn run run` shows compilation errors that prevent using the app.

I've previously bypassed this by replacing `test.each` with 

```typescript
import { it } from '@jest/globals';
it.each( ...
```

This is hacky but fixing the actual issues seems to always be out of scope.

test/jest/__tests__/utils/utils.ConfigUtils.ts.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+            commentLines: [
+                {
+                    rawValue: '# Acceptable values: valueOne, valueTwo, valueThree',
+                    displayValue: 'Acceptable values: valueOne, valueTwo, valueThree',
+                    isDescription: false,
+                }
+            ],
+        };
+
+        expect(getSelectOptions(configurationEntry)).toMatchObject(['valueOne', 'valueTwo', 'valueThree']);
```
anttimaki: I'm not sure if it's caused by using `it` instead of `test`, but `toMatchObject` here gives a type error, as it expects an array of objects, not primitives. Replacing it (also on L235) with `toEqual` seems to solve this issue.

src/utils/ConfigUtils.ts
ebkr: I don't think it's a bad thing to have a bunch of async functions because Node can suspend them when it deems appropriate. It just means after each part is executed there's a chance it may be halted which I think is fine. I don't think I've ever seen any of our promises get notably halted.

test/jest/__tests__/utils/utils.ConfigUtils.ts.spec.ts
ebkr: `it` and `test` are interchangeable and so it should suffice by doing `import { test } from '@jest/globals';` if I'm not mistaken?

src/utils/ConfigUtils.ts
anttimaki: I'm under the impression that JS runs synchronous code to completion without interruptions. To allow suspension, the code itself must contain something that yields the control, like `await` or `setTimeout`. Just marking a function async doesn't change this if the body of the function itself is fully synchronous.

I'm not 100% sure of this, so if you know some docs that sheds some light to the matter, I'd be interested. Couldn't find anything on a quick googling and don't trust sycophant LLMs.

As stated earlier, this is just a nitpick so no need to be blocked about it. Even if I'm right the worst case is that someone might assume the functions do async stuff when they don't.

test/jest/__tests__/utils/utils.ConfigUtils.ts.spec.ts
anttimaki: Welp, seems the test cases run without problems in this branch so this comment is resolved.

src/utils/ConfigUtils.ts
ebkr: I'm not sure. I was under the impression that it ran each promise to completion, and had the ability to suspend a following promise call. Would be good for one of us to look into it. I'll keep it as-is for now

anttimaki: Ah, seems I misunderstood earlier. I was talking about being able to suspense within the function, while your point is to do so between the functions. So in that case it's about how these functions are called, in which case they're fine as is.

PR #1855, opened 2025-07-24
Config Editor: Multi Select

ebkr:
Previously we had shown these as a text field, requiring people to enter the config text in manually.

image

Is it worth also having a text field toggle? I can't see a huge benefit although people may still want it? Thoughts?

### DIFF
--- a/src/components/config-components/ConfigEntryEditor.vue
+++ b/src/components/config-components/ConfigEntryEditor.vue
@@ -103,7 +103,7 @@ const configurationFileHolder = reactive({
 const collapsedSections = ref<ConfigurationSection[]>([]);
 const entriesWithExpandedComments = ref<ConfigurationEntry[]>([]);
 
-buildConfigurationFileFromPath(props.configFile)
+buildConfigurationFileFromPath(props.configFile.getPath())
     .then(value => configurationFileHolder.configurationFile = reactive(value));
 
 function toggleSectionVisibility(section: ConfigurationSection) {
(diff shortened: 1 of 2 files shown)


src/components/config-components/ConfigEntryEditor.vue
```diff
@@ -79,12 +87,14 @@ const emits = defineEmits<{
     (e: 'changed'): void
 }>();
 
-const configurationFile = ref<ConfigurationFile | null>(null);
+const configurationFileHolder = reactive({
+    configurationFile: {} as ConfigurationFile,
```
anttimaki, line 101: Is this for future proofing, or why is the `configurationFile` the only entry in a wrapper object, instead of it being the object passed to `reactive()`?

src/components/input/MultiSelect.vue
```diff
@@ -0,0 +1,61 @@
...
+}>();
+
+const selectableOptions = computed(() => {
+    return props.options
+        .filter(entry => props.selected.findIndex(value => value.trim() === entry.trim()) === -1)
+});
+
+function updateSelection(event: InputEvent) {
+    const target = event.target as HTMLSelectElement;
+    console.log([...props.selected, target.value]);
```
anttimaki: Nitpick: not needed?

src/components/config-components/ConfigEntryEditor.vue
ebkr: It's a Vue 2 quirk - `reactive` wasn't behaving as intended without the wrapper object. I can't quite remember why, but we were also getting warnings with it too.

In theory the migration to Vue 3 would resolve this issue and we can remove the wrapper.

PR #1863, opened 2025-07-28
Rename download module functions

VilppeRiskidev:


src/store/modules/DownloadModule.ts
```diff
@@ -77,27 +77,20 @@ export const DownloadModule = {
             dispatch('retryDownload', { download });
         },
 
-        _addDownload({state}, params: {
+        _addDownload({state}, download: DownloadProgress) {
```
anttimaki, line 80: First, this should be changed from action to a commit. Not only is this pure commit by nature, but if I understood things correctly while reviewing the Quasar upgrade PR, later versions of Vuex don't like state being changed directly from actions.

~~Then, I would change the verb in the name to avoid confusion between "adding" and "creating" and Download. Since this part adds the item to a list for tracking purposes, better pair of verbs for add/remove would IMO be one of:~~
- ~~track/untrack~~
- ~~enlist/delist~~
- ~~register/unregister~~

It should be named simply addDownload.

src/store/modules/DownloadModule.ts
```diff
@@ -77,27 +77,20 @@ export const DownloadModule = {
...
+            state.allDownloads = [...state.allDownloads, download];
+        },
+
+        _createAndAddDownload({dispatch}, params: {
```
anttimaki, line 84: I would drop this action altogether. It's called only from one place, creates confusion with the naming, and isn't that much code to include in `downloadAndInstallCombos` now that the object creation is separated in a helper function.

src/store/modules/DownloadModule.ts
```diff
@@ -120,17 +113,17 @@ export const DownloadModule = {
                 if (!hideModal) {
                     commit('openDownloadProgressModal', null, { root: true });
                 }
-                downloadId = await dispatch('_addDownload', { combos, installMode, game, profile });
```
anttimaki, line 123: To combine the two comments above, this would then be something like:

```typescript
const download = createDownloadProgressObject(combos, installMode, game, profile);
commit('addDownload', download);
const downloadId = download.downloadId;
```

src/store/modules/DownloadModule.ts
```diff
@@ -163,16 +156,16 @@ export const DownloadModule = {
...
                 );
             } catch (e) {
-                commit('setFailed', params.downloadId);
+                commit('setDownloadFailed', params.downloadId);
                 throw e;
             }
         },
 
-        async _installModsAndResolveConflicts({commit, dispatch}, params: {
+        async _installAndResolveConflicts({commit, dispatch}, params: {
```
anttimaki, line 168: Sounds like we're installing conflicts 😄 Would `_installWithConflictResolution` be better or too clunky? `_installThenResolveConflicts`? 🤷🏻

src/store/modules/DownloadModule.ts
```diff
@@ -196,7 +189,7 @@ export const DownloadModule = {
             });
         },
 
-        async _downloadProgressCallback({commit}, params: {
+        async _handleDownloadProgressUpdate({commit}, params: {
```
anttimaki, line 192: `_updateDownloadProgress`?

src/store/modules/DownloadModule.ts
```diff
@@ -260,39 +253,61 @@ export const DownloadModule = {
...
+        setDownloadInstalling(state: State, downloadId: number) {
             state.allDownloads = updateDownloadStatus(state.allDownloads, downloadId, DownloadStatusEnum.INSTALLING);
         },
         // Use actions.toggleIngoreCache to store the setting persistently.
         setIgnoreCacheVuexOnly(state: State, ignoreCache: boolean) {
             state.ignoreCache = ignoreCache;
         },
-        removeAllInactive(state: State) {
-            state.allDownloads = getOnlyActiveDownloads(state.allDownloads);
+        removeAllInactiveDownloads(state: State) {
```
anttimaki, line 282: Isn't the `All` a bit redundant, doesn't `removeInactiveDownloads` mean the same thing?

src/store/modules/DownloadModule.ts
```diff
@@ -302,13 +317,13 @@ function getIndexOfDownloadProgress(allDownloads: DownloadProgress[], downloadId
     return index;
 }
 
-function getOnlyActiveDownloads(downloads: DownloadProgress[]): DownloadProgress[] {
+function filterKeepOnlyActiveDownloads(downloads: DownloadProgress[]): DownloadProgress[] {
```
anttimaki, line 320: I would make this a function that is passed to `.filter()`, i.e:

```typescript
const isActive = (dl: DownloadProgress) => [DownloadStatusEnum.DOWNLOADING, DownloadStatusEnum.INSTALLING].includes(dl.status);
```

Then the callsites would be something like:

```typescript
state.allDownloads = state.allDownloads.filter(isActive);
```

If you find it still ambiguous whether the filter keeps or removes the active ones, the function could be renamed from `isActive` to `removeInactiveDownloads` oslt, (but then the function by itself is a bit oddly named).

anttimaki:
@VilppeRiskidev what's the status of this PR?

VilppeRiskidev:
Closing this as it's just gathering dust here and not really something I'm going to focus on in the near future

PR #1869, opened 2025-08-01
I18n/base

ebkr:
## i18n base

This is a PR for barebones i18n support which covers the GameSelectionScreen.vue file.

### PR summary
- Uses the `vue-i18n` library we've had half setup since the project started 💀 
  - Defaults to English
  - Gives us built-in reactivity support
  - Keybind (`Alt` + `L`) to switch between English and French (temporary)
    - Allows us to verify visual appearance with translations
- Adds an expected structure on how we build translations

### Expected work on further PRs
- All pages ported
- Language selection (modal? page?)
- Persistence

#### English
image

#### French
image


src/i18n/en/index.ts
```diff
@@ -0,0 +1,64 @@
...
+            }
+        },
+        platforms: {
+            STEAM: "Steam",
+            STEAM_DIRECT: "Steam",
+            EPIC_GAMES_STORE: "Epic Games Store",
+            OCULUS_STORE: "Oculus Store",
+            ORIGIN: "Origin / EA App",
+            XBOX_GAME_PASS: "Xbox Game Pass",
+            OTHER: "Autre",
```
ebkr, line 58: This is fixed in the following PR (#1870)

PR #1870, opened 2025-08-01
I18n/splash

ebkr:
image

image


src/App.vue
```diff
@@ -145,7 +145,7 @@ watchEffect(() => {
 const { locale } = useI18n();
 
 document.addEventListener('keydown', (e: KeyboardEvent) => {
-    if (e.altKey && e.key === 'l') {
+    if (e.altKey && ['l', '¬'].includes(e.key.toLowerCase())) {
```
ebkr, line 151: This is the same shortcut but on macOS. It doesn't like to register as the L key 💢

PR #1875, opened 2025-08-06
I18n/profiles

ebkr:


src/store/modules/TsModsModule.ts
```diff
@@ -214,7 +219,7 @@ export const TsModsModule = {
             commit('startThunderstoreModListUpdate');
 
             try {
-                commit('setThunderstoreModListUpdateStatus', 'Checking for mod list updates from Thunderstore...');
+                commit('setThunderstoreModListUpdateStatus', 'checkingForUpdates');
```
anttimaki, line 219: Is it possible to call the translations here? If yes, what are your thoughts on doing so? It does muddy the waters a bit between separation of concerns, but on the other hand having a magic string here, that then gets parsed into the translation key on the template is a bit hard to follow for the reader. Or alternatively have the entire translation key present here, making it more obvious it's a translation key?

PR #1877, opened 2025-08-07
I18n/manager view

ebkr:


src/components/importing/LocalFileImportModal.vue
```diff
@@ -284,31 +312,31 @@ async function importFile() {
 
     switch (0) {
         case modName.value.trim().length:
-            validationMessage.value = "The mod name must not be empty.";
```
anttimaki, line 287: Similar to comment in an earlier PR in this stack, would it be clearer if the whole translation key was used here?

src/i18n/fr/index.ts
```diff
@@ -2,21 +2,36 @@ import {GameSelectionTranslation} from "./pages/GameSelectionTranslation";
 import {SplashTranslation} from "./pages/SplashTranslation";
 import {PlatformTranslation} from "./platforms/PlatformTranslation";
 import {ProfileSelectionTranslation} from "./pages/ProfileSelectionTranslation";
+import {ManagerTranslation} from './pages/ManagerTranslation';
+import { DateTimeFormat } from 'src/i18n';
```
anttimaki, line 6: Absolute import

PR #1881, opened 2025-08-12
Calculate and display the download progress based on the file size

VilppeRiskidev:
* Calculate the downloaded file size and the progress percentage based on that
* Save the values to the local store in the callback
  * (Importing uses the callback values without storing them)
* Display the the progress (percentage), current progress (file size) and total file size to be downloaded
* Use the values in DownloadMonitor, DownloadProgressModal and ImportProfileModal

### DIFF
--- a/src/components/views/DownloadProgressModal.vue
+++ b/src/components/views/DownloadProgressModal.vue
@@ -2,6 +2,7 @@
 import { Progress } from '../all';
 import { getStore } from '../../providers/generic/store/StoreProvider';
 import { State } from '../../store';
+import FileUtils from '../../utils/FileUtils';
 
 const store = getStore<State>();
 
@@ -29,7 +30,10 @@ function closeModal() {
                         Installing {{$store.getters['download/currentDownload'].modName}}
                     </h3>
 
-                    <p>Downloading: {{$store.getters['download/currentDownload'].downloadProgress}}% complete</p>
+                    <p>
+                        Downloading: {{$store.getters['download/currentDownload'].downloadProgress}}% complete
+                        ({{FileUtils.humanReadableSize($store.getters['download/currentDownload'].downloadedSize)}} / {{FileUtils.humanReadableSize($store.getters['download/currentDownload'].totalDownloadSize)}})
+                    </p>
 
                     <Progress
                         :max='100'
--- a/src/components/views/DownloadProgressModal.vue
+++ b/src/components/views/DownloadProgressModal.vue
@@ -31,8 +31,8 @@ function closeModal() {
                     </h3>
 
                     <p>
-                        Downloading: {{$store.getters['download/currentDownload'].downloadProgress}}% complete
-                        ({{FileUtils.humanReadableSize($store.getters['download/currentDownload'].downloadedSize)}} / {{FileUtils.humanReadableSize($store.getters['download/currentDownload'].totalDownloadSize)}})
+                        Downloading: {{$store.getters['download/currentDownload'].downloadProgress}}% complete of
+                        {{FileUtils.humanReadableSize($store.getters['download/currentDownload'].totalDownloadSize)}}
                     </p>
 
                     <Progress
--- a/src/components/views/DownloadProgressModal.vue
+++ b/src/components/views/DownloadProgressModal.vue
@@ -31,7 +31,7 @@ function closeModal() {
                     </h3>
 
                     <p>
-                        Downloading: {{$store.getters['download/currentDownload'].downloadProgress}}% complete of
+                        Downloading: {{$store.getters['download/currentDownload'].downloadProgress}}% of
                         {{FileUtils.humanReadableSize($store.getters['download/currentDownload'].totalDownloadSize)}}
                     </p>
 
@@ -42,7 +42,7 @@ function closeModal() {
                     />
 
                     <p v-if="$store.getters['download/currentDownload'].installProgress">
-                        Installing: {{$store.getters['download/currentDownload'].installProgress}}% complete
+                        Installing: {{$store.getters['download/currentDownload'].installProgress}}%
                     </p>
                     <p v-else>Installing: waiting for download to finish</p>
 
--- a/src/r2mm/downloading/BetterThunderstoreDownloader.ts
+++ b/src/r2mm/downloading/BetterThunderstoreDownloader.ts
@@ -24,26 +24,24 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
 
         let modInProgressName = combos[0].getMod().getName();
         let finishedModsDownloadedSize = 0;
-        let modInProgressSizeProgress = 0;
 
-        const singleModProgressCallback = (downloadProgress: number, modInProgressSize: number, status: number, err: R2Error | null) => {
-            const clampedProgress = Math.max(0, Math.min(100, downloadProgress));
+        const singleModProgressCallback = (downloadedBytes: number, status: number, err: R2Error | null) => {
+            let modInProgressDownloadedSize;
+
             if (status === StatusEnum.FAILURE) {
                 throw err;
-            }
-
-            if (status === StatusEnum.PENDING) {
-                modInProgressSizeProgress = (clampedProgress / 100) * modInProgressSize;
+            } else if (status === StatusEnum.PENDING) {
+                modInProgressDownloadedSize = downloadedBytes;
             } else if (status === StatusEnum.SUCCESS) {
-                finishedModsDownloadedSize += modInProgressSize;
-                modInProgressSizeProgress = 0;
+                finishedModsDownloadedSize += downloadedBytes;
+                modInProgressDownloadedSize = 0;
             } else {
                 console.error(`Ignore unknown status code "${status}"`);
                 return;
             }
 
             totalProgressCallback(
-                finishedModsDownloadedSize + modInProgressSizeProgress,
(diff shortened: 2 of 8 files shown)


VilppeRiskidev:
<del>Behaviour is buggy when some of the mods are in cache before downloading... I'm fixing it...</del>
Fixed! 👍

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -9,42 +9,55 @@ import FsProvider from '../../providers/generic/file/FsProvider';
...
-import { generateProgressPercentage } from '../../utils/DownloadUtils';
+import ManagerSettings from '../manager/ManagerSettings';
 
 export default class BetterThunderstoreDownloader extends ThunderstoreDownloaderProvider {
 
     public async download(
         combos: ThunderstoreCombo[],
         ignoreCache: boolean,
-        totalProgressCallback: (progress: number, modName: string, status: number, err: R2Error | null) => void
+        settings: ManagerSettings,
```
anttimaki: Don't pass in the settings. `ignoreCache` is already passed in, and it's the only setting that's used.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -9,42 +9,55 @@ import FsProvider from '../../providers/generic/file/FsProvider';
...
+        totalProgressCallback: (progress: number, downloadedSize: number, totalDownloadSize: number, modName: string, status: number, err: R2Error | null) => void
     ): Promise<void> {
         let modInProgressName = combos[0].getMod().getName();
         let downloadCount = 0;
```
anttimaki: Not used anymore.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -9,42 +9,55 @@ import FsProvider from '../../providers/generic/file/FsProvider';
...
+                modInProgressSizeProgress = 0;
                 downloadCount += 1;
             } else {
                 console.error(`Ignore unknown status code "${status}"`);
                 return;
             }
-            totalProgressCallback(Math.round(totalDownloadProgress), modInProgressName, status, err);
+
+            const progressPercentage = totalDownloadSize == 0 ? 100 : Math.round((finishedModsDownloadedSize + modInProgressSizeProgress) / totalDownloadSize * 100);
+            totalProgressCallback(
```
anttimaki, line 43: This comment builds on a comment that's written below, and I haven't tested if this is feasible at all, but would it make sense to:

1. DownloadModule to store the totalDownloadSize when the object is initiated
2. Progress callback to just update downloadedSize, modInProgressName, status, and err
3. DownloadModule to calculate the progressPercentage from downloadedSize & totalDownloadSize (e.g. via DownloadUtils)

IMO it would clean up the download provider implementation nicely and move reusable code elsewhere. Provider's responsivility would be to download and report how much has been downloaded, and how this is used in UI would be responsibility of the components that actually deal with the UI.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -120,4 +138,16 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
         }
     }
 
+    public async getTotalDownloadSizeInBytes(combos: ThunderstoreCombo[], settings: ManagerSettings): Promise<number> {
```
anttimaki: Thoughts on:

1. Calling this from DownloadModule and setting it once during DownloadProgress object creation, rather than passing it through the progress callback and "updating" it on every progress tick?
2. Moving `getTotalDownloadSizeInBytes` and `isVersionAlreadyDownloaded` to DownloadUtils? They don't depend on this class, have no side effects, and I see no reason to duplicate their implementation between the different provider implementations.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -120,4 +138,16 @@ export default class BetterThunderstoreDownloader extends ThunderstoreDownloader
...
+        const filteredCombos = [];
+        if (!settings.getContext().global.ignoreCache) {
```
anttimaki: If cache is ignored, nothing gets added to the list, completely breaking the feature.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
VilppeRiskidev: Makes sense, the profile importing saves the total size separately, as we don't save the download progress for the tracker. 
(I think we could in theory save it and add it to the tracker if we want to, but the importing process handles the installation part a bit differently so this would require some digging).

src/r2mm/downloading/BetterThunderstoreDownloader.ts
```diff
@@ -9,42 +9,52 @@ import FsProvider from '../../providers/generic/file/FsProvider';
...
 
 export default class BetterThunderstoreDownloader extends ThunderstoreDownloaderProvider {
 
     public async download(
         combos: ThunderstoreCombo[],
         ignoreCache: boolean,
-        totalProgressCallback: (progress: number, modName: string, status: number, err: R2Error | null) => void
+        totalProgressCallback: (downloadedSize: number, modName: string, status: number, err: R2Error | null) => void
     ): Promise<void> {
+        if (combos.length === 0) {
```
anttimaki, line 21: Is this for some specific scenario or just a generic precaution? Should the behaviour on "download these zero packages" be to set progress to 100/success and return early? Not necessarily a change request, just thinking out loud.

src/components/views/DownloadProgressModal.vue
```diff
@@ -29,7 +30,10 @@ function closeModal() {
                         Installing {{$store.getters['download/currentDownload'].modName}}
                     </h3>
 
-                    <p>Downloading: {{$store.getters['download/currentDownload'].downloadProgress}}% complete</p>
+                    <p>
+                        Downloading: {{$store.getters['download/currentDownload'].downloadProgress}}% complete
+                        ({{FileUtils.humanReadableSize($store.getters['download/currentDownload'].downloadedSize)}} / {{FileUtils.humanReadableSize($store.getters['download/currentDownload'].totalDownloadSize)}})
```
anttimaki: There's currently a possibility that a download with a size of e.g. 1.2GB shows "1.2GB / 1.2GB" here when the progress percentage is something like 97%. I guess it's caused by `humanReadableSize` rounding something like 1.15GB to 1.2GB (for the progress) and something like 1.249 to 1.2GB (total size) too. So it's not a bug per se, but to me it looks a bit funny.

What's your thoughts on this? We could for example change to using MB precision, which would remove the issue but should mitigate it greatly, as the downloading the last MB is a lot faster than the last 0.1GB. Or we could just skip showing the progress in size and display something like "78% of 1.2GB downloaded"? I'm not opposed to keeping it as it is, or some other ideas either.

VilppeRiskidev: the "78% of 1.2GB downloaded" option seems reasonable, I added a commit with these changes.

src/r2mm/downloading/BetterThunderstoreDownloader.ts
VilppeRiskidev: My thoughts, not any strong opinions:
I think that a 0 combo download would be an erroneous situation. 
As a user I'd probably find it confusing if I tried to initiate a download and there was just no feedback, the error would tell me that something went wrong.

Currently I don't see any situation where the user _should_ be able to trigger a download for 0 mods.

Maybe if:
1. One shares a profile with a mod via a code
2. The mod gets deleted from thunderstore
3. One tries to import the profile via the code
4. ???

Or downloading via a broken link or something...

I've not looked if those get handled before the download (probably).

TL;DR: I don't trust the app nor the user to not _somehow_ end up in a 0 combo download situation😆.

src/utils/DownloadUtils.ts
```diff
@@ -64,20 +64,24 @@ export async function isVersionAlreadyDownloaded(combo: ThunderstoreCombo): Prom
...
+export function generateProgressPercentage(currentProgress: number, targetProgress: number): number {
+    // Target progress might be 0 e.g. if all mods are cached and
+    // nothing needs to be downloaded.
+    if (targetProgress === 0) {
         return 100;
     }
-    if (progressSoFar > targetProgress) {
-        console.warn(`When generating a progress percentage, the progress so far (${progressSoFar}) is greater than the target progress (${targetProgress})`);
+
+    if (currentProgress > targetProgress) {
```
VilppeRiskidev, line 80: This _should_ prevent any cases where the percentage from being over 100%, therefore I omitted the clamping, but yeah, I guess it's better to not depend on it.

src/components/views/DownloadProgressModal.vue
anttimaki: Removed the IMO awkward "x% completed of y MB" structures. IMO the "complete" part is unnecessary here given the context.

src/utils/DownloadUtils.ts
anttimaki: True, I missed that.

anttimaki:
@VilppeRiskidev also added a commit for preventing old download status being shown briefly on the screen before the new one is initiated. I'll go ahead with the merging now but review the changes normally.

anttimaki (approved):

VilppeRiskidev:
Reviewed, seems good 👍

PR #1883, opened 2025-08-14
Cover image optimization

VilppeRiskidev:


VilppeRiskidev:
https://github.com/thunderstore-io/ecosystem-schema/pull/192 Generating the schema from this version of ecosystem-schema repo seems to work with these changes.

VilppeRiskidev:
game_selection folder total size went from 9,68MB to 4,83MB

anttimaki (approved):
- I'm still unsure why bash scripts were chosen for this task, but if it works it works
- `assets/images` seems like a wrong place for the script files. Could the perhaps be moved to e.g. `/scripts` without breaking them?

anttimaki (changes requested):
Uh, the new changes are fine but I probably should've noticed this in the earlier review already: dropping the webp convertor script might be a mistake, as having a script that just renames image files without actually converting them seems like a bad idea. So if we want to keep the scripts around for future use, we should either keep both convert/rename, or combine them into one that converts and renames non-webp files at one go.

anttimaki (approved):
Approved, although this should be merged only after #1823 and at that point the scripts need to be reapplied (moved and added images).

PR #1887, opened 2025-08-21
Improve performance of DownloadUtils.getTotalDownloadSizeInBytes

anttimaki:
Using Promise.all rather than checking for cached versions of the mods in a loop reduces execution time roughly from 50% to 80%. Granted, we were talking about tens of milliseconds to begin with even for large mod packs, but faster is faster.


VilppeRiskidev (approved):
LGTM, didn't find any bugs

PR #1889, opened 2025-08-23
Improve single select configs

Marioalexsan:
Adds two changes for single select style config entries:
- When a value is encountered that is not part of the acceptable values, it's displayed as a temporary option that is selected by default. It disappears when selecting another valid option.
- The list of acceptable values is sorted alphabetically. In the case of enums like KeyCode, this should make it easier to find a specific enum value if you know what you're looking for.

### DIFF
--- a/src/components/config-components/ConfigEntryEditor.vue
+++ b/src/components/config-components/ConfigEntryEditor.vue
@@ -41,6 +41,11 @@
                             <template v-if="entry.displayType === 'single-select' || entry.displayType === 'boolean'">
                                 <div class="settings-input-container">
                                     <select class="select select--full" v-model="entry.value">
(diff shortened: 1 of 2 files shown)


src/components/config-components/ConfigEntryEditor.vue
```diff
@@ -41,6 +41,11 @@
                             <template v-if="entry.displayType === 'single-select' || entry.displayType === 'boolean'">
                                 <div class="settings-input-container">
                                     <select class="select select--full" v-model="entry.value">
+                                        <template v-if="!getSelectOptions(entry).includes(entry.value)">
+                                            <option :value="entry.value" key="existing-value">
```
ebkr: Is there a purpose of having the `key` attribute? It's not part of a `v-for` so I don't think we gain much by having it

PR #1891, opened 2025-08-26
Add games: "Pair A Dice" and "Maltese's Fluffy Onsen"

VilppeRiskidev:


VilppeRiskidev:
"Pair A Dice" seems to crash when launched as modded, but to me it seems to be an issue with the game+(godot)modloader and not really with the mod manager 🤷‍♂️

Jacob1Tm:
I already sent this on the Pair a Dice game request thread on thunderstore's discord but just in case i will send it here too
The reason for the game crashing is modmanager extracting only the modloader itself without it's dependency as seen on a screenshot here
image
After adding this manually it works just fine

anttimaki:
#1897 will resolve the crashing issue.

PR #1892, opened 2025-08-26
Remove superfluous promises

Lord-Kamina:
Removed most if not all unnecessary promises from the FS api. Note, this was tested by running the app on a new profile but my setup isn't giving me any real errors for some reason, so check and/or correct accordingly.

### DIFF
--- a/src/providers/generic/file/NodeFs.ts
+++ b/src/providers/generic/file/NodeFs.ts
@@ -55,25 +55,19 @@ export default class NodeFs extends FsProvider {
     async unlink(path: string): Promise<void> {
             return await NodeFs.lock.acquire(path, async () => {
                 return await fs.promises.unlink(path);
-            }).catch(e) {
-                throw e;
-            }
+            });
     }
 
     async writeFile(path: string, content: string | Buffer): Promise<void> {
             return await NodeFs.lock.acquire(path, () => {
                 fs.writeFileSync(path, content);
-            }).catch(e) {
-                throw e;
-            }
+            });
     }
 
     async rename(path: string, newPath: string): Promise<void> {
         return await NodeFs.lock.acquire([path, newPath], async () => {
             return await fs.promises.rename(path, newPath);
-            }).catch(e) {
-                throw e
-            }
+            });
     }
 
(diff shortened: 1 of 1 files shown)


src/providers/generic/file/NodeFs.ts
```diff
@@ -8,148 +8,115 @@ export default class NodeFs extends FsProvider {
...
-        });
-    }
-
-    async stat(path: string): Promise<StatInterface> {
+    async exists(path: string): boolean {
+        await NodeFs.lock.acquire(path, async () => {
+            const result = await fs.promises.access(path, fs.constants.F_OK).catch(return false);
+            return true;
+        }).catch(return false);
+    }
```
ebkr: 1. You can't have `return false` as a value inside a catch block. A method invocation is required. This would mean it becomes: `.catch(() => false)` in both cases
2. Is the second catch block necessary? Why can we not just:
```suggestion
        await NodeFs.lock.acquire(path, async () => {
            return await fs.promises.access(path, fs.constants.F_OK).catch(() => false);
        });
```

src/providers/generic/file/NodeFs.ts
```diff
@@ -8,148 +8,115 @@ export default class NodeFs extends FsProvider {
...
-            NodeFs.lock.acquire(path, async () => {
-                try {
-                    await fs.promises.mkdir(path, { recursive: true });
-                    resolve();
-                } catch (e) {
-                    reject(e);
-                }
+    async mkdirs(path: string): void {
+            await NodeFs.lock.acquire(path, async () => {
+                await fs.promises.mkdir(path, { recursive: true }).catch (e) {
```
ebkr: `.catch( ... throw)` does the exact same thing as not having a catch block in the first place. You don't need the catch block.

src/providers/generic/file/NodeFs.ts
Lord-Kamina: Hadn't seen these comments. I'll fix them when I've got a minute.

Although, maybe in the long run it'd be better to rethrow/convert to your custom error class?

PR #1896, opened 2025-08-28
Added BepisLoader support

ebkr:
# BepisLoader

## Install structure

With the following changes, the profile folder looks like:
```
.
├── _state
│   └── installation_state.yml
├── BepInEx
│   ├── core
│   │   ├── 0Harmony.dll
│   │   ├── 0Harmony.pdb
│   │   ├── 0Harmony.xml
│   │   ├── BepInEx.Core.dll
│   │   ├── BepInEx.Core.xml
│   │   ├── BepInEx.NET.Common.dll
│   │   ├── BepInEx.NET.Common.xml
│   │   ├── BepInEx.NET.CoreCLR.deps.json
│   │   ├── BepInEx.NET.CoreCLR.dll
│   │   ├── BepInEx.NET.CoreCLR.xml
│   │   ├── BepInEx.Preloader.Core.dll
│   │   ├── BepInEx.Preloader.Core.xml
│   │   ├── Mono.Cecil.dll
│   │   ├── Mono.Cecil.Mdb.dll
│   │   ├── Mono.Cecil.Pdb.dll
│   │   ├── Mono.Cecil.Rocks.dll
│   │   ├── MonoMod.Backports.dll
│   │   ├── MonoMod.Core.dll
│   │   ├── MonoMod.Iced.dll
│   │   ├── MonoMod.ILHelpers.dll
│   │   ├── MonoMod.RuntimeDetour.dll
│   │   ├── MonoMod.Utils.dll
│   │   └── SemanticVersioning.dll
│   ├── patchers
│   └── plugins
│       └── ResoniteModding-BepInExResoniteShim
│           ├── BepInExResoniteShim
│           │   ├── BepInExResoniteShim.dll
│           │   └── BepInExResoniteShim.pdb
│           ├── icon.png
│           ├── LICENSE
│           ├── manifest.json
│           └── README.md
├── BepisLoader.deps.json
├── BepisLoader.dll
├── BepisLoader.pdb
├── BepisLoader.runtimeconfig.json
├── hookfxr.ini
├── hostfxr.dll
├── hostfxr.pdb
└── mods.yml
``` 
once BepisLoader (and the dependency) are installed.


src/installers/BepInExInstaller.ts
```diff
@@ -19,7 +19,7 @@ export class BepInExInstaller implements PackageInstaller {
 
         const mapping = MODLOADER_PACKAGES.find((entry) =>
             entry.packageName.toLowerCase() == mod.getName().toLowerCase() &&
-            entry.loaderType == PackageLoader.BEPINEX,
+            (entry.loaderType == PackageLoader.BEPINEX || entry.loaderType == PackageLoader.BEPISLOADER),
```
anttimaki: I wonder if this line is needed at all? Isn't the package name by itself already a unique identifier, so what does the extra check accomplish? I also *think* `GenericProfileInstaller` already does a similar check and this `install` is only ever called if the package is associated with `BepInExInstaller` in the `registry.ts`. So unless I'm mistaken we could omit this line, keeping the BepInExInstaller unaware of BepisLoader, while allowing BepisLoader to reuse the code?

src/r2mm/launching/instructions/instructions/loader/BepisLoaderGameInstructions.ts
```diff
@@ -0,0 +1,17 @@
...
+import Profile from '../../../../../model/Profile';
+import { DynamicGameInstruction } from '../../DynamicGameInstruction';
+import path from "path";
+
+export default class BepisLoaderGameInstructions extends GameInstructionGenerator {
+
+    public async generate(game: Game, profile: Profile): Promise<GameInstruction> {
+        return {
+            moddedParameters: `--hookfxr-enable --bepinex-target ${path.join(DynamicGameInstruction.PROFILE_DIRECTORY, 'BepInEx')}`,
+            vanillaParameters: `--hookfxr-disable` // Intentionally use the wrong target
```
anttimaki: The comment gives me "more questions than answers" vibes if I'm reading this with zero context knowledge.

src/installers/BepInExInstaller.ts
ebkr: I don't believe there's a reason for it, and if there was it was likely legacy. There's no apparent repercussions from removing the line as nothing else should be using it. All of our definitions should be accurate and so the root folder shouldn't be an issue anymore. Likely a remnant from sharing the installing with MelonLoader?

src/installers/InstallRuleInstaller.ts
```diff
@@ -113,6 +113,30 @@ async function installSubDirNoFlatten(profile: ImmutableProfile, rule: ManagedRu
...
+            }
+            if (fileParts[i] === ruleParts[i]) {
+                numberOfMatches++;
+            }
+        }
+        return {
+            rule: value,
+            count: numberOfMatches
+        };
+    }).reduce((previousValue, currentValue) => {
```
anttimaki, line 143: Nitpick: is there a sane fallback value of type `{rule, count}` that could be given as a starting value for this `reduce`? I think reduce would currently throw if the `matchingRules` arg is an empty array. It's currently controlled at the call site that this function is not called with empty arrays, but there's no guarantees that someone won't inadvertently change it in the future.

src/installers/InstallRuleInstaller.ts
```diff
@@ -149,9 +173,18 @@ async function buildInstallForRuleSubtype(
...
+
+        if (matchingRules.length > 1) {
+            // Back-resolve path to find the best fitting rule.
+            const fileParts = file.getTarget().split(path.sep).reverse();
+            matchingRule = getBestFitRule(matchingRules, fileParts).rule;
+        } else if (matchingRules.length === 1) {
+            matchingRule = matchingRules[0];
+        }
+
+        if (matchingRule === undefined) {}
```
anttimaki: Nitpick: some leftover.

PR #1905, opened 2025-09-07
Add file size display to profile selection

VilppeRiskidev:


VilppeRiskidev:
OK, I found an issue with this implementation:
1. Import profile to an existing profile
2. The profile's file size doesn't update 

I'm looking into how to best trigger the update...

VilppeRiskidev:
Closing this as priorities changed and this is now deep in the backlog

PR #1911, opened 2025-09-10
Fix bepisloader launch arguments

hazre:
Fixes BepisLoader launch issues in Resonite:

* Properly quotes the `--bepinex-target` path so launches no longer fail when spaces are present.
* Reverts Doorstop V4 parameter changes (`--doorstop-enabled` / `--doorstop-target-assembly`) that broke compatibility, restoring the correct arguments (`--doorstop-enable` / `--doorstop-target`).


hazre:
Actually beyond the --bepinex-target path fix, there is something more underling that causes the doorstep args to be ignored completely and the locally copied over BepinExRenderer is ran regardless.

PR #1912, opened 2025-09-12
Several fixes for Resonite

ebkr:
- Quoted bepinex-target arg

- No error thrown when Renderer package is not installed

- Renderer package can now be uninstalled

- Renderer folder is not copied to game directory

### DIFF
--- a/src/installers/BepisLoaderInstaller.ts
+++ b/src/installers/BepisLoaderInstaller.ts
@@ -1,9 +1,8 @@
 import { InstallArgs, PackageInstaller } from "./PackageInstaller";
 import { BepInExInstaller } from './BepInExInstaller';
-import ProfileModList from 'src/r2mm/mods/ProfileModList';
-import R2Error from 'src/model/errors/R2Error';
-import FileUtils from 'src/utils/FileUtils';
-import FsProvider from 'src/providers/generic/file/FsProvider';
+import ProfileModList from '../r2mm/mods/ProfileModList';
+import R2Error from '../model/errors/R2Error';
+import FileUtils from '../utils/FileUtils';
 
 export class BepisLoaderInstaller implements PackageInstaller {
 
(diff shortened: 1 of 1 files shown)


src/installers/BepisLoaderInstaller.ts
```diff
@@ -1,5 +1,9 @@
 import { InstallArgs, PackageInstaller } from "./PackageInstaller";
 import { BepInExInstaller } from './BepInExInstaller';
+import ProfileModList from 'src/r2mm/mods/ProfileModList';
```
anttimaki: Absolute imports.

src/installers/BepisLoaderInstaller.ts
```diff
@@ -9,9 +13,22 @@ export class BepisLoaderInstaller implements PackageInstaller {
...
+        Resonite deals with two mod loaders. As it's currently the only game using BepisLoader, for now we'll hard-code
+        mod loader packages and we'll remove the appropriate folders. We'll only remove the renderer for now.
          */
+        const { profile } = args;
+        const modList = await ProfileModList.getModList(profile);
+        if (modList instanceof R2Error) {
+            throw modList;
+        }
+
+        const hasRendererModLoaderInstalled = modList.findIndex(value => value.getName() === 'ResoniteModding-BepInExRenderer') >= 0;
```
anttimaki: Nitpick: could use `Array.some()`

src/installers/BepisLoaderInstaller.ts
```diff
@@ -9,9 +13,22 @@ export class BepisLoaderInstaller implements PackageInstaller {
...
+        if (!hasRendererModLoaderInstalled) {
+            const rendererPath = profile.joinToProfilePath('Renderer');
+            await FileUtils.ensureDirectory(rendererPath);
+            await FileUtils.emptyDirectory(rendererPath);
+            await FsProvider.instance.rmdir(rendererPath);
```
anttimaki: Nitpick: I've at some point added `FileUtils.recursiveRemoveDirectoryIfExists` which basically does the same thing and may or may not be more readable depending on how one thinks about these things.

src/installers/BepisLoaderInstaller.ts
```diff
@@ -9,9 +13,22 @@ export class BepisLoaderInstaller implements PackageInstaller {
...
+        }
         return;
```
anttimaki: Nitpick: unnecessary.

anttimaki (changes requested):
The imports should be fixed.

PR #1939, opened 2025-09-22
path.sep can now be accessed via IPC

ebkr:


ebkr:
Ah oops, it was to debug to identify that it was actually resulting in being undefined.

PR #1940, opened 2025-09-22
Remove some unneeded promises

Lord-Kamina:
I've remade this PR by  cherry-picking and squashing the commits from the original onto the latest develop branch.
Tests on my fork look fine. Locally, tests succeed except for the install logic ones (I suspect there's an issue with the paths but not sure what), and running it locally with a mock profile seems to work fine.

Of note: There are still several `new Promise` and `Promise.resolve` and `.then` bits in the code. It's very likely that most of those can be changed to use await, but I'd rather leave them to someone else for now.

### DIFF
--- a/src/providers/node/fs/NodeFsImplementation.ts
+++ b/src/providers/node/fs/NodeFsImplementation.ts
@@ -3,12 +3,10 @@ import Lock from 'async-lock';
 
 const lock = new Lock();
 
-function acquireLockAndDo(path: string, action: (resolve: any, reject: any) => Promise<any>): Promise<any> {
-    return new Promise((resolve, reject) => {
-        lock.acquire(path, async () => {
-            await action(resolve, reject);
-        })
-    })
+async function acquireLockAndDo(path: string, action: Promise<any>) {
+    return await lock.acquire(path, async () => {
+        return await action;
+    });
 }
 
 function resolveStat(statLike: any) {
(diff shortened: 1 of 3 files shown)


src/providers/node/fs/NodeFsImplementation.ts
ebkr, line 1: These aren't working as intended

```ts
return await acquireLockAndDo(path, window.node.fs.writeFile(...args));
```

effectively becomes:

```ts
const result = window.node.fs.writeFile(...args)
return (lock, result);
```

we actually want it to work like:

```ts
return lock(() => window.node.fs.writeFile(...args))
```

_These are not necessarily syntactically accurate, just pseudo_

src/utils/LaunchUtils.ts
```diff
@@ -91,10 +91,12 @@ export async function getDeterminedLaunchType(game: Game, launchType: LaunchType
...
-    return Promise.resolve()
-        .then(async () => await (GameDirectoryResolverProvider.instance as LinuxGameDirectoryResolver).getLaunchArgs(game))
-        .then(launchArgs => typeof launchArgs === 'string' && launchArgs.startsWith(path.join(PathResolver.MOD_ROOT, 'linux_wrapper.sh')))
-        .catch(() => false);
+    try {
+        let launchArgs = await (GameDirectoryResolverProvider.instance as LinuxGameDirectoryResolver).getLaunchArgs(game);
+    } catch {
+        let launchArgs = false;
+    }
+    return (typeof launchArgs === 'string' && launchArgs.startsWith(path.join(PathResolver.MOD_ROOT, 'linux_wrapper.sh')));
```
ebkr, line 99: This needs updating, but I think this will be sufficient for the conflict?

```suggestion
export async function areWrapperArgumentsProvided(game: Game): Promise<boolean> {
    try {
        const isFlatpak = await isManagerRunningOnFlatpak();
        const flatpakWrapper = path.join(PathResolver.MOD_ROOT, 'web_start_wrapper.sh');
        const linuxWrapper = path.join(PathResolver.MOD_ROOT, 'linux_wrapper.sh');
        const appropriateWrapper = isFlatpak ? flatpakWrapper : linuxWrapper;
        const launchArgs = await getProvidedWrapperArguments(game);
        return launchArgs.includes(appropriateWrapper);
    } catch {
        return false;
    }
}
```

ebkr (changes requested):
Apologies for the delay in getting to this. Still a couple of requested changes

ebkr:
Apologies for keeping this open for so long, it slipped my mind completely.

This area of code is no longer used anymore and instead NodeFsImplementation is used, which does all of the communication over IPC.

PR #1961, opened 2025-10-07
Add other executable names for Silksong

capitalistspz:
Add Silksong mac/linux executable names


ebkr:
You’ll want to do this here instead: https://github.com/thunderstore-io/ecosystem-schema

capitalistspz:
> You’ll want to do this here instead: https://github.com/thunderstore-io/ecosystem-schema

Alright

Issue #1964, opened 2025-10-12
[BUG] - Files in extracted mod zips do not preserve their permissions


capitalistspz:
**Describe the bug**
On Linux, downloading a mod from the profile screen, or importing a zip file will extract the file to its profile folder without the permissions it originally had.

1. Go to a profile for a game (in this case: Hollow Knight Silksong) in r2modman
2. Download a mod which contains files with permissions other than `rw-rw-rw-`. (in this case: [BepInEx Pack Silksong](https://thunderstore.io/c/hollow-knight-silksong/p/BepInEx/BepInExPack_Silksong/)
3. Open profile directory, and view file permissions for the files (in this case all files are `rw-rw-rw-` when their zipped versions were `rwxr-xr-x`)

**Expected behavior**
For file permissions to preserved, or at least preserve the executable bit on files.

PR #1965, opened 2025-10-15
Preserve zip contents' perms on extraction

capitalistspz:
Fixes https://github.com/ebkr/r2modmanPlus/issues/1964


ebkr:
The issue you'd outlined only shows permission on extract. `+x` is marked at runtime when you click `Start modded`.

I don't think this is needed. It's also less ideal from a security standpoint.

PR #1975, opened 2025-10-22
addChineseAlias

lostone-zc:
> Added Chinese aliases for some games for easy searching.

Same as #1342


ebkr:
This is automatically synced from https://github.com/thunderstore-io/ecosystem-schema

This file should not be manually edited

PR #1983, opened 2025-11-03
Make NodePathProvider::resolve variadic

### LINKED ISSUE #1963: [BUG] - SteamOS autodetect
**Describe the bug**
path to steam folder didnt autodetect on steamos

**To Reproduce**
1. Install fresh steamOS on the device of choice (likely steamdeck)
1.1 Install silksong or any other linux native game
2. Make steamOS no longer readonly
3. Install and launch r2modman
4. Choose game from 1.1
5. Observe the bug

**Expected behavior**
Automatic detection of steam install folder

**Additional context**
Game folder also didnt detect, and it wouldnt accept the [executable](https://discord.com/channels/727304496522461185/727304496522461189/1426779862571679837) until i restarted r2modman with the steam folder manually set

RossBrunton:
The associated node function is variadic, and this is assumed by the linux path management.

I think this partially fixes #1963, although on my local system there's another issue preventing Linux from working.

### DIFF
--- a/src/providers/node/path/path.ts
+++ b/src/providers/node/path/path.ts
@@ -30,7 +30,7 @@ const nodePath: NodePathProvider = {
     extname: path => getImplementation().extname(path),
     basename: path => getImplementation().basename(path),
     relative: (pathOne, pathTwo) => getImplementation().relative(pathOne, pathTwo),
-    resolve: paths => getImplementation().resolve(paths),
+    resolve: (...args) => getImplementation().resolve(...args),
     get sep() { return getImplementation().sep }
 };
 


ebkr:
Could you elaborate on how exactly this would fix the linked issue?

I don't believe this is needed, as it works on an application restart as mentioned which suggests it's an odd runtime env bug, rather than a resolve bug.

RossBrunton:
`src/r2mm/manager/linux/GameDirectoryResolver.ts` contains this:

```ts
            const dirs = [
                path.resolve(os.homedir(), '.local', 'share', 'Steam'),
                path.resolve(os.homedir(), '.steam', 'steam'),
                path.resolve(os.homedir(), '.steam', 'root'),
                path.resolve(os.homedir(), '.steam'),
                path.resolve(os.homedir(), '.var', 'app', 'com.valvesoftware.Steam', '.local', 'share', 'Steam'),
                path.resolve(os.homedir(), '.var', 'app', 'com.valvesoftware.Steam', '.steam', 'steam'),
                path.resolve(os.homedir(), '.var', 'app', 'com.valvesoftware.Steam', '.steam', 'root'),
                path.resolve(os.homedir(), '.var', 'app', 'com.valvesoftware.Steam', '.steam')
            ];
```

As the `resolve` helper function is not variadic, only the first argument is used. This means it effectively checks `/home/whatever` eight times for the steam library, which fails.

Once it is set manually (and apparently after a restart?) this autodiscovery is skipped in favour of the value in the preferences.

PR #2025, opened 2025-12-09
Added Chinese (zh-CN) i18n translations.

TianMengLucky:


ebkr:
Ah this got accidentally closed. The new target is: https://github.com/ebkr/r2modmanPlus/pull/2030

I still need to do a rebase to bring it up to the latest develop commit. This new branch does however have **most** of the translations now though. There are a couple of extras that have been added to develop since that I'll need to verify.

It should be a simple rebase onto `feature/i18n` for now as it's the same base as the branch this was previously based off, and all changes have been merged rather than rebased. Let me know once you've done that and when I get around to it I'll do a further rebase of my branch

lostone-zc:
@TianMengLucky are you still work on it? i18n branch is merged into main branch now, so r2modman now support multiple languages in released versions, so I think you can just make a pr to main branch. Also, I correct the Gale translation, you can refer that translation.

PR #2026, opened 2025-12-09
Add support for Scrap Mechanic and Rivet

BenMcAvoy:
https://store.steampowered.com/app/387990/Scrap_Mechanic/
https://github.com/ReDoIngMods/Rivet/

Following discussion in the Scrap Mechanic game request channel

### DIFF
--- a/src/installers/RivetInstaller.ts
+++ b/src/installers/RivetInstaller.ts
@@ -1,5 +1,4 @@
 import path from "../providers/node/path/path";
-import fs from "../providers/node/fs/fs";
 
 import {
     disableModByRenamingFiles,
@@ -34,15 +33,38 @@ export class RivetInstaller implements PackageInstaller {
     async uninstall(args: InstallArgs): Promise<void> {
         const { packagePath, profile } = args;
 
-        try {
-            // Delete `version.dll` file
-            const versionDllPath = profile.joinToProfilePath("version.dll");
-            fs.unlink(versionDllPath);
+        for (const file of RivetInstaller.TRACKED) {
+            const profilePath = profile.joinToProfilePath(file);
 
-            const rivetDirPath = profile.joinToProfilePath("Rivet");
-            await FileUtils.recursiveRemoveDirectoryIfExists(rivetDirPath);
-        } catch (e) {
-            throw FileWriteError.fromThrownValue(e, "Failed to uninstall Rivet loader");
+            if (!(await FsProvider.instance.exists(profilePath))) {
+                continue;
+            }
+
+            if (file === "Rivet") {
+                // Remove everything apart from Mods
+                const rivetContents = await FsProvider.instance.readdir(profilePath);
+                for (const item of rivetContents) {
+                    if (item.toLowerCase() === "mods") {
+                        continue;
+                    }
+
+                    const itemPath = path.join(profilePath, item);
+                    if ((await FsProvider.instance.stat(itemPath)).isDirectory()) {
+                        await FileUtils.recursiveRemoveDirectoryIfExists(itemPath);
+                    } else {
+                        await FsProvider.instance.unlink(itemPath);
+                    }
+                }
+
+                // Keep going to next tracked file
+                continue;
+            }
+
+            if ((await FsProvider.instance.stat(profilePath)).isDirectory()) {
+                await FileUtils.recursiveRemoveDirectoryIfExists(profilePath);
+            } else {
+                await FsProvider.instance.unlink(profilePath);
+            }
         }
     }
 }
@@ -50,10 +72,6 @@ export class RivetInstaller implements PackageInstaller {
 export class RivetPluginInstaller implements PackageInstaller {
     private getModsPath(args: InstallArgs): string {
         return args.profile.joinToProfilePath("Rivet", "Mods", args.mod.getName());
-    }
-
-    private getUserDataPath(args: InstallArgs): string {
-        return args.profile.joinToProfilePath("Rivet", "UserData", args.mod.getName());
     }
 
     private throwActionError(e: unknown, action: string): void {
@@ -71,14 +89,8 @@ export class RivetPluginInstaller implements PackageInstaller {
             for (const item of files) {
                 const sourceFull = path.join(args.packagePath, item);
 
-                if (item === "UserData") {
-                    const userDataPath = this.getUserDataPath(args);
-                    await FileUtils.ensureDirectory(userDataPath);
-                    await FileUtils.copyFileOrFolder(sourceFull, userDataPath);
-                } else {
-                    const targetPath = path.join(modsPath, item);
-                    await FileUtils.copyFileOrFolder(sourceFull, targetPath);
-                }
+                const targetPath = path.join(modsPath, item);
+                await FileUtils.copyFileOrFolder(sourceFull, targetPath);
             }
         } catch (e) {
             this.throwActionError(e, "install");
@@ -88,7 +100,6 @@ export class RivetPluginInstaller implements PackageInstaller {
     async uninstall(args: InstallArgs): Promise<void> {
         try {
             await FileUtils.recursiveRemoveDirectoryIfExists(this.getModsPath(args));
-            await FileUtils.recursiveRemoveDirectoryIfExists(this.getUserDataPath(args));
         } catch (e) {
             this.throwActionError(e, "uninstall");
         }


src/assets/data/ecosystem.json
```diff
@@ -13652,6 +13652,48 @@
...
+          "internalFolderName": "ScrapMechanic",
+          "dataFolderName": "Scrap Mechanic_Data",
+          "distributions": [
+            {
+              "platform": "steam",
+              "identifier": "387990"
+            }
+          ],
+          "settingsIdentifier": "scrap-mechanic",
+          "packageIndex": "https://thunderstore.io/c/scrap-mechanic/api/v1/package-listing-index/",
```
anttimaki: Change this to `https://thunderstore.dev/c/test/api/v1/package-listing-index/` while developing and testing. Then you can upload Rivet package and a test mod at https://test.thunderstore.dev/ . It can take a couple of hours for the mod manager to see the newly uploaded mods, but once it does, you test if installation, launching etc. works as expected.

src/installers/registry.ts
```diff
@@ -63,7 +66,8 @@ function isPluginInstaller(loader: PackageLoader): loader is PluginInstallers {
         loader === PackageLoader.GODOTML ||
         loader === PackageLoader.MELONLOADER ||
         loader === PackageLoader.NORTHSTAR ||
-        loader === PackageLoader.UMM
+        loader === PackageLoader.UMM ||
+        loader === PackageLoader.RIVET
```
anttimaki: Remove this as Rivet is a PluginInstaller (note the `!` on L63).

src/installers/RivetInstaller.ts
```diff
@@ -0,0 +1,94 @@
...
+    enableModByRenamingFiles,
+    InstallArgs,
+    PackageInstaller,
+} from "./PackageInstaller";
+
+import FileWriteError from "../model/errors/FileWriteError";
+import FsProvider from "../providers/generic/file/FsProvider";
+import FileUtils from "../utils/FileUtils";
+
+export class RivetInstaller implements PackageInstaller {
```
anttimaki, line 14: This needs to implement uninstall as well.

src/installers/RivetInstaller.ts
```diff
@@ -0,0 +1,94 @@
...
+        const { packagePath, profile } = args;
+
+        try {
+            for (const file of RivetInstaller.TRACKED) {
+                const cachePath = path.join(packagePath, file);
+                const profilePath = profile.joinToProfilePath(file);
+                await FileUtils.copyFileOrFolder(cachePath, profilePath);
+            }
+        } catch (e) {
+            throw FileWriteError.fromThrownValue(e, "Failed to install Rivet mod");
```
anttimaki: Perhaps use "mod loader" instead of "mod" here to make it clear it's the Rivet itself that caused trouble.

src/installers/RivetInstaller.ts
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+}
+
+export class RivetModInstaller implements PackageInstaller {
```
anttimaki: Please call this RivetPluginInstaller for consistency with the rest of the code base.

src/assets/data/ecosystem.json
```diff
@@ -32414,6 +32456,11 @@
       "packageId": "UMM-UMM",
       "rootFolder": "UMM",
       "loader": "umm"
+    },
+    {
+      "packageId": "ReDoIngMods-Rivet",
```
anttimaki: It's usually a good idea to have a separate team for the mod loader. That way the maintenance burden can be shared with others, without giving them to access to the rest of your mods. E.g. for some mod loaders we have 753 as team member just in case the original author doesn't maintain the package anymore, since all the mods that depend on the mod loader need to update to depend on the new package if the team name changes.

src/installers/RivetInstaller.ts
```diff
@@ -0,0 +1,112 @@
+import path from "../providers/node/path/path";
+import fs from "../providers/node/fs/fs";
```
anttimaki: Don't import this directly, use FsProvider instead.

src/installers/RivetInstaller.ts
```diff
@@ -0,0 +1,112 @@
...
+    InstallArgs,
+    PackageInstaller,
+} from "./PackageInstaller";
+
+import FileWriteError from "../model/errors/FileWriteError";
+import FsProvider from "../providers/generic/file/FsProvider";
+import FileUtils from "../utils/FileUtils";
+
+export class RivetInstaller implements PackageInstaller {
+    private static readonly TRACKED = ["version.dll", "Rivet.ini", "Rivet"];
```
anttimaki, line 15: Just pointing out that the two older versions of the mod loader don't contain `Rivet.ini`, and installing them currently fails. It's not an issue since you probably aren't planning on uploading the old versions to production environment anyway, but if you think the contents of the mod loader package itself are not yet stable, it might make sense to check that the file exists in cachePath before trying to copy it.

src/installers/RivetInstaller.ts
```diff
@@ -0,0 +1,112 @@
...
+            }
+        } catch (e) {
+            throw FileWriteError.fromThrownValue(e, "Failed to install Rivet loader");
+        }
+    }
+
+    async uninstall(args: InstallArgs): Promise<void> {
+        const { packagePath, profile } = args;
+
+        try {
```
anttimaki: Any reason RivetInstaller.TRACKED isn't used here too? Currently this leaves Rivet.ini behind, meaning the uninstall isn't perfectly clean.

src/installers/RivetInstaller.ts
```diff
@@ -0,0 +1,112 @@
...
+            // Delete `version.dll` file
+            const versionDllPath = profile.joinToProfilePath("version.dll");
+            fs.unlink(versionDllPath);
```
anttimaki: 1. Use `await FsProvider.instance.unlink`
2. Check that the file exists (`await FsProvider.instance.exists`) before deleting it. Mod manager uninstalls the old version of mod loader before installing the new one, so `uninstall` might be called without `install` ever being called

src/installers/RivetInstaller.ts
```diff
@@ -0,0 +1,112 @@
...
+            await FileUtils.recursiveRemoveDirectoryIfExists(rivetDirPath);
+        } catch (e) {
+            throw FileWriteError.fromThrownValue(e, "Failed to uninstall Rivet loader");
+        }
+    }
+}
+
+export class RivetPluginInstaller implements PackageInstaller {
+    private getModsPath(args: InstallArgs): string {
+        return args.profile.joinToProfilePath("Rivet", "Mods", args.mod.getName());
```
anttimaki, line 74: Mods are placed under Rivet, which gets deleted when the mod loader is uninstalled. This causes all the mod files to be deleted, but mod manager still thinking the mods themselves are present, which can confuse the user. Perhaps the Mods folder should reside outside Rivet, or the mod loader uninstallation should skip removing the Mods subfolder.

src/installers/RivetInstaller.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    private getUserDataPath(args: InstallArgs): string {
+        return args.profile.joinToProfilePath("Rivet", "UserData", args.mod.getName());
```
anttimaki: If there's no plan on the mods containing UserData, this whole thing can be dropped as unnecessary. I know it's on my example so that's on me. Some game/mod loader used it to have a location that doesn't get deleted if the mod itself is uninstalled (e.g. when a mod is updated). But the current implementation deletes the UserData folder on uninstall, so it's a bit pointless.

src/r2mm/manager/ModLinker.ts
```diff
@@ -88,6 +88,11 @@ export default class ModLinker {
                 return path.join(installDirectory, game.dataFolderName, "Binaries", "Win64");
             }
             return null;
+        } else if (game.packageLoader == PackageLoader.RIVET) {
+            if (lowercased === "version.dll") {
```
anttimaki, line 92: Once this file has been copied to the game folder, the console window pops up even when the game is launched vanilla, e.g. directly through Steam. Could you change it so the console is shown only when launched modded (i.e. when the launch parameters are correct)?

src/assets/data/ecosystem.json
BenMcAvoy: ReDoIng mods is a team with DLL modders in it and isn’t just me
Should it still be in a separate team despite this?

src/installers/RivetInstaller.ts
BenMcAvoy: Resolved by previous changes to keep the `Rivet/Mods` directory

src/r2mm/manager/ModLinker.ts
BenMcAvoy: This can definitely be done and has been added to the todo issues 😄

src/assets/data/ecosystem.json
anttimaki: Well let's put it this way: if you're fine with adding someone from Thunderstore staff to the team, giving them the access to manage the mods, then it's fine.

PR #2028, opened 2025-12-18
Deck

ebkr:
# Steam Deck / Flatpak Support

_This PR adds Flatpak building potential to r2modman which allows native installing to the deck (and running in Game Mode)._

## Status

As of now, this branch can be built to a Flatpak image and it will run.

It has currently broken launching of every other platform. Works great if you're on Steam Deck - not if you're a regular user.

## Differences

Flatpak has an absolute ton of restrictions. To work around this for Flatpak-based installs we now do the following:

- Invoke Steam via `steam://run/<appid>`.
- Require a separate launch wrapper script (part of this PR).
  - Arguments are consumed as part of a text file and not passed via CLI.
  - This is again, a limitation of Flatpak and a hacky workaround.

## Extra changes

I've taken out the automated file modifications to enable WINEDLLOVERRIDES. Instead we are now making the user supply the override as an environment variable as it's more reliable. This is done as part of the launch arguments copyable string.

Users will be prompted if their launch args don't match. For existing launches they'll already have the file modified so it won't affect older users regardless of the re-prompt.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,6 +17,9 @@ jobs:
           - kind: windows
             os: windows-latest
             platform: win
+          - kind: flatpak
+            os: ubuntu-latest
+            platform: flatpak
     steps:
       - uses: actions/checkout@v4
 
@@ -29,7 +32,7 @@ jobs:
       - name: Install Yarn dependencies
         run: yarn install --immutable
 
-      - if: matrix.platform == 'linux'
+      - if: matrix.platform == 'linux' || matrix.platform == 'flatpak'
         name: Install bsdtar # Required by electron-builder when targeting pacman.
         run: |
           sudo apt-get update -y
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "publish": "quasar build --mode electron --publish always",
     "publish-win": "quasar build --mode electron -T win32 --publish always",
     "publish-linux": "quasar build --mode electron -T linux --publish always",
+    "publish-flatpak": "BUILD_FLATPAK=true quasar build --mode electron -T linux --publish always",
     "test": "vitest run ./test/vitest"
   },
   "dependencies": {


src/r2mm/launching/instructions/instructions/loader/GDWeaveGameInstructions.ts
```diff
@@ -7,8 +7,8 @@ import { DynamicGameInstruction } from '../../DynamicGameInstruction';
 export default class GDWeaveGameInstructions extends GameInstructionGenerator {
     public async generate(game: Game, profile: Profile): Promise<GameInstruction> {
         return {
-            moddedParameters: `--gdweave-folder-override="${DynamicGameInstruction.GDWEAVE_FOLDER}"`,
-            vanillaParameters: "--gdweave-disable"
+            moddedParameterList: [`--gdweave-folder-override=${DynamicGameInstruction.GDWEAVE_FOLDER}`],
```
anttimaki, line 10: The quotes were [intentionally added](https://github.com/ebkr/r2modmanPlus/commit/d17c3630759e2a0630b011f8e2f3210d6dfbe098) here so ensure removing them doesn't break things.

src/r2mm/launching/instructions/GameInstructionParser.ts
```diff
@@ -37,6 +37,19 @@ export default class GameInstructionParser {
         return resolvedString;
     }
 
+    public static async parseList(args: string[], game: Game, profile: Profile): Promise<string[] | R2Error> {
+        args.forEach(arg => this.parse(arg, game, profile));
```
anttimaki: Is this some leftover line?

src/r2mm/launching/instructions/instructions/loader/GDWeaveGameInstructions.ts
ebkr: The behaviour should be _almost_ identical

We go from:
```diff
-: --gdweave-folder-override="some path"
+: "--gdweave-folder-override=some path"
```

Both are almost entirely similar, however this new way prevents all arg parsers from reading `"` as part of one of the value.

Compared to the original reason, we'd end up with:
```diff
-: --gdweave-folder-override=some path
+: "--gdweave-folder-override=some path"
```
and this would have been done so that the command was treated as a single entry.
I haven't tested with GDWeave, but in general, a quote surrounded argument is handled correctly.

ethangreen-dev:
Can confirm that this works on my Steam deck

PR #2032, opened 2026-01-02
Local Mod List Improvements #2

ebkr:
# Local Mod List Improvements

### Primary change 

Dropped mods.yml size down from 16139KB to 105KB. This is on a profile with 177 mods.

This was done using the `reducer` when stringify-ing the content of mods.yml and stripping out the `icon` key/value.

### Secondary effect

This has an added bonus of **drastically** improving the load times to get to the profiles screen.

## Fixing dragging ghost-effect

Prior to this change, when you drag a mod, due to the asynchronous behaviour of the manager, the mod list is treated as being reset before the state has changed. This gave the appearance of the mod jumping back to the original drag position, and then momentarily after being updated to the real position.

There's now a new ref variable which holds the local state and is updated by both the draggable list and the VueX state.

## Suspense

I've also introduced a new SkeletonLocalModCard. This is used as a loading placeholder whilst the actual state and dependency calculation is done in the background for mods.

It means that the installed list can be seen far quicker, albeit non-interactive until loaded. Regardless it's still a nice improvement that's been needed for a while.

## Summary

Overall this PR provides massive speed boosts in larger profiles for the following actions:
- Loading
- Installing
- Uninstalling
- Disabling
- Enabling
- Re-ordering mods

In addition to that, once all profiles have been updated, the time to load into profile selection screen has been significantly improved. Assuming all profiles have been updated to the new format, it can load to the selection screen near instantly.

The suspense implementation provides much faster feedback to see what's available in the current profile.

### DIFF
--- a/src/components/views/LocalModList.vue
+++ b/src/components/views/LocalModList.vue
@@ -11,19 +11,13 @@
 
         <div class="mod-list-content">
             <div class="draggable-content">
-                <draggable v-model='draggableList'
-                           group="local-mods"
-                           handle=".handle"
-                           @start="drag=store.getters['profile/canSortMods']"
-                           @end="drag=false"
-                           :force-fallback="true"
-                           :scroll-sensitivity="100"
-                           item-key="id">
-                    <template #item="{element}">
-                        <local-mod-card
-                            :mod="element" />
+                <Suspense>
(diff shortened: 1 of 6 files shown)


anttimaki (approved):
LGTM

PR #2038, opened 2026-01-18
Fix UMM issues for native linux and proton

alexneargarder:
I noticed some issues when trying to launch launch Broforce with UMM installed on Linux, so I made some fixes to make it work. The code is pretty much just copied from the way that Bepinex does it.

I've tested it on Debian and the Steam Deck and it seems to work both with the native install and through proton as well.

I created a new release of UMM with the doorstop .so files, but I figured I'd wait until this PR is merged before uploading to Thunderstore: https://github.com/alexneargarder/unity-mod-manager/releases/tag/v1.1.0

### DIFF
--- a/src/r2mm/launching/instructions/instructions/loader/UMMGameInstructions.ts
+++ b/src/r2mm/launching/instructions/instructions/loader/UMMGameInstructions.ts
@@ -16,35 +16,38 @@ export default class UMMGameInstructions extends GameInstructionGenerator {
         }
     }
 
-    private getProfileArgs(profile: Profile): string[] {
-        if (["linux", "darwin"].includes(appWindow.getPlatform().toLowerCase())) {
-            return ['--r2profile', DynamicGameInstruction.PROFILE_NAME];
-        }
-        return ['--profile', profile.getProfilePath()];
-    }
-
     private async genDoorstopV3(game: Game, profile: Profile): Promise<GameInstruction> {
+        const launchArgs: string[] = [
+            '--doorstop-enable',
+            'true',
+            '--doorstop-target',
+            DynamicGameInstruction.UMM_PRELOADER_PATH
+        ];
+
+        if (["linux", "darwin"].includes(appWindow.getPlatform().toLowerCase())) {
+            launchArgs.push('--r2profile', DynamicGameInstruction.PROFILE_NAME);
+        }
+
         return {
-            moddedParameterList: [
-                ...this.getProfileArgs(profile),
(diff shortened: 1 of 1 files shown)


alexneargarder:
@ebkr I see there's a merge conflict now. It looks like the commit that changed that file is marked WIP. Should I go ahead and resolve the merge conflicts or will there be more stuff changing that I should wait for?

Also very cool that you're working on improving Steam Deck support. I didn't even realize that while I was making these changes.

ebkr:
You can resolve the merge conflicts. It's generally always best to anyway otherwise things get stale.

There's no major changes immediately planned so there shouldn't be any other conflicts?

alexneargarder:
@ebkr Hey I noticed there may be a slight bug with the new release. It seems like the help menu which shows the launch args for launching games from outside the mod manager is giving a comma separated list? 
For example:
image

I'm guessing that's unintentional? I included a fix in this commit, which is also included in this current PR: https://github.com/ebkr/r2modmanPlus/pull/2038/commits/c6fb911106ddcc81c8e7e92fd06108640eef4a4f

Which seems to work in my testing:
image

I also got the other changes from this PR updated to work with the new format and tested them again to confirm they're still working. So if you'd like to review it and merge we could kill two birds with one stone, or you can just copy the fix and do it in a separate PR. Either way is good with me, just let me know if you'd like me to revert it.

ebkr:
@alexneargarder It's incorrect, good catch!

I'll cherry pick the commit into a separate PR (unless you want to create a separate one), and we'll get that merged in sooner. 

I'm not the best person to review the actual UMM changes and are probably more relevant for one of: @anttimaki @VilppeRiskidev @ethangreen-dev

alexneargarder:
Yep that's fine you can go ahead and make the other PR and I'll just revert it here.

ebkr:
Created the PR here: https://github.com/ebkr/r2modmanPlus/pull/2049

Cherry-picked the commit before you nuked it so you get co-author contributions 💪 

Tested and merged

alexneargarder:
> why use hard coded `Z:` as part of profile path on Proton?

Yeah I'm not sure if it's best practice, but I figured it'd be fine since Bepinex is already doing it and it works in my testing. 
Thanks for the review.

ebkr:
`Z:` is the mount point for Proton as though it were a fake drive - It's necessary as the DLLs are located there.

Given that antti is happy with it and it continues to work on Windows, I'm happy to merge it.

Thanks 🔥

PR #2054, opened 2026-01-22
Use map lookups for LocalModList

ethangreen-dev:


src/store/modules/ProfileModule.ts
```diff
@@ -89,6 +89,14 @@ export default {
             return state.modList;
         },
 
+        modMap(state): Map<string, ManifestV2> {
```
ebkr, line 92: Is this not rebuilding the map every time this getter is called?

src/r2mm/mods/Dependants.ts
```diff
@@ -2,42 +2,64 @@ import ManifestV2 from '../../model/ManifestV2';
...
+                this.getDependantList(mod, modList, dependants, index);
+            }
+        }
         return dependants;
     }
 
-    public static getDependencyList(manifestMod: ManifestV2, modList: ManifestV2[], dependencySet?: Set<ManifestV2>): Set<ManifestV2> {
+    public static getDependencyList(
+        manifestMod: ManifestV2,
+        modList: ManifestV2[],
```
ebkr, line 48: I think we should be replacing modList entirely with modMap otherwise we're having double the memory consumption

src/r2mm/mods/Dependants.ts
```diff
@@ -2,42 +2,64 @@ import ManifestV2 from '../../model/ManifestV2';
 
 export default class Dependants {
 
-    public static getDependantList(manifestMod: ManifestV2, modList: ManifestV2[], dependantSet?: Set<ManifestV2>): Set<ManifestV2> {
+    private static buildDependantIndex(modList: ManifestV2[]): Map<string, ManifestV2[]> {
+        const index = new Map<string, ManifestV2[]>();
+        for (const mod of modList) {
+            for (const dep of mod.getDependencies()) {
+                const depName = dep.substring(0, dep.lastIndexOf('-'));
```
ebkr, line 9: Same outcome but you have changed the logic here. I'd probably say to revert so that it's a bit easier to read and doesn't rely on knowing the format as much.

ebkr (changes requested):

PR #2066, opened 2026-02-06
feat: enable OSX builds

patttterson:
I don't know exactly what changed in Github since #1447, but macOS builds seem to work fine right now.

Because Apple is Apple, the application won't open unless this suspicious looking `xattr -rd com.apple.quarantine /Applications/r2modman.app` is executed.

Even if macOS isn't officially supported by r2modman, it would be very nice to just provide the option for people to use it. You could also add a warning that macOS isn't officially supported to the app, so people don't spam the issues page with errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-> [!WARNING] 
+> [!WARNING]
 > There has been a rise in third-party redistribution sites which are claiming to be official.
 >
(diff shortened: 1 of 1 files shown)


ebkr:
It's not beneficial to do this at the moment because the macOS builds currently do not work.

I don't think there's a reason to add support and give people false hope until we have an actual working build

PR #2069, opened 2026-02-08
macOS support

LewsTherinTelescope:
Adds support for Mac, specifically for Steam and the direct game runner (I didn't realize until double checking just now that Epic Games Store actually exists on Mac). I have not implemented any of the publishing changes from #2066 though, because I figured that was probably a separate question.

I've tried to avoid touching logic for other platforms, but I did make some slight tweaks to the BepInEx wrapper, and the simplest way to call said wrapper in `DirectGameRunner.ts` ended up applying to Linux as well. It _should_ be completely fine, but I don't own a Linux machine to test, so if someone who does can verify I didn't explode anything that'd be appreciated.

Question: how should backward compatibility be handled? The current solution I have implemented here is deleting the old `macos_proxy` and `macos_wrapper.sh` files from the repo in favor of just using `linux_wrapper.sh`, with a check added to alert anyone using them that their launch options need to be updated. I figured this was probably not an enormous deal since Mac isn't officially supported right now anyway, and even for those who built it themselves the old code doesn't actually copy `macos_proxy` to the profile in the first place.

However, it did cross my mind that this could pose a future compatibility problem if the two scripts ever need to be split apart again; would it be better to keep using one or both of the old names, and just point them to call `linux_wrapper.sh` for now? If so, should the functionality instead be extracted to a new generic `unix_wrapper.sh` and make `linux_wrapper.sh` forward the call there too to make it clear that the logic is shared?

I've marked this as a draft for now until that question is decided and until I've taken a look at EGS.

### DIFF
--- a/src/components/computed/WrapperArguments.ts
+++ b/src/components/computed/WrapperArguments.ts
@@ -19,7 +19,7 @@ function determineWrapperArguments() {
     const _ = activeGame.value;
 
     let wrapperPath = "";
-    if (isFlatpakExecutable.value) {
+    if (isFlatpakExecutable.value || appWindow.getPlatform() === 'darwin') {
         wrapperPath = `${path.join(PathResolver.MOD_ROOT, 'web_start_wrapper.sh')}`
     } else {
         wrapperPath = path.join(PathResolver.MOD_ROOT, 'linux_wrapper.sh');
--- a/src/pages/Splash.vue
+++ b/src/pages/Splash.vue
@@ -166,6 +166,7 @@ async function moveToNextScreen() {
     } else if (appWindow.getPlatform() === 'darwin') {
         const activeGame: Game = store.state.activeGame;
         await ensureWrapperInGameFolder('linux_wrapper.sh');
+        await ensureWrapperInGameFolder('web_start_wrapper.sh');
         if (!(await areWrapperArgumentsProvided(activeGame))) {
             return router.push({name: 'linux'});
         }
--- a/src/r2mm/launching/runners/darwin/SteamGameRunner_Darwin.ts
+++ b/src/r2mm/launching/runners/darwin/SteamGameRunner_Darwin.ts
@@ -11,6 +11,9 @@ import LoggerProvider, { LogSeverity } from '../../../../providers/ror2/logging/
 import ChildProcess from '../../../../providers/node/child_process/child_process';
 import GameInstructions from '../../instructions/GameInstructions';
 import GameInstructionParser from '../../instructions/GameInstructionParser';
+import PathResolver from "../../../manager/PathResolver";
+import { ensureScriptsAreAllExecutable } from "../../../../utils/LaunchUtils";
+import { parse as parseShell } from "shell-quote";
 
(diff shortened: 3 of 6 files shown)


ebkr:
If this is converted to use the web_start_wrapper.sh, we can dynamically switch between linux_wrapper and any other new ones we add without requiring users to change launch args if we decide to rename it.

IMO the web wrapper is much more consistent and easier to work with.

public/linux_wrapper.sh
```diff
@@ -32,16 +35,38 @@ while [ $i -lt $max ]; do
 done
 
 if [ -z "$R2PROFILE" ]; then
-    echo "[R2MODMAN LINUX WRAPPER] Launching vanilla!"
+    echo "[R2MODMAN WRAPPER] Launching vanilla!"
     exec "$@"
-fi
+else
+    # Some old versions of run_bepinex.sh require being in the same directory as BepInEx
+    cd "$BASEDIR/profiles/$R2PROFILE/"
```
arrowmaster: This will break new versions of run_bepinex.sh running through Steam. Do not alter the working directory.

public/linux_wrapper.sh
```diff
@@ -32,16 +35,38 @@ while [ $i -lt $max ]; do
 done
 
 if [ -z "$R2PROFILE" ]; then
-    echo "[R2MODMAN LINUX WRAPPER] Launching vanilla!"
+    echo "[R2MODMAN WRAPPER] Launching vanilla!"
     exec "$@"
-fi
+else
```
arrowmaster: Do you really need to indent every thing below to use an else here? The above exec stops any further processing of the script if its triggered.

LewsTherinTelescope:
Ah that does seem tidier, I'll switch to that then.

ebkr:
I've currently tested this on macOS with Timberborn and I was unable to get it to load. I think it may be an issue with my game but I'll keep you updated when I next get around to testing.

LewsTherinTelescope:
Hm, is it showing a specific error, or just silently not launching, or launching into the vanilla game, or what? Unfortunately I don't own Timberborn so can't check it myself.

ebkr:
I've also tried it with Lethal League Blaze with the following launch args: `/usr/bin/env WINEDLLOVERRIDES="winhttp,version=n,b" "/Users/cade/Library/Application Support/r2modmanPlus-local/LethalLeagueBlaze/web_start_wrapper.sh" %command%` and it doesn't start.

Also tried with `"/Users/cade/Library/Application Support/r2modmanPlus-local/LethalLeagueBlaze/web_start_wrapper.sh" %command%`

I can't find any meaningful logs from the Console app.

LewsTherinTelescope:
Weird... Does it work if you run the command from the terminal directly, replacing `%command%` with the path to the game binary? (I have no idea what your base knowledge is on this, so just in case: right-click in Steam library → `Browse Local Files` → right-click whatever `.app` file is present → `Show Package Contents` → `Contents` → `MacOS` → select whatever binary file is present → press `Cmd+Opt+C` to copy path.)

Trying to figure out if the problem lies with the launcher script, the config it says to give Steam, etc.

PR #2070, opened 2026-02-08
add OnTogether.exe name

patttterson:
No clue when this happened, but the exe renamed. Figured might as well keep the old one, it's unique enough
image
image


ebkr:
Thanks - A little bit annoying but all ecosystem edits should be done here: https://github.com/thunderstore-io/ecosystem-schema

PR #2071, opened 2026-02-09
Added tracking for online vs local sourced mods

ebkr:
# Tracking Online Sources

_Prerequisite to identifying if package versions no longer match listed packages on Thunderstore_

## What's covered?

| Action                   | In cache | Online/Local tracked |
|--------------------------|----------|----------------------|
| Download, cache enabled  | No       | Online               |
| Download, cache disabled | No       | Online               |
| Download, cache enabled  | Yes      | Based on last result |
| Download, cache disabled | Yes      | Online               |
| Import                   | No       | Local                |
| Import                   | Yes      | Local                |

## Functionality

The cache has a new untracked folder called "_state" (to mirror the profile's _state folder), which records a list of files and versions that were downloaded from an online source.

Upon download the state files are looked up and if previously in the cache (and caching is enabled), then pull from the cache and set the onlineSource property equivalent to if the state file exists.

If the file is not cached (or caching is disabled), then fetch it from Thunderstore, and write the appropriate state file to disk. Repeat the lookup as part of the previous step.

If a file is locally imported, do not write a state file.

## Follow-up work for later PRs

- Inform the user if the mod is already cached
- If an `onlineSource` enabled mod isn't in the online cache, then inform the user that it is a potentially vulnerable package

### DIFF
--- a/src/model/ManifestV2.ts
+++ b/src/model/ManifestV2.ts
@@ -31,6 +31,8 @@ export default class ManifestV2 {
     private enabled: boolean = true;
     private icon: string = '';
 
+    private onlineSource: boolean = false;
+
     // Intended to be used to import a mod with only minimal fields specified.
     // Should support manifest V1. Defaults to an "Unknown" author field if not found.
     public makeSafeFromPartial(data: any): R2Error | ManifestV2 {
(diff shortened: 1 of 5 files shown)


anttimaki:
Should the local import explicitly check if `_state/online` exists in the zip file being imported and ignore it while extracting (or remove it after extraction, or something)? Theoretically malicious mods downloaded from external sources and imported locally could otherwise bypass this feature.

ebkr:
> Should the local import explicitly check if `_state/online` exists in the zip file being imported and ignore it while extracting (or remove it after extraction, or something)? Theoretically malicious mods downloaded from external sources and imported locally could otherwise bypass this feature.

I don't think we need to worry.

Cache items would be placed in:
`/cache/Author-ModName/Version/...`

These are tracked in `/cache/_state/online/...`

`_state` is an invalid team name.

If we wanted further safety, we could check if the secondary folder name is a version number or not, but I don't think it's particularly necessary.

x753 (approved):
Code looks sane, didn't build and test or anything though.

PR #2073, opened 2026-02-10
Remove shimloader default install rules

ethangreen-dev:


ethangreen-dev:
Reasonable, both ways are about the same amount of work. I'll update the installer, tests, and then the relevant schema entries.

PR #2077, opened 2026-02-17
FIX: Launch Steam with xdg-open for Flatpak release

### LINKED ISSUE #2076: [BUG] - Flatpak Release Cannot Launch Flatpak Steam
**Describe the bug**
On a desktop, when using the Flatpak release of R2 with Flatpak Steam, the "Start Modded" and "Start Vanilla" buttons don't launch Steam.

**To Reproduce**
0. Have Flatpak Steam installed and R2 Modman installed as a Flatpak
1. Go to a game profile page
2. Click on "Start Modded" or "Start Vanilla"
3. Watch for errors/Steam not loading

**Expected behavior**
Steam to launch and load the game

**Additional context**
For me, when I click the button, I get this error: `You are missing the following 32-bit libraries, and Steam may not run: libc.so.6`. From what I can tell, this is because R2 Modman attempts to run steam.sh to launch Steam, which doesn't work to launch the Flatpak version (and is also where the 32-bit library error comes from).

TB516:
Changes the Steam executable script to call xdg-open instead of flatpak-spawn, and the SteamGameRunner start method to pass the Steam protocol with the current app id. Would close #2076.

### DIFF
--- a/public/steam_executable_launch.sh
+++ b/public/steam_executable_launch.sh
@@ -1,10 +1,10 @@
 #!/bin/sh
 
 if [ $# -eq 0 ]; then
-    echo "Error: Steam app launch arguments must be provided" >&2
+    echo "Error: flatpak-spawn arguments must be provided" >&2
     exit 1
 fi
 
-echo "Executing: /usr/bin/flatpak-spawn --host steam -applaunch $@" >&1
+echo "Executing: /usr/bin/flatpak-spawn $@" >&1
 
-exec flatpak-spawn --host steam -applaunch "$@"
+exec flatpak-spawn "$@"
--- a/src/r2mm/launching/runners/linux/SteamGameRunner_Linux.ts
+++ b/src/r2mm/launching/runners/linux/SteamGameRunner_Linux.ts
@@ -103,7 +103,8 @@ export default class SteamGameRunner_Linux extends GameRunnerProvider {
         try {
 
             if (await isManagerRunningOnFlatpak()) {
-                console.log("Launching flatpak")
+
+                LoggerProvider.instance.Log(LogSeverity.INFO, `Launching using Flatpak behaviour`);
 
                 const isProton = await getDeterminedLaunchType(game, settings.getLaunchType() || LaunchType.AUTO) === LaunchType.PROTON;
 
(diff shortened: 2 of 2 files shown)


TB516:
With these changes, the Flatpak version of R2 can launch both normal and Flatpak Steam, but I could also make changes so the AppImage does the same thing, just lmk and I'll get on it.

ebkr:
I've just ran this on Steam Deck in Game Mode, and I get:

> Failed to call portal:GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such interface ?org.freedesktop.portal.OpenURI? on object at path /org/freedesktop/portal/desktop

It does however work in regular mode 🤔

TB516:
Hm weird. I don't have a Steam deck but I'll do some research to see if there is anything to fix that.

ebkr:
Happy to run things on mine, but my turnaround will be relatively slow. It's quite tedious to do dev builds on it 💀

TB516:
It seems like xdg-open is just not available in game mode? I'll see if there are any workarounds or anything we can do to work around it. There might be an environment variable to let us decide which way we should launch, with xdg-open being the default.

arrowmaster:
> It seems like xdg-open is just not available in game mode? I'll see if there are any workarounds or anything we can do to work around it. There might be an environment variable to let us decide which way we should launch, with xdg-open being the default.

Sounds like because its not a full desktop either something with XDG environment variables or DBus isnt fully setup. I'd be curious what `xdg-mime query default x-scheme-handler/steam` returns from game mode and if that would be a good test.

TB516:
Ya from some more searching, it seems like the portal doesn't work in game mode. Im going to try to see if doing `flatpak-spawn --host xdg-open url` would work.

TB516:
@ebkr Here's the releases from the most recent build: https://github.com/TB516/r2modmanPlus/releases. When you get a sec, if you could test it, that would be awesome.

ebkr:
> @ebkr Here's the releases from the most recent build: https://github.com/TB516/r2modmanPlus/releases. When you get a sec, if you could test it, that would be awesome.

image

💀

TB516:
Apparently I needed to publish it, try now @ebkr

ebkr:
> UncaughtException: 
> Error: Command Failed: ...
> /usr/bin/xdg-open: line 1045: www-browser: command not found

(produces the same for the following)
- links2
- elinks
- links
- lynx
- w3m

> xdg-open: no method available for opening steam://run/632360

I can't directly copy/paste the content on the Deck annoyingly, so it's all manual typing

TB516:
Ok, I just made some changes, so if it detects env variables that should be used for the Steam Deck game mode, it will try to run the script/flatpak-spawn run the Steam binary on the host; otherwise, use the xdg-open method. Should hopefully work. Link from the workflow to test here: https://github.com/TB516/r2modmanPlus/releases.

ebkr:
@TB516 I've pushed a change to the branch. Pull and give it a go to see if it still works. I've ran it on my end on both the Steam Deck and my CachyOS install (non-flatpak Steam, flatpak r2) and it works as expected.

TB516:
I'll give it a test tomorrow morning, but from looking it over it should be fine. I do want to make a note though: the wrapper script taking all arguments and passing it to flatpak-spawn might be an issue for publishing on flathub. If the steam binary wasn't found at all on deck mode, then maybe we make the script just take the path to steam.sh and the protocol? Otherwise, I could dig into the issues you were having with node's child processes to see if I can resolve it/move the flatpak-spawn command to a more controlled environment.

TB516:
Also something I've been thinking about, if you want (and it works, though, I don't see why it wouldn't), I could make changes so no matter if it's an appimage or a flatpak, if we aren't on the steam deck in game mode, it uses xdg-open. Even though I'd imagine most people using flatpak steam would want to use r2 as a flatpak, it should give the appimage the ability to launch all versions of steam.

ebkr:
> the wrapper script taking all arguments and passing it to flatpak-spawn might be an issue for publishing on flathub

I think it'll be an issue when we get to it, but I'd imagine they can understand the purpose of it given the nature of a mod manager.

In addition to that, I'm pretty sure the execution script is safe from injection due to the order that we pass values through.

> Also something I've been thinking about, if you want (and it works, though, I don't see why it wouldn't), I could make changes so no matter if it's an appimage or a flatpak, if we aren't on the steam deck in game mode, it uses xdg-open. Even though I'd imagine most people using flatpak steam would want to use r2 as a flatpak, it should give the appimage the ability to launch all versions of steam.

The plan is to eventually switch everyone over to the web wrapper, but it's not a move I want to make just yet because I'd want it to be in-sync across all platforms and I don't have the time to spend on that yet.

In theory, all platforms can support the steam://run protocol, even Windows. But I'd also like to avoid having to have people manually add launch arguments if possible. Especially on Windows. 

It's quite a big time investment and probably not that worthwhile atm

TB516:
Fair enough, and that makes sense for moving everything over all at once. Back to the changes you made earlier, the actual logic for launching the flatpak version didn't change at all, and I've been running the custom build for the past few days with no issues, so from my perspective, it should be good to merge.

TB516:
Sorry for the delay but I got a chance to run it, and ya it still works perfectly!

ebkr:
Merged, thank you 🔥

TB516:
Glad to help!

PR #2080, opened 2026-02-28
InstallRulePluginInstaller

anttimaki:
Currently some of the mod installation logic is split between GenericProfileInstaller and InstallRuleInstaller. This PR begins a journey that aims for the following improvements:

- Make InstallRulePluginInstaller a proper self-contained plugin installer
- Remove mod loader specific loader/plugin installation logic from GenericProfileInstaller
- Ensure each supported mod loader explicitly defines which installer and plugin installer should be used when dealing with a game that uses the mod loader
- Clean up the legacy code which is unnecessarily complicated now that Thunderstore ecosystem schema acts as the source of truth for things that used to be embedded in the mod manager codebase

This first PR contains mostly cleanup and preparation for the actual changes.


src/installers/InstallRulePluginInstaller.ts
```diff
@@ -256,7 +257,58 @@ async function installState(args: InstallRuleArgs) {
...
+                    }
+                }
+            }
+        }
+    } catch (e) {
+        return R2Error.fromThrownValue(e, `Error installing mod: ${mod.getName()}`);
+    }
+}
+
+export class InstallRulePluginInstaller implements PackageInstaller {
```
ebkr, line 311: This file should be renamed to match

src/installers/NoOpInstaller.ts
```diff
@@ -0,0 +1,13 @@
+import { InstallArgs, PackageInstaller } from './PackageInstaller';
+
+/**
+ * Doesn't do anything. Provides combatible API for games that
```
ebkr: Small typo
```suggestion
 * Doesn't do anything. Provides compatible API for games that
```

src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
```diff
@@ -189,16 +144,8 @@ export default class GenericProfileInstaller extends ProfileInstallerProvider {
...
-        if (loaderInstaller) {
-            await loaderInstaller.install(args);
-            return Promise.resolve(null);
-        } else {
-            return new R2Error(
-                "Installer not found",
-                `Failed to find an appropriate installer for the package ${mapping.packageName}`
-            );
-        }
+        await PackageLoaderInstallers[mapping.loaderType].install(args);
```
ebkr, line 147: Might just be missing something obvious but is there a reason we no longer throw an error if it's not found?

src/installers/InstallRulePluginInstaller.ts
anttimaki: So we currently have `FooInstaller` naming pattern that currently in some cases contains `FooInstaller`, and in some cases also `FooPluginInstaller`. I think this is the first file to contain only `FooPluginInstaller` though. Given this, should we split the other files too (in which case I'll do it separately later), or are we happy with having just `InstallRulePluginInstaller`?

src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
anttimaki: - `PackageLoaderInstallers` defines values for all loaders. It's typed so that if a loader is missing, it causes a TypeScript error
- I'm like 90% sure that when I created this back in August or something, the type error would cause the app from compiling and thus be obvious that developer forgot to update the registry
- However, this seems no longer to be the case. I didn't check if the timelines match, but moving to vite might have caused this
  - According to LLM this is due to the build currently skipping type checking, and if we so choose it can be turned on via config. This would make sense to me. Alternatively we could add another yarn command that does a typecheck and run it in CI, but IMO that's poorer DX
- Another aspect is the real-time game addition feature, which may introduce new package loaders, but even in that case I think the checkup should be done when ecosystem schema is synced, not on runtime while installing a package
- Finally, at the end of my planned changes, GenericProfileInstaller will be a rather simple wrapper for the installers. It might at that point have an unified error-catching on all the public methods that are provided

Thoughts?

ebkr: > PackageLoaderInstallers defines values for all loaders. It's typed so that if a loader is missing, it causes a TypeScript error

Fair enough. It removes the need for it so that makes sense.

> According to LLM this is due to the build currently skipping type checking, and if we so choose it can be turned on via config.

I'd thought we did have type checking, although given the odd file with type errors (EG: missing `!`), we probably don't. Counter-point though is that it really depends on how long it takes to run the type checking too, and if that affects `yarn run run` or not.

> Another aspect is the real-time game addition feature, which may introduce new package loaders, but even in that case I think the checkup should be done when ecosystem schema is synced, not on runtime while installing a package

My assumption was that we'd filter these out if the type didn't exist so we wouldn't have this option anyway. I wouldn't expect a user to select a game we can't install a mod loader package for.

> Finally, at the end of my planned changes, GenericProfileInstaller will be a rather simple wrapper for the installers. 

Sounds good to me 👍

src/installers/InstallRulePluginInstaller.ts
ebkr: I think we probably would want to separate the files, but if they share logic then it becomes a bit annoying.

Is there a plan to move all of them from `FooInstaller` to `FooPluginInstaller`, or are they distinctly separate topics which we can separate?

But in general for now I'm happy with just renaming this file. Ideally you should be able to search for a class by file name IMO.

anttimaki: Renamed the single file for now.

src/installers/NoOpInstaller.ts
anttimaki: Fixed.

src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
anttimaki: - Running `tsc` for the first time took a while, subsequent runs were ~3.4 seconds. Dunno if that's too much compared to the hassle of having to remember run a separate typecheck yarn command locally or to see the CI fail is IMO debatable.
- According to LLM chat we can control type checking in dev and build modes separately. Might require some additional dependencies (`vue-tsc`, `vite-plugin-checker`). Didn't test this in practice.
- Running `tsc` shows 121 type errors in the current code base. I think that's a strong argument we need the type checking somewhere in the pipeline. To implement that would require also fixing those type errors, which I consider out of scope for this PR. But Someone™ should look into that sooner rather than later.

ebkr: Might be worth raising as a refactor task, but I think realistically an IDE should pick up any issues (WebStorm does tell me about errors now), and so we should only need CI related checks to confirm.

If we have the tsc pipeline as part of a package.json script, we can run both locally and in CI with the same setup 🤷

PR #2086, opened 2026-03-10
Remove Beetleball melonloader override

ethangreen-dev:
Beatle does not equal beetle. Noted and remembered.


anttimaki:
Changes already merged to develop.

PR #2089, opened 2026-03-11
Baby Steps Server Support

ebkr:
- Required a change to DirectGameRunner execution. We now use `spawn` instead of `exec`.
- Tested with Baby Steps Server.
- Tested using an existing game (Timberborn) to ensure the "Other" option still launches games correctly.


src/r2mm/launching/runners/multiplatform/DirectGameRunner.ts
```diff
@@ -45,20 +46,23 @@ export default class DirectGameRunner extends GameRunnerProvider {
 
             const mappedArgs = args.map(value => `"${value}"`).join(' ');
```
anttimaki, line 47: Changes below leaves this row unused.

PR #2093, opened 2026-03-17
Prepare repo for Flathub build process

TB516:
# Prepare r2modman repository for Flathub build processes

This PR updates the current Flatpak build process to mirror the one used by Flathub, adding and changing the files needed to do this. This will be a large step toward completing #729.

## App changes
Changes to the actual runtime behavior of the app

- Added a guard to the auto updater to skip checking for/applying updates when the app is running as a Flatpak

## Project changes
Changes to the project structure/repo and CI

- Updated mise to add python/pipx as a project tool (needed to install flatpak-node-generator)
- Added Flatpak-related files to the flatpak directory
    - `io.github.ebkr.r2modman.yaml` is the Flatpak manifest that contains the instructions to build the Flatpak
    - `io.github.ebkr.r2modman.metainfo.xml` is the appstream metadata used by storefronts
    -  `io.github.ebkr.r2modman.desktop` is the .desktop file that is used to launch the Flatpak
    - `io.github.ebkr.r2modman` is a launch script that is used to run the r2modman binary inside of the Flatpak
    - `generate-release-manifest.js` is a quick script to take the manifest that is in the git repo, and update the [flatpak sources](https://docs.flatpak.org/en/latest/module-sources.html) to pull from git instead of using the local file system

### Documentation
- Added images of r2modman to the docs directory, which will be used in the Flathub metainfo

### Build config
- Removed previous Flatpak building config from quasar and package.json
- Added a script to build an unpackaged Linux build (built with electron binaries, but not packaged into a distribution format)
- Added a script to generate a modified Flatpak manifest file that can be used on Flathub & other CI runners
- Added a script to generate `generated-sources.json`, a mapping of yarn lockfile deps to flatpak sources
- Added a script (and helper scripts) to build r2modman as a Flatpak

### CI
- Removed old flatpak build step from build workflow and release workflow
- Added a new build-flatpak workflow to run a standalone flatpak build using the new manifest
- Added a new job to the release workflow that builds and releases a standalone `r2modman.flatpak` file using the new manifest and flatpak-builder (mimicks how Flathub would build the app)

## What's missing

Currently, this PR doesn't have any workflows to automatically update the Flathub repo, as the submission hasn't started yet. If/once r2modman is accepted, then workflows can be added to push the generated manifest and the other files in the Flatpak directory to the Flathub repo.


.github/workflows/build.yml
ebkr, line 1: Have you ran this build step?

I would have errors without the specific addition to Flatpak dependency/remote installing.

We still want to build Flatpak images to make sure they generate correctly.

.github/workflows/release.yml
ebkr, line 1: We should still publish an image

.yarn/releases/yarn-4.11.0.cjs
ebkr, line 1: Can we have a command to generate this instead?
Not comfortable accepting this in a review.

I can run this myself and commit if we actually need it in the repo.

package.json
```diff
@@ -7,18 +7,18 @@
...
   "license": "MIT",
   "type": "module",
+  "main": "src-electron/electron-main.ts",
   "scripts": {
     "run": "quasar dev -m electron --modern",
     "dev": "quasar dev -m electron --modern --devtools",
     "sync": "tsx ./scripts/sync.ts",
     "build-win": "quasar build --mode electron -T win32",
     "build-linux": "quasar build --mode electron -T linux",
-    "build-flatpak": "BUILD_FLATPAK=true quasar build --mode electron -T linux",
```
ebkr, line 16: We should keep this. A raw flatpak image is useful.

package.json
```diff
@@ -7,18 +7,18 @@
   "private": true,
   "license": "MIT",
   "type": "module",
+  "main": "src-electron/electron-main.ts",
```
ebkr: Do we actually need this?

quasar.config.ts
```diff
@@ -1,13 +1,11 @@
 // Configuration for your app
 // https://v2.quasar.dev/quasar-cli-vite/quasar-config-file
 
-import { defineConfig } from '#q-app/wrappers';
-import { nodePolyfills } from 'vite-plugin-node-polyfills';
+import { defineConfig } from "#q-app/wrappers";
+import { nodePolyfills } from "vite-plugin-node-polyfills";
```
ebkr: Keep these as single quotes.

Applies to the rest of the file, in general we use single quotes.

quasar.config.ts
```diff
@@ -180,109 +179,63 @@ export default defineConfig((ctx) => {
...
                 // appBundleId: '',
                 // appCategoryType: '',
                 // osxSign: '',
                 // protocol: 'myapp://path',
-
                 // Windows only
                 // win32metadata: { ... }
             },
 
             builder: {
```
ebkr, line 214: Related to other comments, this should still be needed

ebkr (changes requested):

quasar.config.ts
TB516: I think my prettier extension had formatted with its defaults, but this should be fixed now.

TB516:
I addressed some of the smaller issues. For building a standalone flatpak file, im going to bring the conversation over to the Discord for better communication, but here is a quick summary of my thoughts: if we still want to build a standalone flatpak file, I think the best approach would be to add a job to the release worklfow that uses flatpak-builder to build from the flathub manifest, as that would make sure that our standalone flatpak file is the same as the one people would get from flathub, and ensure there isnt any drift between the two.

TB516:
Moving this to draft while the changes are being addressed.

TB516:
I added the workflows & scripts to do a local Flatpak build, along with building & releasing a standalone Flatpak file, so I think all of the feedback should be covered now!

.github/workflows/flatpak-node-generator.yml
ebkr, line 1: Might be good to have a comment explaining what this is for. (I don't know)

flatpak/generate-release-manifest.js
```diff
@@ -0,0 +1,84 @@
...
+
+if (!module) {
+  fail('Could not find module "r2modmanPlus"');
+}
+
+if (!Array.isArray(module.sources) || module.sources.length === 0) {
+  fail('Module has no sources');
+}
+
+// Replace ONLY the first source entry
```
ebkr: Why?

flatpak/generate-release-manifest.js
```diff
@@ -0,0 +1,84 @@
...
+}
+
+const { input, output, repo, commit } = parseArgs(process.argv);
+
+const raw = fs.readFileSync(input, 'utf8');
+const manifest = YAML.parse(raw);
+
+if (!manifest || typeof manifest !== 'object') {
+  fail('Invalid YAML manifest');
+}
```
ebkr: Is this needed? Does it not throw an error if it's invalid?

flatpak/generated-sources.json
ebkr, line 1: Should this have been committed?

flatpak/io.github.ebkr.r2modman.metainfo.xml
```diff
@@ -0,0 +1,50 @@
...
+    <name>ebkr</name>
+  </developer>
+  <description>
+    <p>The official Flathub release of r2modman, a simple and easy to use mod manager for many games using Thunderstore.</p>
+
+    <p>Important Notes:</p>
+    <p>**Some Flatpak overrides might be needed!**</p>
+
+    <ul>
+      <li>You must ensure that the r2modman Flatpak has permissions to access the directory containing the game files for the game you want to mod. The only default supported locations are <code>~/.local/share/Steam</code> and <code>~/.var/app/com.valvesoftware.Steam</code></li>
```
ebkr: We should get access here by default with the `host` permission

package.json
```diff
@@ -11,14 +11,15 @@
     "run": "quasar dev -m electron --modern",
     "dev": "quasar dev -m electron --modern --devtools",
     "sync": "tsx ./scripts/sync.ts",
+    "generate-release-manifest": "node flatpak/generate-release-manifest.js --input flatpak/io.github.ebkr.r2modman.yaml --output flatpak/io.github.ebkr.r2modman.release.yaml --repo https://github.com/ebkr/r2modmanPlus.git --commit $GITHUB_SHA",
+    "install-flatpak": "flatpak-builder --user --install --force-clean flatpak/dist flatpak/io.github.ebkr.r2modman.yaml",
```
ebkr: Do we need to `install`?

Building is usually enough, and then a manual `flatpak install` against the built file.

The reason is because builds usually aren't only for local testing but for sharing with other people to ensure that issues have been resolved.

I barely run built files as I spend most of my time in the dev hot-reloadable environment anyway

flatpak/README.md
ebkr, line 1: I think it'd be good to include a chain of steps involved in a release.

EG:

```
# Making a release

1. Update `io.github.ebkr.r2modman.metainfo.xml` with the new release information
2. ... Commands to run
```

ebkr (changes requested):
Added some more comments, nothing major though.

For comments that aren't being actioned, leave a comment as to why (for future reference), and mark as resolved.

.github/workflows/flatpak-node-generator.yml
TB516: Just added some comments to the workflow file to give a high-level overview of what the workflow does and why. Lmk if I should add some more in-depth comments throughout the workflow!

```
# This regenerates the generated-sources.json file 
# It does this whenever the yarn lockfile is changed
#
# generated-sources.json needs to be up to date with the yarn lockfile
# as it maps the yarn packages to downloadable tar.gz files
# that are used by flatpak-builder (because flatpak-builder doesn't have network access when building the .flatpak)
```

flatpak/generate-release-manifest.js
TB516: In the manifest that we are committing to the repo, the first source is the current directory (so when you run a local build, it builds with the files on your disk), but for flathub, we would want the source to be something that is stable and reproducible, so the script changes it to a git source block with the current commit. I just have it replacing the first source because it was quick and easy, but I can change it to find the exact directory source if you want.

flatpak/generate-release-manifest.js
TB516: Just updated this to try/catch instead, should be good now!

flatpak/generated-sources.json
TB516: The file essentially converts the yarn lockfile to archives that are downloaded by flatpak-builder, so technically, we could not commit it to this repo, but then it would be generated every time, even if the lockfile didn't change. It would also need to be in the flathub repo, since we can't pre-generate it before the CI run there, so i think overall, the best way to do things is keep it committed.

flatpak/io.github.ebkr.r2modman.metainfo.xml
TB516: I think getting the `host` permission is a super long shot, so I just have the default filesystem access include `~/.var/app/com.valvesoftware.Steam` and `~/.local/share/Steam:rw`, which is why the metainfo says this. I think it would be best to update it once we confirmed have the permissions.

package.json
TB516: I'll update this to just build a standalone file then!

TB516: Ok, just updated the scripts so `yarn build-flatpak` will build the flatpak repo, then bundle it to the .flatpak file.

flatpak/README.md
TB516: I think the only extra step should be to update the metainfo.xml, but if there is a location that has the release steps, I'll update it real quick.

flatpak/generated-sources.json
ryanabx: I was wondering this too. I don't typically see flatpak generated-sources files tracked in repos. Usually they are generated on-the-fly for CI builds, and tracked in the flathub repos, but not usually in the upstream repo.

It might get cumbersome to update both the yarn lockfile and the generated sources together, right?

In my rust projects I don't check in the cargo generated-sources file for this reason.

TB516: > It might get cumbersome to update both the yarn lockfile and the generated sources together, right?

I think the workflow should help with this, but ya, it might be a good idea to just generate it every time.

TB516: The only annoyance would be that when doing a local build, youd need to generate the sources too.

ryanabx: > The only annoyance would be that when doing a local build, youd need to generate the sources too.

You could potentially make a shell script that does the generate sources step before doing a flatpak build.

ebkr: There should be support for `pre` hooks as part of commands.

I believe `yarn build-flatpak` would run:

`yarn prebuild-flatpak`
`yarn build-flatpak`
`yarn postbuild-flatpak` 

If all were defined.

Worth looking in to.

TB516: I can make it so its generated by the pre-hook instead of the workflow if thats what we want to do instead.

WesCook:
I just want to say that I'm looking forward to having r2modman available via flatpak.  It should make modding certain games on immutable Linux much easier.  Thanks for your efforts, Thomas and Cade.

.github/workflows/flatpak-node-generator.yml
TB516: Made the changes to generate the workflow locally so this is removed now

TB516:
Made the changes to generate the node sources locally with flatpak-node-generator, and updated the scripts to build a local version instead of building and installing it. I added some helper scripts to do this, and added python/pipx as a dependency in mise as it is probably the easiest way to get flatpak-node-generator for development.

I wana make a note that while for local dev, `yarn build-flatpak` is used, im still using the `flatpak/flatpak-github-actions/flatpak-builder` action in CI because it takes care of a few extra steps that aren't needed for locally building the flatpak.

flatpak/generate-release-manifest.js
TB516: Just updated the script to use a function to find the `{ type: 'dir', path: '..' }` source block and replace that instead of always replacing the first one.

.github/workflows/release.yml
```diff
@@ -74,3 +55,50 @@ jobs:
...
+
+      - name: Mark checkout as safe for git
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
+      - name: Append standalone flatpak to release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG_NAME: ${{ steps.tag.outputs.tag }}
+        run: |
+          gh release upload "$TAG_NAME" r2modman.flatpak --clobber
```
ebkr, line 104: > When using --clobber, existing assets are deleted before new assets are uploaded. If the upload fails, the original assets will be lost.

This sounds problematic?

flatpak/generate-release-manifest.js
```diff
@@ -0,0 +1,91 @@
...
+    else if (arg === '--output') args.output = argv[++i];
+    else if (arg === '--repo') args.repo = argv[++i];
+    else if (arg === '--commit') args.commit = argv[++i];
+    else fail(`Unknown argument: ${arg}`);
+  }
+
+  if (!args.input || !args.output || !args.repo || !args.commit) {
+    fail(`
+      Usage:
+        node script.mjs \
```
ebkr: Assuming `script.mjs` here refers to `generate-release-manifest.js` ?

yarn.lock
ebkr, line 1: I think this is the last real concerning thing.

I wouldn't have expected any version changes for existing dependencies in the package lock unless those dependencies were changed in the package.json. That's the point of the lockfile 🤔

ebkr (changes requested):

.github/workflows/release.yml
TB516: That should only be for the one file that we are uploading, and I had it clobber because if the workflow was being ran a second time, then the upload step would fail. I figured its better to do it like that, but if the release workflow is only going to be ran once, then I can change it to not clobber.

flatpak/generate-release-manifest.js
TB516: Yep just fixed that!

yarn.lock
TB516: I think it would have been changed when the `electron-installer-flatpak` dep was removed, then re-added as an optional dep in a merge from main, then removed again by me after the merge. Thats the only changes im seeing that could've caused any lockfile changes, but I can take some more time tomorrow to see if theres anything im missing.

TB516: Not super sure what happened there but should be good now.

ebkr (approved):

ebkr (changes requested):
Waiting for rebase

ebkr (approved):

PR #2104, opened 2026-03-30
Extract mod uninstallation logic from GenericProfileInstaller

anttimaki:


src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
```diff
@@ -215,23 +147,13 @@ export default class GenericProfileInstaller extends ProfileInstallerProvider {
...
-        const uninstallSubDir = await this.uninstallSubDir(mod, profile);
-        if (uninstallSubDir instanceof R2Error) {
-            return uninstallSubDir;
-        }
-        try {
-            await this.uninstallPackageZip(mod, profile);
+            // Fallback to legacy uninstallation.
+            throwForR2Error(await this.uninstallModLoader(mod, profile));
+            const args = this.getInstallArgs(mod, profile);
+            await this.legacyInstaller.uninstall(args);
```
ebkr, line 154: What do we not cover with these changes? Why the need for the fallback?

ebkr (approved):

src/r2mm/installing/profile_installers/GenericProfileInstaller.ts
anttimaki: This is just a move-code-around refactoring commit, so to keep the mod manager behaving like it did before the commit, I don't remove the legacy fallback here. (TBH I don't remember if there's actually some use cases that actually still hit this code path at this point.) It will be removed in an upcoming PR once all the logic has been moved from GenericProfileInstaller to PackageInstaller implementations.

PR #2124, opened 2026-04-14
Added explicit rejection of executables from profile exports

ebkr:


ethangreen-dev (approved):
We should be performing this validation on the API, but as a half-measure this is fine for now.

ebkr:
LGTM 👍

PR #2125, opened 2026-04-14
Update spinner

8Qbit:
Adds a new EcosystemUpdateModule Vuex store module that tracks game list update progress (isInProgress, error) via startUpdate, finishUpdate, and setError mutations

Adds a StatusIndicator component to the activity bar on the game selection screen, displaying a priority chain of statuses: ecosystem update > download progress > mod list sync > default ("You have the latest game list")


src/components/navigation/StatusIndicator.vue
```diff
@@ -0,0 +1,119 @@
+<template>
+    <div
+        :class="['status-indicator', status.severity, { 'status-indicator--clickable': status.onClick }]"
```
ebkr, line 3: This feels weird for two reasons:

1. You effectively have:
```html
<div class="status-indicator info status-indicator--clickable">
```
where I think we'd expect `status-indicator--info` for the second one for consistency.

2. You are conditionally assigning the `'status-indicator--clickable'` class based on a not boolean value. `status.onClick` is a method definition, not a boolean.

src/components/navigation/StatusIndicator.vue
```diff
@@ -0,0 +1,119 @@
...
+    iconClass: string;
+    severity: 'info' | 'warning' | 'error';
+    onClick?: () => Promise<void>;
+    tooltip?: string;
+}
+
+const status = computed<IndicatorStatus>(() => {
+    if (store.state.ecosystemUpdate.isInProgress) {
+        return {
+            text: "Updating game list...",
```
ebkr, line 30: Nit: I don't think we need the `...`. The spinner is an indicator that something is happening

src/components/navigation/StatusIndicator.vue
```diff
@@ -0,0 +1,119 @@
...
+
+    if (store.getters['tsMods/isModListOutdated']) {
+        return {
+            text: "Outdated online mod list",
+            iconClass: "fas fa-exclamation-triangle",
+            severity: "warning",
+            tooltip: "Refresh",
+            onClick: () => store.dispatch('tsMods/syncPackageList'),
+        };
+    }
```
ebkr, line 78: These aren't relevant to the PR

src/components/navigation/StatusIndicator.vue
ebkr, line 1: Worth renaming this file to something like: `EcosystemUpdateIndicator.vue`

src/App.vue
```diff
@@ -7,6 +7,7 @@
                 </div>
             </main>
             <div id="activity-bar">
+                <StatusIndicator v-if="router.currentRoute.value.path === '/'" />
```
ebkr, line 10: This shouldn't be here.

The other example of the activity bar usage uses a [Teleport](https://vuejs.org/guide/built-ins/teleport) and includes the component only on the relevant pages. Using v-if and querying the route isn't particularly suitable

src/App.vue
```diff
@@ -194,8 +196,15 @@ main {
         background-color: var(--preview-panel-background-color);
         border-top: 1px solid var(--border, #e1e1e1);
         padding: 0.25rem 0.75rem;
+        min-height: 45px;
```
ebkr, line 199: I don't think we should be setting a minimum height. We should be making sure that the text element itself has consistent padding.

src/App.vue
```diff
@@ -194,8 +196,15 @@ main {
...
         gap: 0.5rem;
         overflow: hidden;
     }
+
+    > .status-indicator {
+        margin-left: auto;
+        flex-shrink: 0;
+        order: 999;
```
ebkr, line 207: Do we need `order`?

I'd opt for correctly positioning the item semantically in HTML rather than arbitrarily setting an order via CSS.

ebkr (changes requested):

src/components/navigation/EcosystemUpdateIndicator.vue
```diff
@@ -0,0 +1,63 @@
...
+
+.status-indicator__text {
+    padding: 0.4rem 0;
+    font-size: 0.875rem;
+    line-height: 1.5;
+}
+
+.status-indicator--error {
+    color: var(--notification-danger, #cc0f35);
+}
```
ebkr: I think this is my only remaining concern.

Image

Image


The text isn't very legible on either theme. These screenshots don't look so difficult, but fullscreen the app and it'll become clearer.

Worth adding new CSS variables (using the --var style in the v2 scss) which suit their corresponding backgrounds better.

I believe `--notification-danger` is a background colour for the notification alert boxes rather as opposed to being used for text.

ebkr (changes requested):

src/store/modules/EcosystemUpdateModule.ts
```diff
@@ -0,0 +1,60 @@
...
+            if (state.isInProgress) {
+                return;
+            }
+
+            commit('startUpdate');
+
+            try {
+                // Re-evaluate the ecosystem schema. Future remote refresh
+                // logic should be plugged in here.
+                void EcosystemSchema.supportedGames;
```
ebkr: What does this line do?

ebkr (commented):
Everything looks good, just need a clarification on the void line

src/store/modules/EcosystemUpdateModule.ts
8Qbit: It's a placeholder for the future remote-refresh logic mentioned in the comment above. The void is just to silence no-unused-expressions so the empty try/catch scaffold compiles cleanly. Once fetchLatestSchema() in EcosystemSchema.ts is implemented, the call will go here.

src/store/modules/EcosystemUpdateModule.ts
```diff
@@ -0,0 +1,60 @@
...
+            if (state.isInProgress) {
+                return;
+            }
+
+            commit('startUpdate');
+
+            try {
+                // Re-evaluate the ecosystem schema. Future remote refresh
+                // logic should be plugged in here.
+                void EcosystemSupportedGames.value;
```
ebkr, line 52: As this is a stub, fair enough

ebkr (approved):

PR #2121, opened 2026-04-12
Refactor GenericProfileInstaller into a simple wrapper for PackageInstaller calls

anttimaki:

PR #2130, opened 2026-04-15
Refactor GenericProfileInstaller into a simple wrapper for PackageInstaller calls

anttimaki:
Remake of #2121 which seems wonky after a rebase.


anttimaki:
Never mind, it was pointing to wrong base branch for some reason.

PR #2132, opened 2026-04-16
feat: filter game select by installed on steam

patttterson:
> [!NOTE]
> i didnt test the `SteamLibraryScanner` on windows (or macos), only tested on linux

## Preview
image
image
image

### DIFF
--- a/src/pages/GameSelectionScreen.vue
+++ b/src/pages/GameSelectionScreen.vue
@@ -64,7 +64,7 @@
                                 <i class="button fas fa-th-large" @click="toggleViewMode"></i>
                             </div>
                         </nav>
-                        <nav class="level mb-2" v-else>
+                        <nav class="level mb-0" v-else>
                             <div class="level-item">
                                 <div class="card-header-title">
                                     <div class="input-group input-group--flex margin-right">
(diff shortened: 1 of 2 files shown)


patttterson:
this might be better suited as a part of a bigger general filtering system, which could serve as almost a replacement browser for thunderstore.io

whatchu think about that idea @ebkr

patttterson:
will just implement after game groups become a thing in #2135

ethangreen-dev:
Hey! Meant to respond to this but I forgor. Overall I (as in me, idk about ebkr) like the idea but, yeah, wrap it into ebkr's category work in 2135. It might also be good for you to phrase this as a "filter by installed games" thing and not just by Steam, but that would require a lot more work on your end. 

Just food for thought. I'll chat with ebkr about this and see what a reasonable implementation would look like (or if it's something we want in the first place).

ebkr:
So to add to it, I like the general idea and I do think seeing installed games would be nice.

The UI itself needs a little work but we can probably just remove the configuration bit and pull it from another game's settings.

We could also potentially merge it with a "My games" section which includes games you've previously launched.

I haven't had a look at the code implementation yet as it has been closed, but we can always go over it if/when you decide to re-make this.

patttterson:
> So to add to it, I like the general idea and I do think seeing installed games would be nice.
> 
> The UI itself needs a little work but we can probably just remove the configuration bit and pull it from another game's settings.
> 
> We could also potentially merge it with a "My games" section which includes games you've previously launched.
> 
> I haven't had a look at the code implementation yet as it has been closed, but we can always go over it if/when you decide to re-make this.

for sure, i probably wont work on this till after my exams so give me a little bit lolol

PR #2139, opened 2026-04-25
Add Inscryption XBox Gamepass to Ecosystem

Chaosyr:
This adds Inscryption Xbox Gamepass support to R2ModMan

The only issue im facing is that R2modman isnt actually able to launch it modded, but it does launch successfully *I've tested with local building*

### DIFF
--- a/src/assets/data/ecosystem.json
+++ b/src/assets/data/ecosystem.json
@@ -7547,6 +7547,10 @@
           "identifier": "1092790"
         },
         {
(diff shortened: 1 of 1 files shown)


ethangreen-dev:
Closing this for now as per our discussion in Discord. Supporting the Gamepass version of the game requires significant technical work to be completed in r2modman to support multi-architecture games on different stores. The path forward is to continue advocating for manual installs with the correct Bepinex version.

PR #2142, opened 2026-05-01
Identify newly added games

ebkr:
image


src/components/game-selection/GameSelectionList.vue
```diff
@@ -118,7 +121,7 @@
                                     :is-selected="isGameSelected(game)"
                                     :is-favourited="isFavourited(game)"
                                     :active-tab="activeTab"
-                                    :highlight-favourite="true"
+                                    :is-new="newGames.has(game.internalFolderName)"
```
ethangreen-dev: I couldnt get the badges to render until I changed this to `game.thunderstoreIdentifier` which tracks with how `newGames` is set:

```vue
const newGames = computed(() => {
    const allGames = EcosystemSupportedGames.value.map(([id, game]) => id);
    registerGames(allGames);
    const result = new Set<string>();
    for (const game of allGames) {
        if (isGameNewlyAdded(game)) {
            result.add(game);
        }
    }
    return result;
});
```

Unless im missing something.

ethangreen-dev (changes requested):

ethangreen-dev (approved):

PR #2143, opened 2026-05-04
[TS-3696] Use CDN image assets

ethangreen-dev:
This is basically a four-step approach. The provider:
1. Checks to see if the images are bundled.
2. Checks to see if an instance of the schema is being served on localhost:1337
3. Checks to see if any images have been cached locally.
4. Retrieves the image from the CDN.

This PR has a soft dependency on https://github.com/thunderstore-io/ecosystem-schema/pull/330. I recommend cloning that specific branch to test the localhost functionality. 

It is, however, hard blocked by https://github.com/thunderstore-io/ecosystem-schema/pull/320

My big concern at the moment are to do with race conditions on startup. I didnt see anything strange from my own testing, but there's likely an edge case where the community list renders before image assets have been acquired. Not super experienced with vue so I'm unsure. Will perform more testing later. 

And about the new assets - if needed I can move these to a separate PR.

### DIFF
--- a/quasar.config.ts
+++ b/quasar.config.ts
@@ -101,6 +101,8 @@ export default defineConfig((ctx) => {
             port: 9020,
             open: true, // opens browser window automatically
             proxy: {
+                // Proxy for a local ecosystem-schema deployment.
+                // https://github.com/thunderstore-io/ecosystem-schema
                 '/_local-assets': {
                     target: 'http://localhost:1337',
                     changeOrigin: true,
--- a/quasar.config.ts
+++ b/quasar.config.ts
@@ -100,6 +100,13 @@ export default defineConfig((ctx) => {
             https: false,
             port: 9020,
             open: true, // opens browser window automatically
+            proxy: {
+                '/_local-assets': {
+                    target: 'http://localhost:1337',
+                    changeOrigin: true,
+                    rewrite: (path) => path.replace(/^\/_local-assets/, '/assets'),
+                },
+            },
         },
 
         // https://v2.quasar.dev/quasar-cli-vite/quasar-config-file#framework
--- a/src/components/composables/GameImageComposable.ts
+++ b/src/components/composables/GameImageComposable.ts
@@ -1,12 +1,12 @@
-import { ref, watch } from 'vue';
+import { ref } from 'vue';
 import GameImageProvider from '../../providers/generic/image/GameImageProvider';
 
-export function useGameImageComposable(iconUrl: () => string) {
-    const imageSrc = ref<string>(GameImageProvider.instance.placeholderUrl);
+export function useGameImageComposable() {
+    const imageSrc = ref<string>(GameImageProvider.placeholderUrl);
 
-    watch(iconUrl, async (current) => {
-        imageSrc.value = await GameImageProvider.instance.resolve(current);
-    }, { immediate: true });
+    async function setIcon(iconUrl: string) {
+        imageSrc.value = await GameImageProvider.resolve(iconUrl);
+    }
 
-    return { imageSrc };
+    return { imageSrc, setIcon };
 }
--- a/src/components/composables/GameImageComposable.ts
+++ b/src/components/composables/GameImageComposable.ts
@@ -1,12 +1,12 @@
-import { ref, watchEffect } from 'vue';
+import { ref, watch } from 'vue';
 import GameImageProvider from '../../providers/generic/image/GameImageProvider';
 
 export function useGameImageComposable(iconUrl: () => string) {
-    const imageSrc = ref<string>(GameImageProvider.placeholderUrl);
+    const imageSrc = ref<string>(GameImageProvider.instance.placeholderUrl);
 
-    watchEffect(async () => {
-        imageSrc.value = await GameImageProvider.resolve(iconUrl());
-    });
+    watch(iconUrl, async (current) => {
+        imageSrc.value = await GameImageProvider.instance.resolve(current);
+    }, { immediate: true });
 
     return { imageSrc };
 }
--- a/src/components/composables/GameImageComposable.ts
+++ b/src/components/composables/GameImageComposable.ts
@@ -1,0 +1,12 @@
+import { ref, watchEffect } from 'vue';
+import GameImageProvider from '../../providers/generic/image/GameImageProvider';
+
+export function useGameImageComposable(iconUrl: () => string) {
+    const imageSrc = ref<string>(GameImageProvider.placeholderUrl);
+
+    watchEffect(async () => {
+        imageSrc.value = await GameImageProvider.resolve(iconUrl());
+    });
+
+    return { imageSrc };
+}
--- a/src/components/game-selection/GameSelectionCard.vue
+++ b/src/components/game-selection/GameSelectionCard.vue
@@ -3,7 +3,7 @@
         <div class="game-card">
             <div class="game-card__image">
                 <template v-if="activeTab === GameInstanceType.GAME">
-                    <img :src="getImageHref(`/images/game_selection/${game.gameImage}`)" alt="Game Logo" class="game-card__thumbnail"/>
+                    <img :src="imageSrc" alt="Game Logo" class="game-card__thumbnail"/>
                 </template>
                 <template v-else>
                     <div class="game-card__placeholder">{{ game.displayName }}</div>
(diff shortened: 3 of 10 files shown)


src/providers/generic/image/GameImageProvider.ts
```diff
@@ -0,0 +1,158 @@
+import buffer from '../../node/buffer/buffer';
+import path from '../../node/path/path';
+import CdnProvider from '../connection/CdnProvider';
+import FsProvider from '../file/FsProvider';
+import PathResolver from '../../../r2mm/manager/PathResolver';
+
+const BUNDLED_PROTOCOL_PREFIX = "public://images/game_selection/";
```
ebkr: As discussed, this is incorrect for TSMM and needs to use ProtocolProvider

src/providers/generic/image/GameImageProvider.ts
ebkr, line 1: Also as discussed,. this would probably go hard as a composable

src/providers/generic/connection/CdnProvider.ts
```diff
@@ -62,6 +62,15 @@ export default class CdnProvider {
             : url;
     }
 
+    public static cdnUrlFor(pathSuffix: string): string | undefined {
```
ebkr: I haven't checked, but surely we're already doing this for downloads? Reusable?

ebkr (changes requested):

src/providers/generic/connection/CdnProvider.ts
ethangreen-dev: Good catch, thanks!

src/components/composables/GameImageComposable.ts
```diff
@@ -0,0 +1,12 @@
+import { ref, watchEffect } from 'vue';
+import GameImageProvider from '../../providers/generic/image/GameImageProvider';
+
+export function useGameImageComposable(iconUrl: () => string) {
+    const imageSrc = ref<string>(GameImageProvider.placeholderUrl);
+
+    watchEffect(async () => {
```
ebkr: Why are we using watchEffect?

src/providers/generic/image/GameImageProvider.ts
```diff
@@ -0,0 +1,165 @@
...
+import path from '../../node/path/path';
+import CdnProvider from '../connection/CdnProvider';
+import FsProvider from '../file/FsProvider';
+import PathResolver from '../../../r2mm/manager/PathResolver';
+import ProtocolProvider from '../protocol/ProtocolProvider';
+import { getCdns } from '../../cdn/CdnHostList';
+
+const BUNDLED_ASSET_DIR = "/images/game_selection";
+const PLACEHOLDER_FILE = "thunderstore-beta.webp";
+const LOCALHOST_DEV_BASE = "http://localhost:1337";
```
ebkr: Why? This one might be a viable use-case for the Vite proxy

ebkr (changes requested):

src/providers/generic/image/GameImageProvider.ts
```diff
@@ -0,0 +1,165 @@
...
+const LOCALHOST_PROBE_TIMEOUT_MS = 2000;
+const CDN_FETCH_TIMEOUT_MS = 10000;
+const CDN_BREAKER_THRESHOLD = 3;
+const CACHE_SUBDIR = "image-cache";
+
+export default class GameImageProvider {
```
ebkr: This is not a provider. A new implementation can't be provided.

It doesn't necessarily have to be a provider, but if we'd need any other implementation for TSMM then we'd want to make it one.

src/providers/generic/image/GameImageProvider.ts
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        return GameImageProvider.placeholderUrl;
+    }
+
+    private static bundledUrlFor(iconUrl: string): string {
+        return ProtocolProvider.getPublicAssetUrl(`${BUNDLED_ASSET_DIR}/${iconUrl}`);
+    }
+
+    private static ensureInit(): Promise<void> {
```
ebkr: I'd expect this to be at the top for people to know that it exists, but it seems to just be an internal equivalent to a provider pattern.

I don't think this is necessary? You don't gain any benefit from this provider pattern if usages are kept internal.

src/providers/generic/image/GameImageProvider.ts
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+
+    private static async urlReachable(url: string): Promise<boolean> {
+        try {
+            const res = await fetch(url);
+            return res.ok;
+        } catch {
+            return false;
+        }
```
ebkr: This can probably just be condensed to
```suggestion
        return fetch(url).then(res => res.ok).catch(() => false);
```

src/providers/generic/image/GameImageProvider.ts
```diff
@@ -0,0 +1,165 @@
...
+    private static cachePathFor(iconUrl: string): string | undefined {
+        if (!GameImageProvider.cacheRoot) {
+            return undefined;
+        }
+        return path.join(GameImageProvider.cacheRoot, ...iconUrl.split("/"));
+    }
+
+    private static fileUrlOf(filePath: string): string {
+        return `file:///${filePath.replace(/\\/g, "/")}`;
+    }
```
ebkr: Any need for this?
See `setIconPath` in `ProfileModList.ts`.

I believe there'll be issues with this between TSMM and/or Linux

src/providers/generic/image/GameImageProvider.ts
```diff
@@ -0,0 +1,165 @@
...
+            return undefined;
+        }
+    }
+
+    private static recordCdnFailure(): void {
+        GameImageProvider.cdnConsecutiveFailures += 1;
+        if (GameImageProvider.cdnConsecutiveFailures >= CDN_BREAKER_THRESHOLD) {
+            GameImageProvider.cdnBreakerTripped = true;
+        }
+    }
```
ebkr: What exactly is this recording?

Worth logging (LoggerProvider) / being able to write to sentry. An argument for using the proper provider pattern. It can be a separate provider if we don't need to reimplement anything else.

src/providers/generic/image/GameImageProvider.ts
```diff
@@ -0,0 +1,165 @@
...
+        }
+        return path.join(GameImageProvider.cacheRoot, ...iconUrl.split("/"));
+    }
+
+    private static fileUrlOf(filePath: string): string {
+        return `file:///${filePath.replace(/\\/g, "/")}`;
+    }
+
+    private static async fetchFromCdnAndCache(iconUrl: string, cachePath: string): Promise<string | undefined> {
+        const main = getCdns()[0];
```
ebkr:
```suggestion
        const main = CdnProvider.current;
````
Is probably the way to go

**Review by ebkr — CHANGES_REQUESTED** [review]

File: src/providers/generic/image/GameImageProvider.ts (hunk 4 above)
Comment (ethangreen-dev) [reply src/providers/generic/image/GameImageProvider.ts]: No kidding, that's really convenient. Good catch.

File: src/components/game-selection/GameSelectionCard.vue [hunk 11]
```diff
@@ -51,9 +51,7 @@ const emit = defineEmits<{
     'toggle-favourite': [game: Game];
 }>();
 
-function getImageHref(image: string) {
-    return ProtocolProvider.getPublicAssetUrl(image);
-}
+const { imageSrc } = useGameImageComposable(() => props.game.iconUrl);
```
ebkr: Move the ref inside the composable and return it later to consume it.

Check other composables, we use this pattern.

src/providers/generic/image/GameImageProvider.ts
```diff
@@ -0,0 +1,23 @@
+import ProviderUtils from '../ProviderUtils';
+
+export default abstract class GameImageProvider {
```
ebkr: Nit: This is now the old provider format. We instead opt for a newer method export style. See ProtocolProvider for the new way.

I don't care if we don't switch to it - it's just cleaner.

quasar.config.ts
```diff
@@ -100,6 +100,13 @@ export default defineConfig((ctx) => {
             https: false,
             port: 9020,
             open: true, // opens browser window automatically
+            proxy: {
+                '/_local-assets': {
+                    target: 'http://localhost:1337',
+                    changeOrigin: true,
+                    rewrite: (path) => path.replace(/^\/_local-assets/, '/assets'),
+                },
```
ebkr, line 110: Might want to comment this out?
Add a comment above saying that it's to test the schema locally

ebkr (changes requested):

ebkr (approved):

PR #2145, opened 2026-05-04
Sync root mod files to game directory after install

NikkelM:
When mod loader packages (e.g. ReturnOfModding, BepInEx) are updated via the mod manager, root files like e.g. `d3d12.dll` are only synced to the game directory when clicking "Start Modded". Users who launch games using launch arguments on e.g. Steam will run with an outdated mod loader DLL until they use "Start Modded" again.

This PR calls `ModLinker.link()` after any mod installation completes, using the same logic already used by "Start Modded". The call is non-fatal - if the game is running and files are locked, or the game directory isn't configured, it fails silently and "Start Modded" remains the fallback - though the game will likely still crash just through the new mod being installed, but it won't be due to this change.

Tested this locally for ReturnOfModding/Hades II. After an update to Hell2Modding/any mod, the newest dll was copied to the install directory.


ebkr (approved):
Great idea and works well, thanks

PR #2148, opened 2026-05-07
Pass --overlay-dir to shimloader launch args

ethangreen-dev:


ebkr (approved):
super minor nit, I usually end list lines with a comma so you can avoid the diff line in the future. Makes no difference to me if you add it or not though

PR #2160, opened 2026-05-15
[TS-3708] Pull game images from cdn / localhost in sync script

ethangreen-dev:


scripts/sync.ts
```diff
@@ -58,6 +74,53 @@ async function updateSchema() {
...
+    for (const iconUrl of iconUrls) {
+        try {
+            await fetchGameImage(baseUrl, iconUrl);
+        } catch (e) {
+            console.warn(`  ${(e as Error).message}`);
+            failures += 1;
+        }
+    }
+    if (failures > 0) {
+        console.warn(`Failed to fetch ${failures} of ${iconUrls.size} game images.`);
```
ebkr, line 100: Would be good to say which urls failed

ebkr (changes requested):

scripts/sync.ts
ethangreen-dev: This error message contains the URL: https://github.com/ebkr/r2modmanPlus/pull/2160/changes#diff-2f98d718940dc25a48931b60f8827dd2bcd2e201c705eab7e2e08357d2e229feR95

ebkr (approved):

PR #2163, opened 2026-05-16
More flatpak adjustments

TB516:
- Updates the metainfo file to pin the link to the images instead of having them be the latest on the branch. They probably wont be changing so I just have it pinned against `v3.2.16` instead of making it so the links are generated by a script to match the exact current commit.
- Fixes module block location
- Adds `--filesystem=~/.var/app/com.valvesoftware.Steam:rw` override
- Changes the desktop file and metainfo files to be sourced from the source repo instead of local files

### DIFF
--- a/flatpak/io.github.ebkr.r2modman
+++ b/flatpak/io.github.ebkr.r2modman
@@ -2,10 +2,9 @@
 
 if [ -n "$WAYLAND_DISPLAY" ]; then
   exec zypak-wrapper /app/r2modman \
-    --no-sandbox \
     --enable-features=UseOzonePlatform \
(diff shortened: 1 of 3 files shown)


ebkr (dismissed):

TB516:
Mentioned this in the discord but I want to make a few more changes to address new feedback from the flathub team so don't merge just yet

TB516:
Ok the functionality stuff should be good now and this is good to review/merge. The only things that might change would be permissions if the flathub team wants more restrictive permissions and we can find a solution.

ebkr (approved):

PR #2164, opened 2026-05-16
Update flatpak permissions

TB516:
Explicitly adds the steam flatpak location, because the host permission excludes the flatpak directories, which causes the auto detection to fail on flatpak steam installs when there isnt a link to the steam files in one of the earlier checked paths.


TB516:
Im gana roll this into #2163 actually

PR #2165, opened 2026-05-17
fix: update banner url missing release tag

jukibom:
The update banner includes a clickable link which points to https://github.com/ebkr/r2modmanPlus/releases/tag/ which is a 404. It looks like a previous refactor here omitted the assignment of `updateTagName` which is supposed to be appended to the url in 
``` 
<ExternalLink :url="`https://github.com/ebkr/r2modmanPlus/releases/tag/${updateTagName}`">
```


ebkr:
I think we can actually just replace it entirely with a link to https://github.com/ebkr/r2modmanPlus/releases/latest

jukibom:
> I think we can actually just replace it entirely with a link to https://github.com/ebkr/r2modmanPlus/releases/latest

Oh yeah duh, that makes way more sense! Updated.

ebkr (approved):

ebkr:
Thank you 🔥

PR #2172, opened 2026-05-19
Make OnlinePreviewPanel support full width mode through a setting

Oksamies:
Add a new setting for toggling the preview panel mode
Add full width mode to online preview panel

### DIFF
--- a/src/components/settings-components/SettingsView.vue
+++ b/src/components/settings-components/SettingsView.vue
@@ -260,6 +260,18 @@ let settingsList = [
     ),
     new SettingsRow(
         'Other',
+        'Toggle mod preview full view',
+        'When enabled, clicking a mod opens the preview panel as a full overlay instead of a side panel.',
+        async () => {
+            return settings.value.getPreviewPanelFullView()
+                ? 'Current: full view'
+                : 'Current: side panel (default)';
+        },
+        'fa-exchange-alt',
+        () => emitInvoke('TogglePreviewPanelFullView')
+    ),
+    new SettingsRow(
+        'Other',
         'Refresh online mod list',
         'Check for any new mod releases.',
         async () => {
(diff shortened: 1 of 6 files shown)


ethangreen-dev (changes requested):
Image
The close button (and separators) fail to stretch past a certain point if the window is resized after the details view has been opened. This resolves itself if the view is closed and then re-opened without resizing the window.

src/components/settings-components/SettingsView.vue
```diff
@@ -258,6 +258,18 @@ let settingsList = [
...
+        'Toggle mod preview full view',
+        'When enabled, clicking a mod opens the preview panel as a full overlay instead of a side panel.',
+        async () => {
+            return settings.value.getPreviewPanelFullView()
+                ? 'Current: full view'
+                : 'Current: side panel (default)';
+        },
+        'fa-exchange-alt',
+        () => emitInvoke('TogglePreviewPanelFullView')
+    ),
```
ebkr, line 272: I don't want this - I don't see it being a use case outside of TSMM

ebkr (commented):
So I've looked through it all and I don't particularly want it in r2. I don't see people selecting the full screen view over the split panel and if it becomes wanted then we can always port it across later.

I believe you can just implement this in TSMM and replace the OnlineModList provided value. That way you remove the need for any of the settings changes, and there can just be the barebones implementation of the copied + modified preview panel.

It might also be worth considering a dual-column preview when it's in full view. Left side metadata, right side being the tabs + README/CHANGELOG/Dependencies content.

Oksamies:
> So I've looked through it all and I don't particularly want it in r2. I don't see people selecting the full screen view over the split panel and if it becomes wanted then we can always port it across later.
> 
> I believe you can just implement this in TSMM and replace the OnlineModList provided value. That way you remove the need for any of the settings changes, and there can just be the barebones implementation of the copied + modified preview panel.
> 
> It might also be worth considering a dual-column preview when it's in full view. Left side metadata, right side being the tabs + README/CHANGELOG/Dependencies content.

Makes sense 👍 I approached this from the viewpoint of "ensuring R2 has equal features", but if you're alright with TSMM having a more TSMM specific preview of packages, I'm all good for implementing it only in there 👌

I had a after thought about refactoring the styles of the preview view a bit, so I'll test that out in TSMM and send you some screenshots for comments.

I'll close this PR as it's decided to not be added into r2.

PR #2186, opened 2026-05-29
Fixed "new" indicator if schema is bundled but has no r2modman section

ebkr:


src/r2mm/ecosystem/EcosystemGameStatus.ts
```diff
@@ -22,7 +22,9 @@ export function isGameNewlyAdded(game: string): boolean {
         return false;
     }
     if ((bundledEcosystemSchema.games as any)[game] !== undefined) {
-        return false;
+        if ((bundledEcosystemSchema.games as any)[game].r2modman) {
+            return false;
+        }
```
anttimaki: Would having just one if-block like this work?

```typescript
    if ((bundledEcosystemSchema.games as any)[game]?.r2modman) {
        return false;
    }
```

anttimaki (commented):
I'm not entirely sure how I'm supposed to test this locally?

ethangreen-dev (approved):
Looks good, up to you if you want to go for Antti's recommendation or not.

x753 (approved):

PR #2190, opened 2026-06-06
Add support for > 2GB download and extraction

ethangreen-dev:
This is a simple-as-possible implementation which adds support for package download and extract for archives that are greater than 2GB in size. This is achieved using two mechanisms:
- Downloads now stream to disc in the main process. Instead of the renderer buffering the response via an ArrayBuffer (which has that 2GB limit), a new node:net:download IPC fetches the package and pipes it directly into an output stream.
- Extraction streams one entry at a time. This is done by swapping out adm-zip (which would throw ERR_FS_FILE-TOO_LARGE) for extract-zip, which extracts each entry individually and supports big boy ZIP64.

To keep in mind:
- This has been tested 6 ways to Sunday and it should work fine with TSMM, but I have yet to test it directly. I'm pretty confident though since TSMM's dotnet plugins should wholly supplant this code.
- This drags in an _explicit_ dependency on extract-zip. Technically this package was already present as a dev dependency, so it's not anything new, but this change will now bundle it in the non-dev runtime.

And, to test it:
1. Use this script to create and host a > 2GB archive: 
```python
import os, tempfile, zipfile, functools, http.server
d = tempfile.gettempdir(); z = os.path.join(d, "Big2GBTest.zip")
if not os.path.exists(z):
    p = os.path.join(d, "pl.bin")
    with open(p, "wb") as f: f.truncate(2306867200)
    with zipfile.ZipFile(z, "w", zipfile.ZIP_STORED, allowZip64=True) as zf:
        zf.writestr("manifest.json", '{"name":"Big2GBTest","version_number":"1.0.0","website_url":"","description":"","dependencies":[]}')
        zf.write(p, "payload.bin")
    os.remove(p)
http.server.test(HandlerClass=functools.partial(http.server.SimpleHTTPRequestHandler, directory=d), port=8080)
```

2. Apply this patch to force r2 to download from it:
```patch
diff --git a/src/model/ThunderstoreVersion.ts b/src/model/ThunderstoreVersion.ts
index d18b3f7e..2d50fc5b 100644
--- a/src/model/ThunderstoreVersion.ts
+++ b/src/model/ThunderstoreVersion.ts
@@ -105,7 +105,7 @@ export default class ThunderstoreVersion  {
     }
 
     public getDownloadUrl(): string {
-        return CdnProvider.addCdnQueryParameter(this.downloadUrl);
+        return "http://localhost:8080/Big2GBTest.zip";
     }
 
     public setDownloadUrl(url: string) {
```
3. Download and install any package.

### DIFF
--- a/src-electron/ipc/node-net-impl.ts
+++ b/src-electron/ipc/node-net-impl.ts
@@ -1,0 +1,69 @@
+import { BrowserWindow, ipcMain, net } from 'electron';
+import fs from 'fs';
+import path from 'path';
+import { Readable } from 'stream';
+
+// Throttle progress events so a multi-GB download doesn't flood IPC.
+const PROGRESS_INTERVAL_MS = 200;
+
+export function hookNetIpc(browserWindow: BrowserWindow) {
+    ipcMain.handle('node:net:download', async (event, url: string, destPath: string, downloadId: number) => {
+        await fs.promises.mkdir(path.dirname(destPath), { recursive: true });
+
+        const sendProgress = (loaded: number) => {
+            if (!browserWindow.isDestroyed()) {
+                browserWindow.webContents.send('node:net:download-progress', { downloadId, loaded });
+            }
+        };
+
+        try {
+            await downloadToFile(url, destPath, sendProgress);
+        } catch (e) {
+            // Remove the partial file so isVersionAlreadyDownloaded doesn't later treat
+            // it as a complete download and feed a truncated zip to extraction.
+            await fs.promises.rm(destPath, { force: true }).catch(() => undefined);
+            throw e;
+        }
+    });
+}
+
+function downloadToFile(url: string, destPath: string, onProgress: (loaded: number) => void): Promise<void> {
+    return new Promise((resolve, reject) => {
+        const writeStream = fs.createWriteStream(destPath);
+        const fail = (error: Error) => {
+            writeStream.destroy();
+            reject(error);
+        };
+
+        const request = net.request(url);
+        request.on('redirect', () => request.followRedirect());
+        request.on('error', fail);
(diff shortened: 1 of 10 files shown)


src-electron/ipc/node-net-impl.ts
```diff
@@ -0,0 +1,69 @@
...
+            // it as a complete download and feed a truncated zip to extraction.
+            await fs.promises.rm(destPath, { force: true }).catch(() => undefined);
+            throw e;
+        }
+    });
+}
+
+function downloadToFile(url: string, destPath: string, onProgress: (loaded: number) => void): Promise<void> {
+    return new Promise((resolve, reject) => {
+        const writeStream = fs.createWriteStream(destPath);
```
ebkr, line 32: How well does this handle two of the same dependency being downloaded at the same time, assuming a flow could be that someone invokes a download, closes the modal, goes to another mod and downloads that too.

It might be worth assigning these with temporary IDs, and then rename upon completion.

src-electron/ipc/node-net-impl.ts
```diff
@@ -0,0 +1,69 @@
...
+                }
+            });
+            response.on('error', fail);
+            writeStream.on('error', fail);
+            writeStream.on('finish', () => {
+                onProgress(loaded);
+                resolve();
+            });
+            // Electron's IncomingMessage is a Readable at runtime but isn't typed as one.
+            (response as unknown as Readable).pipe(writeStream);
```
ebkr, line 65: Is this synchronous? Is it not callback defined and then needs some .on handlers?

ebkr (commented):
Added some comments, will test once other work is merged and any conflicts are resolved

PR #2191, opened 2026-06-06
Fix rapidly worsening dexie store startup query

### LINKED ISSUE #2034: [BUG] - Refresh online mod list takes multiple minutes
**Describe the bug**

The mod manager shows any version as outdated. Clicking download doesn't do anything right away but later downloads the mods. Not even the pop up is closed. However I'm not sure what happens if you try to download multiple mods in this mod manager state.
Image

Navigating to the settings, it seems like Refresh online mod list is stuck at "Pruning removed mods from local cache..."

Image

Clicking "Refresh online mod list" after it is done is fast again.

**To Reproduce**
Not sure how to reproduce consistently, since only the first automatic refresh seems to be affected.

**Expected behavior**
My local cache is only about 5GB big, so it shouldn't take around two minutes to prune the mod list.

ethangreen-dev:
Fixes: https://github.com/ebkr/r2modmanPlus/issues/2034

### DIFF
--- a/src/r2mm/manager/PackageDexieStore.ts
+++ b/src/r2mm/manager/PackageDexieStore.ts
@@ -32,7 +32,8 @@ class PackageDexieStore extends Dexie {
             indexHashes: '&community, [community+hash]'
         });
         this.version(3).stores({
-            summaries: '[community+full_name]'
+            summaries: '[community+full_name]',
+            packages: '[community+full_name]'
         });
     }
 }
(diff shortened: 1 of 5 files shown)


ebkr (approved):

src/r2mm/manager/PackageDexieStore.ts
```diff
@@ -128,16 +131,20 @@ export async function isLatestPackageListIndex(community: string, hash: string)
...
-    await db.packages.bulkDelete(oldIds);
+export function selectPackageIdsToPrune(
+    storedKeys: [string, string][],
+    fetchedFullNames: Set<string>
+): [string, string][] {
+    return storedKeys.filter(([, fullName]) => !fetchedFullNames.has(fullName));
+}
+
+export async function pruneRemovedMods(community: string, fetchedFullNames: Set<string>) {
+    // Prune by primary-key diff rather than a [community+date_fetched] range
```
anttimaki: Referring `[community+date_fetched]` in a code comment makes little sense to future readers if date_fetched is completely dropped by these changes.

PR #2192, opened 2026-06-07
Load the online mod list from a flat summary table

ethangreen-dev:
The what?

getPackagesAsThunderstoreMods (great name) scanned the full packages table, forcing indexdb to deserialize every version[] (like 188k in total, about 310MB on lethal-company) just to build a flat per-package list. This caused insane slowdowns at launch, especially for users on underpowered PCs. Also the memory usage was just bad.

The fix? 

Add a summaries table that holds only the fields the list needs. The list read scans it and ignores versions[], while keyed lookups still work the same way.

The performance? 

The lethal-company index, as reported by my benchmarks, had a cold read time reduction from ~2.5s to ~0.5s and peak allocation reduction from ~400MB to ~25MB.

### DIFF
--- a/src/r2mm/manager/PackageDexieStore.ts
+++ b/src/r2mm/manager/PackageDexieStore.ts
@@ -75,7 +75,7 @@ async function rebuildSummaries(community: string): Promise<DexieSummary[]> {
 export async function getPackagesAsThunderstoreMods(community: string) {
     let summaries = await db.summaries.where({community}).toArray();
 
-    if (!summaries.length) {
+    if (summaries.length === 0) {
         summaries = await rebuildSummaries(community);
     }
 
--- a/src/r2mm/manager/PackageDexieStore.ts
+++ b/src/r2mm/manager/PackageDexieStore.ts
@@ -1,6 +1,6 @@
 import Dexie, { Table } from 'dexie';
 
-import { DexiePackage, fetchPackagesByCommunityPackagePairs } from './PackageDexieStoreMockables';
+import { DexiePackage, DexieSummary, fetchPackagesByCommunityPackagePairs } from './PackageDexieStoreMockables';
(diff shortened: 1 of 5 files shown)


src/r2mm/manager/PackageDexieStore.ts
```diff
@@ -30,15 +31,56 @@ class PackageDexieStore extends Dexie {
...
+    });
+}
+
 export async function getPackagesAsThunderstoreMods(community: string) {
-    const packages = await db.packages.where({community}).toArray();
-    return packages.map(ThunderstoreMod.parseFromThunderstoreData)
-                   .sort(ThunderstoreMod.defaultOrderComparer);
+    let summaries = await db.summaries.where({community}).toArray();
+
+    if (!summaries.length) {
```
ebkr: This should be a length comparison - a boolean check whilst valid is incorrect comprehension vs just an assertion to 0

src/store/modules/TsModsModule.ts
```diff
@@ -156,7 +157,10 @@ export const TsModsModule = {
             state.activeGameCacheStatus = status;
         },
         setMods(state, payload: ThunderstoreMod[]) {
-            state.mods = payload;
+            // The mod list is large and immutable, replaced wholesale.
+            // markRaw keeps Vue from deep-proxying every entry, which absolutely dominates
+            // the load memory and time complexity.
```
ebkr, line 162: I think you'll find this has a large impact only in VueX development mode.

If you go to store/index.ts, and turn off strict mode, does this actually make much of a difference?

ebkr (approved):
Approved but check comments

src/store/modules/TsModsModule.ts
ethangreen-dev: This is true!

PR #2208, opened 2026-06-22
Simplify flatpak building & add flatpak repo publishing support

TB516:
Cuts down the flatpak build process by letting the build do network requests (dropping the need for flatpak-node-generator), and by using the same manifest for all builds. Additionally sets up the flatpak repo publishing to allow auto updates to happen.


TB516:
Forgot to mention, mise was having errors when building for windows, so i swapped that to get yarn from the npm backend instead of normal yarn from mise.

TB516:
> Forgot to mention, mise was having errors when building for windows, so i swapped that to get yarn from the npm backend instead of normal yarn from mise.

With everything updated for pnpm this isnt true anymore so should be all good now

ebkr (approved):

PR #2209, opened 2026-06-22
Add `Stoat_Games-TS_Build` to the Mod Exclusions List, its a Tool not a Mod.

Chaosyr:
* Add `Stoat_Games-TS_Build` to the Mod Exclusions.


x753:
@ebkr could I just merge stuff like this whenever it comes up?

ebkr:
> @ebkr could I just merge stuff like this whenever it comes up?

As long as it's actually a valid exclusion, go for it

PR #2211, opened 2026-06-23
Fixed stateMap performance

ebkr:


ethangreen-dev (approved):
Does this actually make a difference? Regardless 🦦

ebkr:
> Does this actually make a difference? Regardless 🦦

Noticeably? No

Efficiency wise? Yes

PR #2228, opened 2026-07-01
Ts 3980/1/typecheck

ebkr:


package.json
```diff
@@ -100,6 +100,7 @@
     "sass": "^1.70.0",
     "sass-loader": "^16.0.5",
     "sinon": "^11.1.1",
+    "tsc": "^2.0.4",
```
x753: This is deprecated and typescript should already do what it needs to do

quasar.config.ts
```diff
@@ -2,8 +2,10 @@
 // https://v2.quasar.dev/quasar-cli-vite/quasar-config-file
 
 import { defineConfig } from '#q-app/wrappers';
+import { ExtendViteConfHandler } from '@quasar/app-vite';
```
x753: Unused import

x753 (changes requested):

x753 (approved):

PR #2231, opened 2026-07-01
Workflow support and vue-tsc + fixes

ebkr:


src/components/v2/OnlinePreviewPanel.vue
```diff
@@ -1,338 +1,338 @@
-<script lang="ts" setup>
```
x753, line 1: Line endings changed here?

x753 (approved):

src/components/v2/OnlinePreviewPanel.vue
ebkr: Yeah, should now be LF instead of CLRF

PR #2237, opened 2026-07-10
Searching now bumps matches based on exact match ranking

ebkr:


x753 (changes requested):
These changes are functional but I don't agree with them. If I'm browsing by highest rated and I search for suits, the first results shouldn't be every mod named `suits` or every package by a team named `suits`, especially without anything separating them from the actual highly rated results. 

My recommendation is to change the default sort behavior of the mod manager to a new sort mode called `Relevant`, which would be the same as last updated except with the exact match ranking bumps.

ebkr:
That's fair, I didn't consider that.

Sort by `Relevant` is a good idea

src/model/enums/SortingStyle.ts
x753: Should the "Default" sort mode just be "Relevance"? Having it just be "Default" means it's just a black box what's actually happening.

x753 (changes requested):

src/model/enums/SortingStyle.ts
ebkr: Default is just matching TS.
It's not clear, but the search behaviour is basically just in order of last updated + pinned.

The alternative is to drop "Relevance" again and instead make the bumping work only on "Default" sort behaviour?

x753 (approved):
Good how it is now.

I think it would be worthwhile to expand the match ranking to treat _ the same as a space, so "More Suits" would show "More_Suits" as the first result for Lethal Company.

PR #2240, opened 2026-07-14
[Redesign] Settings

ebkr:
# Redesigning Settings

_A complete reimplementation of settings_

## Why?

Previously our Settings screen had the following flaws:
- Everything looked indentical.
- The way it was built meant we had to have a constant timer refresh for values.
- Lots of entries.

## How does this fix it?

We can now have:
- More than one action per row.
- Better constraints (EG: enable/disable all mods having a debounce).
- Components to visualise states rather than plain text (EG: switches).
- Individual settings can re-bind their values to a ref, allowing them to now be reactive even when the underlying source isn't reactive, meaning we also no longer require the refresh timer.

## Extras

- Conditional displays no longer need to be appended to the list in a weird onMounted flow. They can be directly referenced inline via `<template v-if...>`.
  - I've dropped the weird array settings handling so that SettingsView.vue can take a layout approach rather than being the driver of everything.
- Better searching (keywords/phrases) and searching by value for text related fields.
- Massive cleanup of Manager.vue - No more business logic.
- The remaining modals in Manager.vue have been ported to actual modal components.

## Caveats

- Each settings component now holds the searching logic (albeit small)
- Additional boilerplate per setting (although I think the tradeoff is worth it)

### All view

image

image

image

image

image

### Section filtered view

image


src/components/settings-components/entries/DataDirectory.vue
```diff
@@ -0,0 +1,108 @@
...
+}
+</script>
+
+<template>
+    <SettingsViewWrapper v-show="isVisible">
+        <template #title>Data and profile directories</template>
+        <template #description>
+            The folder where mods are stored for all games and profiles.
+            The folder will not be deleted, and existing profiles will not carry across.
+        </template>
```
x753, line 85: This should be reworded at a minimum. Below is my suggestion with some other minor formatting differences.
```
<template #description>
	The data folder where mods are stored for all games and profiles.
	<br>
	Changing the data folder does not move or delete existing profiles — they remain in the old folder.
</template>
```

Worth noting that this isn't an AI review, — is just the best punctuation to use here.

src/components/settings-components/entries/GameDirectory.vue
```diff
@@ -0,0 +1,178 @@
...
+
+<template>
+    <SettingsViewWrapper v-show="isVisible">
+        <template #title>{{ activeGame.displayName }} folder</template>
+        <template #description>
+            The game directory is required to place the appropriate files correctly.
+            <span v-if="StorePlatform[activeGame.activePlatform.storePlatform] === StorePlatform.steam">
+                However <code class="code">{{ activeGame.displayName }}</code> will launch without mods if this is not set appropriately.
+            </span>
+        </template>
```
x753, line 161: Ditch the "However" and reformat things
```
<template #description>
	The game folder is required to place the appropriate files correctly.
	<br>
	<span v-if="StorePlatform[activeGame.activePlatform.storePlatform] === StorePlatform.steam">
		<i>{{ activeGame.displayName }}</i> will launch without mods if this is not set appropriately.
	</span>
</template>
```

src/components/settings-components/entries/SteamDirectory.vue
```diff
@@ -0,0 +1,131 @@
...
+        LinkProvider.instance.openLink(dir);
+    }
+}
+</script>
+
+<template>
+    <SettingsViewWrapper v-show="isVisible">
+        <template #title>Steam folder</template>
+        <template #description>
+            Change the location of the Steam folder that {{ appName }} uses.
```
x753: The other folders descriptions are informative, so this reads like an order. Change it to
```
The Steam folder containing <code class="code">steam.exe</code> that {{ appName }} launches games through.
```

src/components/settings-components/entries/ModState.vue
```diff
@@ -0,0 +1,89 @@
...
+    isDisablingState.value = false;
+    await router.push({name: "manager.installed"});
+}
+</script>
+
+<template>
+    <SettingsViewWrapper v-show="isVisible">
+        <template #title>Change mod state</template>
+        <template #description>
+        <p>Enable / disable all of the mods on your profile</p>
```
x753: on your profile -> in your profile

src/components/settings-components/entries/DataDirectory.vue
```diff
@@ -0,0 +1,108 @@
...
+    store.getters['profile/activeProfile']?.getProfilePath() || 'Not set'
+);
+
+const { isVisible } = useSettingSearch(() => props.searchTerm, () => [
+    'Data and profile directories',
+    dataDirectory.value,
+    profileDirectory.value,
+    'Change',
+    'Browse',
+]);
```
x753, line 37: Add 'Folder', 'Directory', and 'Directories' to search terms

src/components/settings-components/entries/DataDirectory.vue
```diff
@@ -0,0 +1,108 @@
...
+    const profilePath = store.getters['profile/activeProfile']?.getProfilePath();
+    if (profilePath) {
+        LinkProvider.instance.openLink(profilePath);
+    }
+}
+</script>
+
+<template>
+    <SettingsViewWrapper v-show="isVisible">
+        <template #title>Data and profile directories</template>
```
x753: Perhaps change this to "folders" instead of "directories"

src/components/settings-components/entries/GameDirectory.vue
```diff
@@ -0,0 +1,178 @@
...
+function syncGameDirectory() {
+    gameDirectory.value = settings.value?.getContext().gameSpecific.gameDirectory || 'Not set';
+}
+
+const { isVisible } = useSettingSearch(() => props.searchTerm, () => [
+    `${activeGame.value.displayName} folder`,
+    gameDirectory.value,
+    'Change',
+    'Browse',
+]);
```
x753, line 42: Add "game", "directory", and "directories" to search terms

x753 (changes requested):
Some of the wording / search terms could use some changes.

src/components/settings-components/entries/ModState.vue
ebkr: Good catch

src/components/settings-components/entries/DataDirectory.vue
ebkr: Will add the others - `Directories` isn't needed because it's in the first phrase

x753 (approved):

PR #2254, opened 2026-08-09
Auto-detect Steam snap install and launch via /snap/bin/steam Fixes

### LINKED ISSUE #1046: [BUG] - Steam executible is not selectable when using Steam snap
**Describe the bug**
The option to set Steam executable doesn't allow you to select the steam executable.

**To Reproduce**
1. Click 'change Steam directory' setting in r2modman
2. Navigate to  '~/snap/bin/'
3. See that Steam executable is not shown despite being in folder

**Expected behavior**
Steam executable should be shown and select-able.

**Additional context**
I'm betting this is because the Steam snap is being used and something about that is unaccounted for. In this instance just switching to a native Steam package or the Flatpak is NOT VIABLE and CANNOT be used as a solution.

kenvandine:
Steam's snap package keeps its data under ~/snap/steam/common instead of the usual ~/.local/share/Steam, so it was never found by getSteamDirectory()'s candidate list.

Add those paths as candidates, and when the resolved directory belongs to the snap, launch Steam through its own /snap/bin/steam wrapper instead of exec'ing steam.sh directly, since invoking the snap's internal script from outside its confinement doesn't reliably work.

Fixes #1046

### DIFF
--- a/src/r2mm/launching/runners/linux/SteamGameRunner_Linux.ts
+++ b/src/r2mm/launching/runners/linux/SteamGameRunner_Linux.ts
@@ -80,7 +80,12 @@ export default class SteamGameRunner_Linux extends GameRunnerProvider {
         if (steamDir instanceof R2Error) {
             return steamDir;
         }
-        const steamExecutable = `${steamDir}/steam.sh`;
+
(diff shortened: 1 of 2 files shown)


Phqen1x:
Tested this PR and it works great for me! Thanks!

PR #2255, opened 2026-08-12
Fix preview panel resize soft lock

Glubus:
## Summary

  Fixes a UI soft lock that can occur when resizing the online mod preview panel, particularly on Linux/Wayland.

  ## Root cause

  The preview panel separator used the native HTML d&d API (`draggable` and drag events) as a resize mechanism.

  On Electron under Wayland, moving the separator could leave the application inside the native d&d interaction. The UI would then stop responding after the panel moved beneath the pointer.

  ## Changes

  - Replace native HTML drag-and-drop with pointer events.
  - Use pointer capture to keep receiving move and release events during resizing.
  - Handle pointer cancellation and lost pointer capture.
  - Clamp the panel width to the available layout bounds.
  - Prevent invalid panel widths from being persisted.
  - Add unit tests for the width-clamping behavior.

  ## Testing

  - `pnpm typecheck`
  - `pnpm exec vitest run test/vitest/tests/utils/PreviewPanelResizeUtils.spec.ts`
  - Manually reproduced and verified on EndeavourOS with KDE Wayland.
  - Tested using the Risk of Rain 2 online mod preview panel.
  
[The Bug on KDE Wayland ](https://github.com/user-attachments/assets/0c51140a-25a9-4ab5-9cea-811b1d73b67d)

### DIFF
--- a/src/components/v2/OnlinePreviewPanel.vue
+++ b/src/components/v2/OnlinePreviewPanel.vue
@@ -1,6 +1,6 @@
 <script lang="ts" setup>
 import ThunderstoreMod from '../../model/ThunderstoreMod';
-import { computed, ref, watch, watchEffect } from 'vue';
+import { computed, ref, watch } from 'vue';
 import MarkdownRender from './MarkdownRender.vue';
 import { valueToReadableDate } from '../../utils/DateUtils';
 import OnlineModList from '../views/OnlineModList.vue';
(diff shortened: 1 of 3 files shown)


ebkr:
@Glubus I'm happy to merge this if you're able to rebase.

I tested it locally and wasn't able to reproduce the issue, but it still worked the same so I'm happy if it solves your issue.

PR #2266, opened 2026-08-25
It simply replaces '_' with spaces in the displayed game titles

MaxEpos:
It simply replaces '_' with spaces in the displayed game titles. Thanks to this minor change, the titles are more readable. I believe I’ve applied the change everywhere necessary.


ebkr:
It's somewhat a difference in how I think that the names should be shown. On the site, they treat underscores as spaces, however it's not accurate to the actual uploaded name and I'd rather show what was actually uploaded.

There's not a massive difference however and no huge push for one over the other.

I'll close the PR for now, as it's unlikely to be merged anytime soon and needs some more discussion internally on if it's something we should or shouldn't do. It's also not currently viable with the i18n change that's going to be merged soon.

PR #2279, opened 2026-09-09
Fix Northstar profile paths on Proton and type checking after Flatpak builds

TB516:
This PR was created by GPT-6

Northstar receives a Unix profile path on Linux with Proton and resolves it on the game's Windows drive. Resolve the Northstar directory with `realpath()` and prefix it with `Z:` when `isProton()` applies, matching the existing loader patterns. Other platforms and launch types retain their current behavior, and failures return `R2Error`.

Also exclude generated Flatpak directories through Quasar's `extendTsConfig` hook. Type checking otherwise scans thousands of bundled JavaScript files from local builds and exhausts Node's heap.

Validation:
- `pnpm typecheck` passes with the default heap, including after a Flatpak build.
- Existing tests: 200 passed, 5 skipped, with generated Flatpak test copies excluded.
- Built a local `.flatpak`; Thomas confirmed it works with the separately patched NorthstarLauncher that supports Windows absolute profile paths.

### DIFF
--- a/quasar.config.ts
+++ b/quasar.config.ts
@@ -50,6 +50,8 @@ export default defineConfig((ctx: QuasarContext) => {
                 strict: true,
                 vueShim: true,
                 extendTsConfig(tsConfig: any) {
+                    // Flatpak output includes bundled JavaScript that should not be type-checked.
+                    tsConfig.exclude.push('../flatpak/build', '../flatpak/repo', '../.flatpak-builder');
                     tsConfig.compilerOptions.paths = {
                         '@r2': ['../src'],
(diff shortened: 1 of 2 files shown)


TB516:
Quick fix with codex for the new northstar errors that we are seeing with r2. Should be accompanied by [this fix](https://github.com/R2Northstar/NorthstarLauncher/pull/939) in the northstar repo which allows it to load absolute paths with the -profile arg.

TB516:
I also tossed in a change in the tsconfig to exclude the flatpak stuff which I think was missed by me initially, but it was causing my tsc to run out of memory from enumerating those files.

